PR #6, opened 2014-09-30
New site

rosshj:
@altsang @crandmck 

First commit of all the new site files. I also removed all the old site files.


CNAME
```diff
@@ -1 +0,0 @@
-loopback.io
```
rmg, line 1: This needs to be restored or the site will disappear when this PR is merged.

PR #9, opened 2014-10-10
Added API Gateway to the About/Landing page

rosshj:
New bullet point under About + new section 'LoopBack API Gateway' near the bottom of the page.


altsang:
:+1:

Issue #14, opened 2015-02-18
Add new Contributing page with corresponding navigation elements


crandmck:
Per @altsang we want to have a new "Contributing" page that will contain the info I added in 4d335a276b43a0a70b04734caf79bb5218c695b7 and a bit more.  We will also need a "Contributing" navigation element in the header.  

I haven't been able to figure out how to add a new nav element without screwing up the layout, so please @rosshj could you do this in a new branch?  If you can add the nav element and new page, I can provide the content before we merge the PR.  Please go ahead and merge the `customers-tweaks` branch first so that can be live while we're working on this.

rosshj:
@crandmck @altsang 

I'm concerned about the amount of navigation links are going into the header. It's starting to get a bit bloated. Usually you want to keep the main navigation focused and concise for not only keeping the design in check, but also clear for the user. With the addition of 'Who's Using' and now 'Contributing' we will be up to 6 + the Github star.

Maybe we can look into consolidating a couple of these sections?

Let me know. If you still want it up there, I will have to refactor the header design a bit to make room.

rosshj:
In some cases (certain screen sizes) with the addition of 'Who's Using' it's already breaking.

![screen shot 2015-02-18 at 10 07 18 pm](https://cloud.githubusercontent.com/assets/723030/6262557/ad02cef6-b7ba-11e4-82fa-ebcc99edb760.png)

crandmck:
Yes, I agree, but need to discuss with @altsang.  We could leave this stuff in the "Resources" page, put it somewhere else, orr rethink the nav entirely.

crandmck:
Al noted that currently the "About" nav link is the same as the home page (and the LoopBack logo), so it's redundant.  He suggested that we could remove "About" and add "Contributing" without making the nav header too big.  Will that work?

rosshj:
That could work. I will still need to adjust the header a little bit, but I should be able to squeeze all that in if we drop the 'About' section. We could always keep 'About' link in the footer nav.

crandmck:
OK, please do that.  I think it's the best option at this point.  If you add a stub "Contributing" page in a branch, I will add the requisite content.

crandmck:
@rosshj Do you need any more information to move forward on this?

rosshj:
@crandmck @altsang Hey guys. I'm going to try and take a stab at this later tonight. 

Is there more content than just what appear on 'Contributing to LoopBack' section this page? http://loopback.io/resources/

[...]

PR #16, opened 2015-03-03
Add governance sections with list of maintainers

crandmck:
@altsang Here is the new info on governance and maintainers.

### DIFF
--- a/contributing/index.html
+++ b/contributing/index.html
@@ -67,20 +67,61 @@
                   <li><a href="https://github.com/strongloop/generator-loopback">generator-loopback</a></li>
                 </ul>
                  <p>See <a href="http://docs.strongloop.com/display/LB/LoopBack#LoopBack-LoopBackframeworkmodules">LoopBack  documentation</a> for a complete list.</p>
-                <h2 class="c-heading -h3">Contributing code</h2>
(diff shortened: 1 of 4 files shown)


altsang:
looks like loopback-boot is listed twice on the modules - we should have `loopback` as the first core module, can add more later, otherwise :+1:

crandmck:
Good catch.  I fixed that.  The file I checked in has the "badges" described in my last comment in #14 commented out in the HTML.  If you want to check it out for yourself, clone the branch and remove the comment section in `contributing/index.html`.

crandmck:
@altsang Can we go ahead and merge this with just the bullet list of maintainers? 

I'll come back to the task of making them look like "cards" which will take a bit more time (the HTML for that is commented out currently).  But I don't think we need to hold up this whole thing for that....

altsang:
:+1:

On Thu, Mar 5, 2015 at 10:28 AM, Rand McKinney notifications@github.com
wrote:

> @altsang https://github.com/altsang Can we go ahead and merge this with
> just the bullet list of maintainers?
> 
> I'll come back to the task of making them look like "cards" which will
> take a bit more time. But I don't think we need to hold up this whole thing
> for that....
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/strongloop/loopback.io/pull/16#issuecomment-77420559.

## 

Al Tsang
Co-Founder and CTO @ StrongLoop, Inc.

_StrongLoop http://strongloop.com/_\* makes it easy to **develop APIs
http://strongloop.com/mobile-application-development/loopback/** in Node,
plus get DevOps capabilities
http://strongloop.com/node-js-performance/strongops/ like monitoring,
debugging and clustering*_._

PR #43, opened 2016-09-08
Repairing anchors

doublemarked:
I noticed a heap of broken anchor links and hacked together a script that fixed many of them. I've checked most of these fixes and they seem to work.

FYI, in the process I noticed some other cross-document links that are broken. A more comprehensive search and fix for broken links is probably required.


crandmck:
Thanks!  A comprehensive search for broken links is a great idea, but not sure how to go about it.  Can we just run a link-checker on the site?

doublemarked:
Yes, it should be fairly easy to identify broken links, though some trouble may come in trying to identify the correct targets. It's a process that may be a hassle to automate, but easy enough to do manually (though tedious).

crandmck:
I can help with the "correcting the targets" part.  @doublemarked It would be awesome if you could help with the "identifying broken links" part. I opened #44 for this general task...thanks!

doublemarked:
Sure, I can give it a look!

PR #70, opened 2016-09-29
Update and improve 3.0 Release Notes

bajtos:
I have reviewed, updated and improved 3.0 Release Notes. I have identified few more missing items, I'll add them to both release notes and the migration guide in a new PR.

@crandmck please review.

I noticed the top-level blue note box is eats the whole page, because the closing `<div>` elements is quoted in HTML. I tried few tweaks in the source but could not fix the problem. Perhaps it's just my machine?

Connect to strongloop-internal/scrum-loopback#908

### DIFF
--- a/pages/en/lb3/3.0-Release-Notes.md
+++ b/pages/en/lb3/3.0-Release-Notes.md
@@ -12,211 +12,498 @@ summary:
 {% include note.html content="
 LoopBack 3.0 has not yet been released. These release notes reflect the current state of the project, and may change before release." %}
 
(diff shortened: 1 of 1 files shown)


bajtos:
I think it may be useful to cross-link each item in Release Notes with the appropriate instructions  in Migration Guide, thoughts?

crandmck:
LGTM.  I will edit ASAP.

> I noticed the top-level blue note box is eats the whole page

@bajtos Yes, this is an unfortunate behavior of these Jekyll "include templates".
The fix is to put the ending `" %}` on a new line.  Simple fix, but somewhat annoying!
I'll take care of that.

Issue #66, opened 2016-09-27
Enhance sidebar nav to take another level of hierarchy


crandmck:
Currently, the sidebar nav only allows three levels of hierarchy:
- Top-level "folders"
  - Second-level "subfolders"
    - Third-level "subfolder items"

This works for most of our content, but there are a few areas where we could use one more level.  The code that processes the `yml` sidebar file does not currently support this.  

In addition, there is an oddity left over from the original implementation where the _first_ item in a "folder" cannot be a "subfolder" (only a non-container leaf node is allowed).  For example, I had to stick a dummy item http://loopback.io/doc/en/lb2/Defining-models.html under the container, because the first item, [Creating models](http://loopback.io/doc/en/lb2/Creating-models.html), is a "subfolder". Without the intervening item, jekyll gives a build error:

```
           Error: (/Users/rand/StrongLoop/loopback.io/_data/sidebars/lb2_sidebar.yml): did not find expected key while parsing a block mapping at line 236 column 5
```

I would like to be able to make an article be a "container" in the sidebar, regardless of where it occurs.

Sequoia:
[`links.html`](https://github.com/strongloop/loopback.io/blob/gh-pages/_includes/links.html#L3-L20) purports to also use the sidebar menu scheme (which I'm revamping), and it looks like [all the pages in `/pages/en/contrib/`](https://github.com/strongloop/loopback.io/blob/gh-pages/pages/en/contrib/) _include_ the file, but **it appears to do nothing**.

To wit, when I add `<hr />`s at the top and bottom of `links.html`, the output looks like this:

![](https://i.imgur.com/r8r3oqK.png)

Note that there is nothing there.

For this reason I **do not** plan to update `links.html` at this time, as it's not clear that it actually does anything. Please let me know if you know what it does & it's important to update (this can always be done after).

Sequoia:
Other templates that use the existing menu YAML schema:
- [`tocpage.html`](https://github.com/strongloop/loopback.io/blob/gh-pages/scripts-pdf/pdfconfigs/tocpage.html): Looks like its for PDF generation. Will update
- [`prince-list.txt`](https://github.com/strongloop/loopback.io/blob/gh-pages/scripts-pdf/pdfconfigs/prince-list.txt): Also appears to be related to PDF generation? Will update as well (or investigate)
[...]

PR #73, opened 2016-09-29
Reworked sidebar menu

Sequoia:
# New Menu Schema

I made all the sidebars I could find in `_data/sidebars` match the new, simplified schema.

`topnav.yml` still uses the old scheme because
1. it needs `external_url`
2. it doesn't use the same (sidebar) template so there's no conflict
# Performance

The new setup increases build time from ~18 seconds to ~25 seconds. Initially it was over 30 seconds but I did what I could to trim that time down (mostly removing `assign`s). Lots of the info I found said "reduce the number of `for` loops" but that's simply not an option here-- the menu is a tree and each level must be stepped thru.

I spent time with various profiling tools but they didn't do much beyond verify that yes, it's the loops that are causing the performance hit. I have the CPU profiles if anyone wants to look thru them! 😸 
screen shot 2016-09-29 at 1 55 27 pm
## Things that might help
1. Incremental builds if we could get it working
2. Cached includes plugin (build the menu 1x per build & cache it for all pages, would require setting `active` class from front-end, fairly trivial). I investigated building such a plugin but it looked like it would take too long. [more info](https://talk.jekyllrb.com/t/caching-includes/3036)
3. Manually "caching" the sidebar. Essentially, making a separate build task just for the sidebar, running it when sidebar changes, then `{%include%}`ing the _built_ output.
# Other languages

As these currently have no language-specific content (they are just placeholder files, essentially, waiting to be translated), I simply overwrote them with thew new LB 2 menu scheme.
# PDFs

I left the PDF templates alone for now. Getting PDF generation to work is a whole 'nother task that will take some time (see [instructions](idratherbewriting.com/documentation-theme-jekyll/mydoc_generating_pdfs.html)). When it's decided that we want to do this, it shouldn't be difficult to update the PDF templates to match the new sidebar template.

Files that I did not update:
- `scripts-pdf/pdfconfigs/tocpage.html`
- `scripts-pdf/pdfconfigs/prince-list.txt`

### DIFF
--- a/_data/sidebars/community_sidebar.yml
+++ b/_data/sidebars/community_sidebar.yml
@@ -1,25 +1,22 @@
-entries:
-- title: sidebar
-  product: LoopBack community projects
-  version:
-  folders:
+title: LoopBack community projects
+children:
+
+- title: 'Community projects'
+  output: 'web'
+  children:
 
   - title: 'Community projects'
+    url: /doc/en/community/index.html
     output: 'web'
-    folderitems:
 
-    - title: 'Community projects'
-      url: /doc/en/community/index.html
-      output: 'web'
-
-    - title: 'How to add a project'
-      url: /doc/en/community/How-to-add-a-project.html
-      output: 'web'
-
-  - title: 'Active projects'
+  - title: 'How to add a project'
+    url: /doc/en/community/How-to-add-a-project.html
     output: 'web'
-    folderitems:
 
-    - title: 'Sample project'
-      url: /doc/en/community/Sample-project.html
-      output: 'web'
+- title: 'Active projects'
+  output: 'web'
+  children:
+
+  - title: 'Sample project'
+    url: /doc/en/community/Sample-project.html
+    output: 'web'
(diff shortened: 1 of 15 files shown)


slnode:

Sequoia:
closes #66

slnode:

crandmck:
Will review ASAP, thanks Sequoia!

crandmck:
This looks great, in general.  It will give us more flexibility in structuring the document hierarchy, and is much easier to work with because the YAML format is simpler and more consistent.

There are some problems with the changes you made to the localized sidebars, though.  Specifically:
- Russian and Japanese had some translated titles (in the `yml` sidebar file), which were lost.
- You replaced the urls of all translated articles with `en` URLs, but I migrated over some translated articles from the old doc site (the content still needs some cleanup, but it's there).  Only two in `ja` but a bunch in `ru`. 

Can you fix these before I merge?

crandmck:
@Sequoia I need to do some other stuff, so I'll go ahead and merge this and fix the above-noted issues, nbd.

PR #74, opened 2016-09-29
Converted Chinese Documentation (zh)

marc-ed-raffalli:
Added raw converted content and side bar.

The original file name is used when it contains chinese / special chars.


slnode:

_data/sidebars/zh_lb2_sidebar.yml
```diff
@@ -1,5 +1,899 @@
 entries:
-- title: sidebar
-  product: LoopBack - 中文
-  home_url: /doc/zh/lb2/index.html
-  folders:
+
```
Sequoia, line 9: I just updated the format of this file in #73 but the difference is not major, just a couple substitutions.

crandmck: I'll review and merge your PR asap.  Then, we'll have to rework these sidebars before merging this new Chinese content.  Just finishing up something else atm.

marc-ed-raffalli: @crandmck, any actions required on my side to solve the merge conflict ?

crandmck:
Hi @marc-ed-raffalli Sorry for the delay.  
A couple of problems need to be addressed in your fork/branch.

It doesn't run "as is".  Jekyll gives this error:

```
Liquid Exception: Invalid syntax for include tag: The model _must_ exist (even if the 
schema is empty) before your app can interact with it. Valid syntax: 
{% include file.ext param='value' param2='value' %} in pages/zh/lb2/6094913.md
```

I think these errors are caused by the image includes, e.g. 

```
  {% include image.html file="6258742.png" alt="" %}
```

Because the image files referenced are not there.  What happened is that the export created files with different names for the images (even though they are actually the same images as used in `en`).  So, the options to fix this are:
- Update the references to use the existing image file names (as used in the en docs).  I don't know how one would do this, other than a laborious manual process.  Is there a clever way to do this?
- Add the images from the zh export.  This would end up with duplicates of _every_ image, which is not ideal.
- Remove all images, i.e. instance of `{% include image.html... %}` which would remove all images from the files.  Also not great, but would help it to build.

Also, could you please update the sidebar `.yml` file to have the new format.  It's actually much simpler. Please see https://github.com/strongloop/loopback.io/blob/gh-pages/_data/sidebars/lb2_sidebar.yml as an example.

marc-ed-raffalli:
no problem.
About the sidebar, just the structure or the content should be updated?

crandmck:
You'll need to update the structure (ie. the format) of the `.yml` for the site to build and work.  You can leave the content (titles and organization) for the translated articles "as is".  

The only issue is that we've reorganized the English sidebar somewhat, so  ideally we'd replace the  untranslated part of the sidebar with the corresponding section of the en sidebar (replacing `en` with `zh` of course).  If that last part is confusing, then don't worry--I can do take care of it.

The most important part is just getting the Chinese content in there in some form...and of course, not breaking the site :-)  I suspect we really need a Chinese-speaker to go through and clean it up to really get it in shape.  Maybe once we have the basic content in place, someone will step up....

Sequoia:
I went to rebase this onto gh-pages & fix the sidebar menu but there are problems building the pages from the markdown.

```
~/strongloop/loopback.io on  b0f94ad! ⌚ 17:12:45
$ bundle exec jekyll serve -w -I
Configuration file: /Users/sequoia/Google Drive/Freelance/strongloop/loopback.io/_config.yml
            Source: /Users/sequoia/Google Drive/Freelance/strongloop/loopback.io
       Destination: /Users/sequoia/Google Drive/Freelance/strongloop/loopback.io/_site
 Incremental build: enabled
      Generating...
  Liquid Exception: Invalid syntax for include tag: Note also that `Model.scope()` is only in `create-customers.js`. Valid syntax: {% include file.ext param='value' param2='value' %} in pages/zh/lb2/6094931.md
jekyll 3.0.3 | Error:  Invalid syntax for include tag:

  Note also that `Model.scope()` is only in `create-customers.js`.

Valid syntax:

  {% include file.ext param='value' param2='value' %}



~/strongloop/loopback.io on  b0f94ad! ⌚ 17:12:52
$ bundle exec jekyll serve -w -I
Configuration file: /Users/sequoia/Google Drive/Freelance/strongloop/loopback.io/_config.yml
            Source: /Users/sequoia/Google Drive/Freelance/strongloop/loopback.io
       Destination: /Users/sequoia/Google Drive/Freelance/strongloop/loopback.io/_site
 Incremental build: enabled
      Generating...
  Liquid Exception: Invalid syntax for include tag: values automatically when you don't specify one. Valid syntax: {% include file.ext param='value' param2='value' %} in pages/zh/lb2/6094931.md
jekyll 3.0.3 | Error:  Invalid syntax for include tag:

  values automatically when you don't specify one.

Valid syntax:

  {% include file.ext param='value' param2='value' %}
```

It appears to not like putting `{{page.lang}}` in the `content` parameter value for an include. I could be wrong here. In any event I was not able to rebase & get the blog to build.

marc-ed-raffalli:
Ok, I'll look at this once I finish this picture merging. As a temporary solution, I can hard code `{{page.lang}}` in the includes

crandmck:
Sorry, I should have mentioned this sooner... Yes, you cannot have a liquid variable in an include parameter.  I've addressed this elsewhere by simply removing the `/doc/{{page.lang}}/lb2/` from the links.  Just using the file name will work. Here's an example: https://raw.githubusercontent.com/strongloop/loopback.io/gh-pages/pages/en/lb2/Adding-logic-to-models.md

In fact, I thought about removing `/doc/{{page.lang}}/lb2/` from _all_ links, because I'm not sure what value it provides.... but I haven't done that yet.

marc-ed-raffalli:
Ok, I removed nested `{{page.lang}}` and replaced by `zh` so the build passes.
I also merged the attachments, updating image source or adding new picture.

As this scenario might occur again and is quite time consuming, here's what I found quite time efficient:
- Get the list of new pictures as text e.g. `ls -l`
- Search for each picture to verify it is actually used in the new doc and not just a previous version. This allowed to safely remove a good 50 / 60 %.
- (Tedious part): Have the new and `images` folders side by side with thumbnails view:
  - replace name in new doc when the pictures are the same
  - import new picture when not found

crandmck:
Wow @marc-ed-raffalli you've gone above and beyond.  I know that was very tedious, so THANK YOU. Getting the old translations over to the new site has been much more work than I had anticipated!

It's late here now (and I've had a couple glasses of wine ;) so I'm not going to merge this now, but I will tomorrow.

Thanks again.

marc-ed-raffalli:
No problem @crandmck, you can merge whenever you get a chance.
It was not the most exciting task I ever worked on indeed, but it looked for me like the best option. Now if you want, you can still send a Loopback sticker :P

crandmck:
We can definitely send you some swag!  Just send me your mailing address (to rmckinn@us.ibm.com), and we'll get something to you asap!

PR #75, opened 2016-09-29
Layout tweaks

### LINKED ISSUE #69: Layout tweaks
There are a few styling tweaks I would like.  @Sequoia I can probably figure these out (when I can find the time) if I try, but if you can easily fix them, please do so.

There is a bit too much whitespace between the page content and the title bar (top bar).  I know a certain amount of whitespace is good, but screen real estate is also valuable, so I just think we can bring it down by 10 - 20px or so...

doc-title-whitespace

The blue highlight that appears around a title in the navbar needs a little more space on the left--probably just a few px more padding:
doc-nav-layout

Also, there is a small glitch in the top bar drop-downs:

dropdown styling

The green highlight should extend all the way to the left edge of the "dropdown" (as it does on the right side).

Sequoia:
Closes #69

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -26,7 +26,6 @@
 {% include topnav.html %}
 <!-- Page Content -->
 <div class="container">
-    <div class="col-lg-12">&nbsp;</div>
     <!-- Content Row -->
     <div class="row">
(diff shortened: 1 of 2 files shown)


slnode:

PR #76, opened 2016-09-29
Schema column order

domwrap:
Schema column order was incorrect. Rearranged, plus added data type.


slnode:

crandmck:
Thanks!

PR #95, opened 2016-10-07
Add markdown linting script

Sequoia:
Related to #49 (tho it does not necessarily close it). See #49 for extensive discussion of the problem & solution contained herein.

This PR does **not** tie the linting task into the `fetch-readmes` task as almost all readmes would fail currently. Once all readmes are cleaned up & commit hooks are set up on targeted repos, we can add the check here as well if needed.

### DIFF
--- a/.markdownlintrc
+++ b/.markdownlintrc
@@ -1,0 +1,5 @@
+{
+  "default": true,
+  "MD018": true,
+  "MD003": { "style": "atx" }
+}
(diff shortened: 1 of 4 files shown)


slnode:

crandmck:
I just noticed that this uses `https://github.com/Sequoia/markdownlint-cli`.  It's fine for now, but in the long run I think we'd prefer to use https://www.npmjs.com/package/markdownlint-cli, or if we can't upstream your changes, then fork it into the strongloop org.

Sequoia:
I had to fork that repo because there was a sort of major bug affecting long output. I submitted a PR but I didn't want to wait to use it so I pointed to my version. Here's the PR https://github.com/igorshubovych/markdownlint-cli/pull/11 as soon as it lands we can point this to upstream.

-Sequoia

PR #101, opened 2016-10-12
Explain limitation of where filter

davidcheung:
Where filter in `api/Books?filter[where][or][0][id]=1` format
can only handle 20 parameters.
See strongloop/loopback/issues/2824 for more details

it will look something like the following picture to save you some compile time if you only want to make text changes.

screen shot 2016-10-12 at 6 43 43 pm

@crandmck PTAL


crandmck:
Thanks!!

PR #103, opened 2016-10-13
Update Define-access-controls.md

antarasi:
First, deny all endpoints, not only Review model


slnode:

crandmck:
Good catch, thanks!

PR #104, opened 2016-10-13
updating external readmes

Sequoia:
Related to changes precipitated by discussion on #49


slnode:

Sequoia:
(I ran `npm run fetch-readmes` & committed the resulting new versions of those files)

Sequoia:
Incidentally, should that be run automatically on blog build? is that possible? Is this blog built by github?

crandmck:
> Incidentally, should that be run automatically on blog build? is that possible? Is this blog built by github?

Yes, it's automatically rebuilt whenever you commit to the gh-pages branch, and I don't know how we would trigger this.  As we discussed, running manually is OK for now.

PR #112, opened 2016-10-14
Remove property lazyConnect in doc

jannyHou:
Connect to https://github.com/strongloop/loopback/issues/2850

There is a bug of ds config property `lazyConnect`, it won't affect our internal use (when used by apic), but it's better not documented to let users set it.

It's not a easy fix so as a workaround I delete `lazyConnect` in doc.

### DIFF
--- a/pages/en/lb2/datasources.json.md
+++ b/pages/en/lb2/datasources.json.md
@@ -65,10 +65,6 @@ All data sources support a few standard properties. Beyond that, specific proper
       </td>
     </tr>
     <tr>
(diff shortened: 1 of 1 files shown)


jannyHou:
@crandmck PTAL. See details above.
Thanks.

crandmck:
OK, I only added this recently.  I didn't realize that it shouldn't be documented.

jannyHou:
@crandmck thanks. My bad lol. I didn't realize including it in datasources.json would cause trouble when I asked you to add the doc.

PR #122, opened 2016-10-19
Update eslint doc

Amir-61:
- Removing unnecessary step for eslint process. There is no need to copy `comma-dangle.js` file anymore since https://github.com/strongloop/loopback/pull/2720 was landed

/to: @strongloop/loopback-dev @superkhau 
/cc: @crandmck

### DIFF
--- a/pages/en/contrib/eslint-guide.md
+++ b/pages/en/contrib/eslint-guide.md
@@ -108,38 +108,6 @@ Disclaimer: this can be a tedious task.
 
 - Go to the files and fix errors manually and commit.
 
-  - **Special case:**
-    There is an issue with `one-var` rule in `eslint` which reports an error for a variable declared in a `for-loop` after an uninitialized variable declaration. for example:
-
-```
-  function reproduce(candidates) {
-    var err;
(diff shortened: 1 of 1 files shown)


superkhau:
@Amir-61 I think you're removing the wrong lines. The actual content you're linking to has to do with the `one-var` rule (see https://github.com/strongloop/loopback.io/pull/122/files#diff-f0436c469cfe5f5dce4559ed6f85b2e4R110).

However I also think we don't need the comma-dangle file too, but please submit that separately (or update the references/test if that is actually true, etc).

superkhau (changes requested):
Update the PR to reflect one-var instead.

crandmck:
Based on @Amir-61 latest commit, LGTM.  @superkhau you want to give final approval?

Amir-61:
@superkhau 

Applied feedback.

> However I also think we don't need the comma-dangle file too, but please submit that separately 

Sure I sent another PR for that based on your suggestion. PR: https://github.com/strongloop/loopback.io/pull/124

superkhau (approved):

PR #126, opened 2016-10-19
Update Glossary.md

emckean:
added missing end punctuation for 'model property' entry


slnode:

crandmck:
Thanks, Erin.  In general, I'd prefer it if you put all your edits to a single file in one PR--if you can.  Obviously, if you open a PR and then subsequently have another edit, that's OK.  But if you can accumulate all your edits to a file into one PR that's best. 

I guess you asked that before, but I didn't think to carefully about my answer! :-)

PR #134, opened 2016-10-25
Fix migration information in new error handler

davidcheung:
the migration content was wrong (release note is correct), it should be `config.json`, and seems like the info wasnt detailed enough for user to follow, added small code block 

https://github.com/strongloop/loopback/pull/2411#issuecomment-256062892

what it would look like
screen shot 2016-10-25 at 12 06 14 pm

### DIFF
--- a/pages/en/lb3/Migrating-to-3.0.md
+++ b/pages/en/lb3/Migrating-to-3.0.md
@@ -31,19 +31,31 @@ We have implemented a new more secure error handler middleware available
 via the package `strong-error-handler` and encourage all uses to switch to
 this new middleware.
 
-We have removed `loopback#errorhandler` short-cut. You should explicitely
-install `errorhandler` (or even better, `strong-error-handler`) in your
-project.
+HTTP error response produced by the new error handler no longer contain
+`error.status` property, as it was renamed to `error.statusCode`. You should
+update any clients reading the HTTP status code from the response body.
 
 The new error handler no longer honours the environment variable
 `NODE_ENV=production` and is configured for production by default.
-You should enable the debug mode in `server/middleware.development.json`.
+You should enable the debug mode in `server/config.development.json`.
 Note that projects scaffolded with a recent version of `apic loopback`
 (or `slc loopback`) are already correctly configured and no change is needed.
 
-HTTP error responsed produced by the new error handler no longer contain
-`error.status` property, as it was renamed to `error.statusCode`. You should
-update any clients reading the HTTP status code from the response body.
+User can specific options in their applications in `config(.development).json` as follow:
+
+```json
+{
+  "restApiRoot": "/api",
+  "host": "0.0.0.0",
+  "port": 3000,
+  "remoting": {
+    "errorHandler": {
+      "debug": true,
+      "log": false
+    }
+  }
+}
+```
 
 ## Middleware getters from Express 3.x were removed
 


davidcheung:
cc @Amir-61

Amir-61 (approved):
LGTM

crandmck:
I chatted with David about this.  I'm working on the RNs and Migration Guide, so it will be easier for me to fold this change into my (extensive) edits rather than merge this.  I will keep this PR open to remind me to include these changes.

crandmck:
Closed in favor of #136.

doublemarked:
I am working on migration from 2.x => 3.1. Why is it acceptable for the error handler to not be configurable via middleware.json? Why is this not a bug? This was raised in strongloop/loopback#2411 and resolved with this documentation about config.json, but it doesn't feel right. At minimum, the migration documentation is misleading. It should not say anything about params in middleware.json, as this configuration is overridden entirely by that in config.json. 

I feel there may be a more appropriate recommendation for configuration. Currently v3 hard fails on boot if we don't add the error-handler to the middleware json, and yet strong-remoting still wants to configure and inject it itself. IMHO, a configuration that is less redundant and more consistent with use of middleware like strong-error-handler is as follows,

1. config.json should have no `errorHandler` sub-section.
2. config.json should have key `remoting.rest.handleErrors` set to `false`, to disable strong-remoting's meddling with the error middleware. See here: https://github.com/strongloop/strong-remoting/blob/master/lib/rest-adapter.js#L320-L322
3. Params for strong-error-handler should exist only in middleware config. Environment-specific error handler behavior should be controlled using env-specific middleware files (as was done in v2).

crandmck:
@bajtos @loay What do you think of the suggestion above?

bajtos:
Thank you @doublemarked for the valuable feedback.

Your proposal makes total sense to me, in fact I believe that's exactly the configuration scaffolded by recent versions of `slc loopback` - see [config.json template](https://github.com/strongloop/loopback-workspace/blob/e4c23d40414fc06a0e9137f0f6e224555e2f502a/templates/projects/empty-server/data.js#L67-L82) and [middleware.development.json template](https://github.com/strongloop/loopback-workspace/blob/e4c23d40414fc06a0e9137f0f6e224555e2f502a/templates/projects/empty-server/files/server/middleware.development.json#L3-L7).

I think the difficult part is how to convert this recommendation into a list of steps for the migration guide? 👈 @crandmck 

BTW when copying GitHub URLs, please press `y` in your page - that way GitHub converts the URL to a permalink containing SHA of the current HEAD (e.g. https://github.com/strongloop/strong-remoting/blob/f8129e9eeecd1a4f3e92385e3cdf90acaeddfb66/lib/rest-adapter.js). Without that, the link points to the latest `master` and line numbers get out of sync with the actual code in the file.

doublemarked:
Happy New Year, @bajtos.

I saw this `handleErrors` key in the workspace template's config.json, but I believe it's broken as it belongs under the `rest` key? At least it didn't work for me as-is. I was going to raise that issue after confirming here that this is a kosher v3 config. If you confirm that this is in error I can put together a patch, though I wonder what to do about already scaffolded v3 apps.

Good point regarding URLs! TYVM!

bajtos:
@doublemarked Happy New Year to you too! 🎉 

You are absolutely right that we have a bug in our workspace template. I have sent a pull request to fix it, see https://github.com/strongloop/loopback-workspace/pull/372.

> I wonder what to do about already scaffolded v3 apps.

I am afraid there isn't much we can do about apps already scaffolded using a broken template. That's the trouble with scaffolding - it's difficult to ship bug fixes and improvements to generated code :(

I think we should add a note to the documentation advising users how to detect whether their application is affected by this issue and what changes to make to fix the problem. Possibly also leave a message in the mailing list.

Thoughts?

doublemarked:
I agree, docs. Including updates to the v3 release notes and migration guide to describe the new correct config with the error handler.

crandmck:
@bajtos @loay ping...

bajtos:
@doublemarked would you mind contributing the documentation changes yourself? In my experience, the documentation written by users is often the best, as the users know best what they would expect to see in the docs.

PR #140, opened 2016-10-28
Add four community projects and update instructions

crandmck:
@jannyHou This PR does the following:
- Adds four community projects: Elasticsearch connector, LoopBack SDK Builder, JSON Schemas, and 
  IoT demo
- I used the technique of simply pulling in the project READMEs, described in http://loopback.io/doc/en/contrib/Including-READMEs.html.
- Updated the instructions so community members can do the same.  I left them the option of creating a custom project page or "reusing" their project README.

If you want to see how this will look when it's merged, follow the instructions in the README to run Jekyll locally.

### DIFF
--- a/pages/en/community/How-to-add-a-project.md
+++ b/pages/en/community/How-to-add-a-project.md
@@ -9,27 +9,89 @@ permalink: /doc/en/community/How-to-add-a-project.html
 
 To add your community project here, follow these steps:
 
-1. Fork the [strongloop/loopback.io repository](https://github.com/strongloop/loopback.io.git) (see [Fork a repo](https://help.github.com/articles/fork-a-repo/) for instructions).
+First, fork the [strongloop/loopback.io repository](https://github.com/strongloop/loopback.io.git) (see [Fork a repo](https://help.github.com/articles/fork-a-repo/) for instructions).
+
+Now, you have two options:
+
+- If your project repository has a good, detailed README, you can reuse it.  Follow the steps in [Reuse your README](#reuse-your-readme).
+- If your project doesn't have a complete README, you can create a custom page.  Follow the steps in [Create a custom project page](#create-a-custom-page).
+
+## Reuse your README
+
+This site has tools and processes for incorporating README files from external repositories.
+For a full description, see [Including READMEs from other repositories](/doc/en/contrib/Including-READMEs.html).  In brief, we use a script to pull down READMEs over the network, then incorporate them into the site.  Currently, to update the READMEs, we periodically run the script manually.
+
+To reuse your project README, follow these steps:
+
+1. Copy one of the existing project pages that uses a README, for example  `loopback.io/pages/en/community/IoT-demo.md`.
+  - Save it into the same directory (`pages/en/community`), and name it based on your project name, for example `pages/en/community/My-awesome-project.md`.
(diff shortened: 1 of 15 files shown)


pages/en/community/How-to-add-a-project.md
```diff
@@ -9,27 +9,89 @@ permalink: /doc/en/community/How-to-add-a-project.html
...
+
+1. Copy one of the existing project pages that uses a README, for example  `loopback.io/pages/en/community/IoT-demo.md`.
+  - Save it into the same directory (`pages/en/community`), and name it based on your project name, for example `pages/en/community/My-awesome-project.md`.
+1. Edit the front matter at the top of the file (see below):
+  - Change the `title` property to be the name of your project.
+  - Change the `source` property to the name of your repository; in the example below `source: StrongLoop-IoT-Demo`.
+  - Change the `org` property to the name of the organization that contains the repository; for example `org: my-org`.
+  - Change the `permalink` property to match the name of the file, for example `/doc/en/community/My-awesome-project.html`.
+  - Add any `keywords` you want for search-engine optimization (optional).
+  - If you don't want an automatically-generated page TOC, add `toc: false`.
```
jannyHou, line 34: Shall we remind users to change `Summary` here as well?

pages/en/community/How-to-add-a-project.md
```diff
@@ -9,27 +9,89 @@ permalink: /doc/en/community/How-to-add-a-project.html
...
+npm install
+cd ..
+./update-community-readmes.sh
+```
+
+## Create a custom page
+
+If you don't want to reuse your README for any reason, follow these steps:
+
 1. Copy the `loopback.io/pages/en/community/Sample-project.md` file, save it into the same directory (`pages/en/community`), and name it based on your project name, for example `pages/en/community/My-awesome-project.md`.
```
jannyHou, line 87: I think we can only keep the front matter part in `loopback.io/pages/en/community/Sample-project.md` and remove other contents starting from [Link](https://github.com/strongloop/loopback.io/blob/gh-pages/pages/en/community/Sample-project.md#link). At first I thought community projects would be something not using LB but good for LB developers to read and learn, so the sift.js sample I put there actually doesn't fit our purpose now.

Btw, maybe add a short summary in the front matter, eg: 

```
summary: This is a sample project with front matter template.
```

jannyHou (approved):
@crandmck Thank you for your samples and process LGTM. 
Only two minor comments.

jannyHou (commented):
@crandmck And I just find that the lint fails due to long commit msg:

```
 99ea9ef - Add four community projects and update instructions: First line should be 50 characters or less (saw 51)
```

LGTM after fixing it.

pages/en/community/How-to-add-a-project.md
crandmck: I will remove the stuff about sift.js and the examples, but the general outline of a "custom page" seems reasonable.  Is there something specific in the outline that you object to?

crandmck:
@jannyHou How do I change the commit message?

pages/en/community/How-to-add-a-project.md
jannyHou: @crandmck nope  

> the general outline of a "custom page"

Also reasonable to me 👍

PR #141, opened 2016-10-29
create a Jenkins pipeline for updating readmes

rmg:
Uses a simple shell script to fetch all the specified project README.md
files and then commit the updated files if there are any changes.

@crandmck I didn't look at what else uses the `_data/repos-*.json` files. If they are used then I can adjust the script to use them as the source of truth. This was just a first pass to show the concept. Let me know what you think.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,36 @@
+#!groovy
+
+def gitEnv = [
+  'GIT_COMMITTER_NAME=StrongLoop Bot',
+  'GIT_COMMITTER_EMAIL=slnode@ca.ibm.com',
+  'GIT_AUTHOR_NAME=StrongLoop Bot',
+  'GIT_AUTHOR_EMAIL=slnode@ca.ibm.com',
+]
+
+node('linux && git') {
+  stage('checkout') {
+    checkout scm
+  }
+  stage('download updates') {
+    sh './update-readmes.sh'
+  }
+  stage('check updates') {
+     sh 'git status'
+  }
+  stage('commit updates') {
+    if (sh(returnStatus: true, script: 'git diff --quiet') != 0) {
+      echo "Updates found, committing and pushing back up to github..."
+      // this will add new files if the readme list grows, but removing
+      // existing files will have to be done by a human for now.
+      sh 'git add pages/en/lb2/readmes'
+      withEnv(gitEnv) {
+        sh 'git commit -m "Update readmes from loopback repos"'
+      }
+      sshagent(credentials: ['loopback-io-deploy']) {
+        sh 'git push origin HEAD:$BRANCH_NAME'
+      }
+    } else {
+      echo "No updates to commit"
+    }
+  }
+}
--- a/_config.yml
+++ b/_config.yml
@@ -29,6 +29,8 @@ exclude:
   - .idea/
   - .gitignore
   - node_modules
+  - Jenkinsfile
+  - update-readmes.sh
 # these are the files and directories that jekyll will exclude from the build
 
 feedback_email: rmckinn@us.ibm.com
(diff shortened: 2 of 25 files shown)


crandmck:
Can you please explain the purpose of this?   Is the script automatically run?  If so, when or under what condition?  What is the advantage of this over using `get-readmes`?

crandmck:
Also please remove the commits to the README files in the `pages/en/lb2/readmes` because they undo temporary changes to these files that fix old doc links and make the READMEs layout properly in the docs.  These changes were made in docs repo but still need to be committed to the appropriate repos` READMEs.  In the long run of course, the READMEs will be single-sourced, but we have not yet committed these changes and the changes in this PR clobber the "fixes" I made to these files.  We'll address this difference soon.

rmg:
The commit that updates the README's is the goal of this change. This PR adds automation that will automatically run the readme updating and commit it back and that commit is the result of that automation being run on this branch.

crandmck:
>  The commit that updates the README's is the goal of this change.

OK, then we won't merge this until the READMEs are updated in the repos because that will be a regression.

The `get-readmes` script determines each repo's latest release and gets the README for that unless a different branch is specified (fallback is to master if there is no tagged release).  Can you add that to your script?

pages/en/lb2/readmes/loopback-example-connector.md
```diff
@@ -0,0 +1,33 @@
+# loopback-example-connector.soap
```
crandmck: This file is spurious and duplicates loopback-example-connector-soap.md.  Not sure how it got generated, but it should be removed.

crandmck (commented):
BTW, for the release stuff, take a look at what I did in the Express repo: https://github.com/expressjs/expressjs.com/blob/gh-pages/get-readmes.sh.  It's a bash script that finds the latest release.

rmg:
To avoid the complication of looking up the latest release, I switched to just fetching the latest README.md content from npmjs.org, which updates its content whenever a release is published. There is conveniently no throttling and the JSON response is trivially piped through `jq` to extract the one text blob we care about.

For the non-release cases where we are fetching from a branch, the script just pulls from GitHub's static content domain. We can do this because the repos are all public. This interface is also conveniently not throttled, so we can avoid the need for GH credentials entirely.

There was a slight problem where certain README.md's were coming back as `null`, but that ended up being the cases where a default `master` branch actually was the correct behaviour instead of looking for a release. I fixed those by just explicitly listing `master` as the branch.

rmg:
There are 2 complications left to sort out.
- [ ] Jenkins has decided to switch to https for cloning instead of ssh, so it is failing on `git push`. I'll need to fix this somehow.
- [ ] Some of the README.md's are getting trailing blank lines appended to them. We can either run all the markdown through some sort of whitespace normalizing script while downloading or we can just ignore it and see if it ends up causing unwanted commit noise or not.

crandmck:
Great!  Shall we wait to merge this until you've resolved those two items?  Or can we merge it now and sort those out later?

Re the trailing blank lines--that won't be an issue for the docs, but is problematic for another reason (e.g. CI linter)?

rmg:
Turns out the switch from git to https was actually because a different Jenkins server scanned this repo and created its own job for it. That should be cleaned up now.

The blank lines shouldn't cause any problems, but I found a reasonable way of addressing it.

pages/en/lb2/readmes/loopback-example-connector.md
rmg: Figured out part of the problem.. the loopback-example-XXX repos that are using branches to differentiate between modules were all clobbering each other because the readme name was being based only on the repo name.

crandmck:
@rmg I realize now the shortcoming of getting the README for the latest release vs. what's on master....
This PR now includes changes to a bunch of connector READMEs that basically reverts them back to their old/outdated content.  

  We've made a ton of updates to the READMEs we're using in the docs (per https://github.com/strongloop-internal/scrum-loopback/issues/1142) but we would need to republish all the connectors to npm to ensure that the changes get pulled into the docs.  

That makes sense, and I assume we could address it fairly easily by simply republishing the connectors.... BUT I'm not sure that we could do that right now for _all_ the repos for which we need to.

rmg:
(answered on Slack, but repeating here to close the loop)

Controlling whether a README.md is fetched from the latest release on npmjs.org vs the master branch in a git repo is a simple matter of including a branch name in the repo list in `update-readmes.sh`.

**update** I've expanded the comment in the script to make this more clear.

PR #146, opened 2016-11-08
Update doc for include filter

jannyHou:
Connect to https://github.com/strongloop/loopback#2922

Scenario:
Take example of `Post.find({include: {owner: 'orders'}}, function() { /* ... */ });`

The `orders` has to be a relation name but not property name.


jannyHou:
@crandmck PTAL thanks :)

crandmck:
LGTM - Thanks!

PR #147, opened 2016-11-09
Fix CORS migration to LB 3.0

bajtos:
Add a missing step to disable the built-in CORS middleware in `server/config.json`.

@crandmck please review

@elkorep I think this was the missing step you encountered while upgrading your older LB 2.x app.


elkorep:
@bajtos yes this is the step

crandmck:
LGTM

PR #148, opened 2016-11-09
lb3: describe reserved "File" type

bajtos:
See https://github.com/strongloop/loopback/issues/2554#issuecomment-253786224 for background information.

@crandmck PTAL


crandmck:
LGTM

PR #149, opened 2016-11-09
Made inline documentation of example clearer.

Sigfried:
I'm also still confused about line 32, 

    Use `fs.writeFile()` to save the output in `common/models/model-name.json`.

What output exactly gets written? The output from the example isn't valid json and doesn't seem to be in the form of model definitions.


slnode:

crandmck:
Addition LGTM.

The JSON files written would be [model JSON file](http://loopback.io/doc/en/lb2/Model-definition-JSON-file.html) files that reflect the database schema.  There should be one model/file per db table. 

 It would be great to have a complete working example here... Arc and API Connect editor can do this for you via GUI, so we haven't focused on documenting how to do it in code.

PR #150, opened 2016-11-10
Add script to get community readmes

crandmck:
In addition to examples, connectors, etc, we also want to pull down READMEs from some community projects.  They have a different destination directory, so I just created a new bash script and added running it to the Jenkins pipeline.

I also deleted the JSON files that I was using previously, since they are no longer needed.

@rmg PTAL


rmg:
In `Jenkinsfile` around line 26 you'll need to make it include the community readmes path in the `git add ...`, otherwise it won't actually commit any of the fetched files.

crandmck:
Thanks for catching!  I fixed, PTAL.

crandmck:
I went ahead and merged this.  Hopefully the rest of the script is OK...

PR #161, opened 2016-11-14
Add new GH tags for `loopback-gettng-started`

Amir-61:
* Add new GH tags for `loopback-getting-started`

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/1151

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/1189

/to: @crandmck


pthieu:
LGTM

PR #162, opened 2016-11-15
lb2: how to disable remoteMethods via config

davidcheung:
screen shot 2016-11-15 at 3 15 29 pm

@raymondfeng @crandmck @gunjpan PTAL


davidcheung:
so currently theres a bug that doesnt hide the relation's method, i dont think it should go into documentation since its not expected behavior,
@crandmck do we put **`Notes:`** as per known bugs?

raymondfeng:
LGTM

pages/en/lb2/Exposing-models-over-REST.md
```diff
@@ -322,7 +322,43 @@ MyModel.disableRemoteMethod('updateAttributes', isStatic);
...
+        "*": false,
+        "login": true,
+        "logout": true
+      }
+    }
+  }
+}
+```
+
+You can also configure all models through `config.json`'s `remoting` object:
```
gunjpan: Can you please change this line to echo following message:
**You can also hide common methods across all models through `config.json`'s `remoting` object as follows:**

pages/en/lb2/Exposing-models-over-REST.md
```diff
@@ -322,7 +322,43 @@ MyModel.disableRemoteMethod('updateAttributes', isStatic);
...
+  "context": false,
+  ...
+  "sharedMethods": {
+    "*": false,
+    "login": true,
+    "logout": true
+  }
+}
+```
+
```
gunjpan, line 360: Please add a `Note:` here stating : Current implementation does not disable methods generated by relations. See filed issue [here](https://github.com/strongloop/loopback/issues/2860)

Please discuss with @crandmck if we should add this note.

gunjpan (changes requested):
Two minor commetns. Otherwise, LGTM.

pages/en/lb2/Exposing-models-over-REST.md
davidcheung: agreed, i think it would be confusing for user if we don't put the note there
screen shot 2016-11-15 at 4 00 24 pm

crandmck:
LGTM, merging!

PR #170, opened 2016-11-18
Document `forceId`

Amir-61:
* Add explanation of `forceId`

Connect to https://github.com/strongloop/loopback/issues/2068

/to: @crandmck @jannyHou @bajtos  PTAL. Thanks!

### DIFF
--- a/pages/en/lb2/Model-definition-JSON-file.md
+++ b/pages/en/lb2/Model-definition-JSON-file.md
@@ -101,6 +101,13 @@ Properties are required unless otherwise designated.
       </td>
     </tr>
     <tr>
+      <td>forceId</td>
+      <td>Boolean</td>
+      <td>
+        The default value is set to true whenever a model uses an auto-generated ID. If it is set to true it prevents users from setting the auto-id value manually.
+      </td>
+    </tr>
+    <tr>
       <td>http.path</td>
       <td>String</td>
       <td>Customized HTTP path for REST endpoints of this model.</td>


jannyHou:
LGTM 
A minor suggestion: maybe it makes more sense if we can change the sentence order to: 

> The default value is set to true whenever a model uses an auto-generated ID. If it is set to true it prevents users from setting the auto-id value manually.

Amir-61:
@jannyHou I initially had it in your suggested order but then I changed it to my order... anyway I do not mind to change it based on your suggestion :-) Thanks for review 🙇

pages/en/lb2/Model-definition-JSON-file.md
```diff
@@ -101,6 +101,13 @@ Properties are required unless otherwise designated.
       </td>
     </tr>
     <tr>
+      <td>forceId</td>
+      <td>Boolean</td>
+      <td>
+        The default value is set to true whenever a model uses an auto-generated ID. If it is set to true it prevents users from setting the auto-id value manually.
```
bajtos, line 107: IIRC, this applies to LoopBack 3.x applications only, doesn't it? I think LoopBack 2.x always defaults `forceId` to `false`.

My point is that this change should be made in `pages/en/lb3/Model-definition-JSON-file.md`, not `pages/en/lb2/Model-definition-JSON-file.md`.

jannyHou: @Amir-61 Sorry I didn't notice that the change is made in `pages/en/lb2/Model-definition-JSON-file.md`, this is a breaking change in lb3.

Amir-61: Thanks @bajtos ! I'm happy that I mentioned you in the PR to catch that https://github.com/strongloop/loopback-datasource-juggler/pull/982 is just made against LB3.x and not back-ported.

@crandmck I made https://github.com/strongloop/loopback.io/pull/178 to rectify default value of `forceId` for LB2.x; in terms of LB3.x, I do not see any pages `pages/en/lb3/Model-definition-JSON-file.md`, what is your plan in this regard? Can we copy & paste `pages/en/lb2/Model-definition-JSON-file.md` to `pages/en/lb3/Model-definition-JSON-file.md`? I believe before copying&pasting, the page needs some modification for LB3.x?

crandmck: Yes, we should copy `Model-definition-JSON-file.md` to the `lb3` folder.  Please do that and make any changes required.

PR #173, opened 2016-11-19
Create Access token REST API.md

qxl1231:


slnode:

crandmck:
@qxl1231 I had to revert this PR, because after I merged it, I noticed several issues:

- You should never translate the keys in the "front matter", the YAML portion at the very top of the file, e.g. 
```
title: "Access token REST API"
lang: en
layout: navgroup
navgroup: access-control
keywords: LoopBack
tags: [authentication, models]
sidebar: lb2_sidebar
permalink: /doc/en/lb2/Access-token-REST-API.html
summary: The built-in access token model represents the access token that LoopBack creates for an authenticated user.
```

Never translate the actual key names, e.g. `keywords`, `tags`, `sidebar`, etc.
The only strings in the front matter that you should translate are the _value_ of the `title` property and the _value_ of the `summary` property

Also, **very important**:
- You _must_ change the language code in the permalink from `en` to `zh`, so: 
```
permalink: /doc/zh/lb2/Access-token-REST-API.html
``` 
Otherwise, this will replace the English article.
 - You must change the value of the `sidebar` property to be the Chinese sidebar:
```
sidebar: zh_lb2_sidebar
```

Please make these changes, and then we can merge this PR.

I'll try to clarify these points in the translation instructions... thank you very much for helping us figure out what needs to be clarified!

crandmck:
Since you haven't responded in over two weeks, I went ahead and made the requisite changes to your change and consolidated all your PRs into #195.  See that for an example of how to translate properly for future reference... :-)

PR #181, opened 2016-11-29
Wrong Data Type in remoteMethod Tutorial

nikolaifischer:
Because the argument is an ID in the getName Method, the datatype has to be a string and not a number.


slnode:

crandmck:
Thanks!

PR #183, opened 2016-11-29
Explain the idType of an embedded model

jannyHou:
Connect to https://github.com/strongloop/loopback-datasource-juggler/issues/750#issuecomment-263564504

> It might be good to add some documentation though to explain where ids in embedded models come from and why they're different from models attached to a persistent data source directly.


jannyHou:
@crandmck PTAL. Thanks.

crandmck:
thx!

PR #187, opened 2016-12-05
Update style guide for ES6

bajtos:
Now that we have dropped support for pre-v4 versions of Node, it's time to start using ES6 goodness in our code base. In this patch, I did:
 - Copy the existing guide to style-guide-es5
 - Add few ES6-specific sections
 - Update all code examples to use the new ES6 syntax

@superkhau @strongloop/loopback-dev PTAL

@sam-github @bnoordhuis Since you are contributing to Node core, which has switched to ES6 long time ago, I think you can help us with avoiding mistakes others have already learnt from. Are any parts of my proposal different from current best practices in Node core? Is there any other ES6-related best practices to add to our style guide?

### DIFF
--- a/pages/en/contrib/style-guide.md
+++ b/pages/en/contrib/style-guide.md
@@ -28,7 +28,7 @@ Style conventions for this document:
 {% include important.html content="
 
 The rules described in this document are evolving, therefore not all of our
-code-base is following them yet. This is expected and OK. Our approach is to
+code-base are following them yet. This is expected and OK. Our approach is to
 incrementally improve the parts that are changing most often, i.e. fix the
 coding style only as part of changes needed for a bug fix/new feature.
 
@@ -36,19 +36,26 @@ coding style only as part of changes needed for a bug fix/new feature.
 
 ### Variable declarations
 
-Prefer to use immutable variables declared with `const` keyword:
+Prefer to use immutable variables declared with `const` keyword.
+
+{% include note.html content="
+
+`const` makes only the reference immutable, the referenced object is still
+mutable.
+
+" %}
 
 ### Good
 
 ```js
-const user = app.models.User;
-const saltWorkFactor = user.settings.saltWorkFactor;
+const User = app.models.User;
+const { saltWorkFactor } = User.settings;
 ```
 
 #### Bad
 
 ```js
-var user = app.models.User;
+var User = app.models.User;
 var saltWorkFactor = user.settings.saltWorkFactor;
 ```
 
@@ -60,7 +67,7 @@ let discount = 0;
 
 if (customer.isPreferred) {
   discount = 0.1;
-} else if (customer.address.state = 'CA') {
+} else if (customer.address.state === 'CA') {
   discount = 0.05;
 }
 ```
@@ -77,7 +84,7 @@ customer.prototype.getDiscount = function() {
     return 0.1;
   }
 
-  if (customer.address.state = 'CA') {
+  if (customer.address.state === 'CA') {
     return 0.05;
   }
 
@@ -85,19 +92,40 @@ customer.prototype.getDiscount = function() {
 }
 ```
 
+{% include important.html content="
+
+`let` currently has a small caveat in that using it in loop constructs causes
+V8 to deoptimize the function, i.e., for now you should write:
+
+```js
+for (var i = 0; i < 42; ++i) doit(i);
+```
+
+Instead of:
+
+```js
+for (let i = 0; i < 42; ++i) doit(i);
+```
+
+This will be fixed in node.js v8.x but probably won't be fixed in v4.x and v6.x.
+
+" %}
+
 ### Arrow functions
 
-The are two considerations to keep in mind when deciding whether to use an
+There are two considerations to keep in mind when deciding whether to use an
 arrow function or a regular function:
 
  - Arrow function preserves `this` from the outer scope. This is much more
-   perfomant than the usual workaround of storing `this` in a `self` variable.
+   performant than the usual workaround of storing `this` in a `self` variable.
 
  - If you are accessing `this` provided by Mocha to your test cases and
    before/after hooks, then you cannot use an arrow function.
 
- - Arrow functions are anonymous, there is no easy way for assigning them a
-   descriptive name. This makes it difficult to understand stack traces.
+ - Arrow functions are anonymous by default, which makes it difficult to
+   understand stack traces. The workaround is to assign an arrow function
+   to a variable, since V8 does a fairly good job of inferring a name
+   in such case.
 
 The rules to follow:
 
@@ -141,7 +169,7 @@ class Foo {
 
 ### Classes
 
-ES6 introduced syntax-sugar for defining classes, start using its' advantages
+ES6 introduced syntax-sugar for defining classes, start using its advantages
 over old `require('util').inherits` approach.
 
 #### Good
--- a/pages/en/contrib/style-guide-es5.md
+++ b/pages/en/contrib/style-guide-es5.md
@@ -1,12 +1,12 @@
 ---
-title: Code style guide
+title: Code style guide (2.x/ES5)
 toc: false
 tags: [contributing, community]
 keywords:
 last_updated: Sept 27, 2016
-summary: "These are LoopBack's general coding style guidelines."
+summary: "These are LoopBack's general coding style guidelines for pre-ES6 branches, e.g. 2.x release lines."
 sidebar: contrib_sidebar
-permalink: /doc/en/contrib/style-guide.html
+permalink: /doc/en/contrib/style-guide-es5.html
 ---
 <!--
 Style conventions for this document:
@@ -15,8 +15,18 @@ Style conventions for this document:
  - level3 headings are reserved to subsections within rules, e.g. `#### Exceptions`
 -->
 
+{% include important.html content="
+
+This document describes the coding style we were using in EcmaScript5 code base
+before upgrading to EcmaScript6 in LoopBack 3.0. Follow these rules when
+back-porting bug fixes to older versions.
+
+See the [current style guide](style-guide.html) for up-to-date conventions
+used for new development.
+
+" %}
+
 {% include see-also.html content="
-- [ES5 style guide for LoopBack 2.x](style-guide-es5.html)
 - [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html)
 - [Google Javascript Style Guide](https://google.github.io/styleguide/javascriptguide.xml)
 " %}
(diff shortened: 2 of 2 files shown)


pages/en/contrib/style-guide.md
```diff
@@ -24,6 +25,152 @@ Style conventions for this document:
...
+
+In the rare cases where you need to split variable declaration and
+initialization, prefer to use `let` over `var`:
+
+```js
+let discount = 0;
+
+if (customer.isPreferred) {
+  discount = 0.1;
+} else if (customer.address.state = 'CA') {
```
bnoordhuis: `==` or `===`?

pages/en/contrib/style-guide.md
```diff
@@ -24,6 +25,152 @@ Style conventions for this document:
...
+```js
+const discount = customer.getDiscount();
+
+// Inside Customer model:
+customer.prototype.getDiscount = function() {
+  if (customer.isPreferred) {
+    return 0.1;
+  }
+
+  if (customer.address.state = 'CA') {
```
bnoordhuis: Likewise.

pages/en/contrib/style-guide.md
```diff
@@ -24,6 +25,152 @@ Style conventions for this document:
...
+}
+```
+
+### Arrow functions
+
+The are two considerations to keep in mind when deciding whether to use an
+arrow function or a regular function:
+
+ - Arrow function preserves `this` from the outer scope. This is much more
+   perfomant than the usual workaround of storing `this` in a `self` variable.
```
bnoordhuis: Typo: performant.  I personally hate that word though. :-)

pages/en/contrib/style-guide.md
```diff
@@ -24,6 +25,152 @@ Style conventions for this document:
...
+      const result = self.processData(data);
+      cb(null, result);
+    });
+  }
+}
+```
+
+### Classes
+
+ES6 introduced syntax-sugar for defining classes, start using its' advantages
```
bnoordhuis: Typo: its, no apostrophe.

pages/en/contrib/style-guide.md
```diff
@@ -24,6 +25,152 @@ Style conventions for this document:
 
 ## General guidelines
 
+{% include important.html content="
+
+The rules described in this document are evolving, therefore not all of our
+code-base is following them yet. This is expected and OK. Our approach is to
```
bnoordhuis: s/is following/are following/

pages/en/contrib/style-guide.md
```diff
@@ -24,6 +25,152 @@ Style conventions for this document:
...
+incrementally improve the parts that are changing most often, i.e. fix the
+coding style only as part of changes needed for a bug fix/new feature.
+
+" %}
+
+### Variable declarations
+
+Prefer to use immutable variables declared with `const` keyword:
```
bnoordhuis: Might be worth pointing out that `const` makes the reference immutable but not the referenced object.  It's something that seems to trip up a lot of people.

pages/en/contrib/style-guide.md
```diff
@@ -24,6 +25,152 @@ Style conventions for this document:
...
+
+### Good
+
+```js
+const user = app.models.User;
+const saltWorkFactor = user.settings.saltWorkFactor;
```
bnoordhuis: Arguably even better: `const { saltWorkFactor } = user.settings;`

Aside: I believe `user` should be `User`?

pages/en/contrib/style-guide.md
```diff
@@ -24,6 +25,152 @@ Style conventions for this document:
...
+
+#### Bad
+
+```js
+var user = app.models.User;
+var saltWorkFactor = user.settings.saltWorkFactor;
+```
+
+In the rare cases where you need to split variable declaration and
+initialization, prefer to use `let` over `var`:
```
bnoordhuis, line 63: `let` currently has a small caveat in that using it in loop constructs causes V8 to deoptimize the function, i.e., for now you should write:
```js
for (var i = 0; i < 42; ++i) doit(i);
```
Instead of:
```js
for (let i = 0; i < 42; ++i) doit(i);
```
This will be fixed in node.js v8.x but probably won't be fixed in v4.x and v6.x.

pages/en/contrib/style-guide.md
```diff
@@ -24,6 +25,152 @@ Style conventions for this document:
...
+    return 0.05;
+  }
+
+  return 0;
+}
+```
+
+### Arrow functions
+
+The are two considerations to keep in mind when deciding whether to use an
```
bnoordhuis: There

pages/en/contrib/style-guide.md
```diff
@@ -24,6 +25,152 @@ Style conventions for this document:
...
+arrow function or a regular function:
+
+ - Arrow function preserves `this` from the outer scope. This is much more
+   perfomant than the usual workaround of storing `this` in a `self` variable.
+
+ - If you are accessing `this` provided by Mocha to your test cases and
+   before/after hooks, then you cannot use an arrow function.
+
+ - Arrow functions are anonymous, there is no easy way for assigning them a
+   descriptive name. This makes it difficult to understand stack traces.
```
bnoordhuis: Half-true.  V8 does a fairly good job of inferring a name when you assign them to variables.

bajtos:
@bnoordhuis lovely ❤️ , thank you for your feedback! I applied all your comments in 0106dea, would you like to take another look?

bnoordhuis (approved):

crandmck:
Would you also like me to add this to the "Contributing to LB" navigation - http://loopback.io/doc/en/contrib/code-contrib.html ?  Easy enough to do - just LMK.

superkhau:
Great feedback, thanks @bnoordhuis and @bajtos for updating. LGTM.

bajtos:
> Would you also like me to add this to the "Contributing to LB" navigation - http://loopback.io/doc/en/contrib/code-contrib.html ? Easy enough to do - just LMK.

The "code-contrib.html" already contains a link to "style-guide.md" AFAICT, see the bullet point "Adhere to code style outlined in the style guide.". (I updated an existing file in this patch, so that's expected.)

Perhaps you are asking whether to add "style-guide-es5.md" to that document? I think that's not necessary, since most back-porting work is done by us, not the community.

@superkhau thoughts?

pages/en/contrib/style-guide.md
```diff
@@ -24,6 +25,180 @@ Style conventions for this document:
...
+
+#### Good
+
+```js
+class MyConnector extends BaseConnector {
+  constructor(settings, dataSource) {
+    // ...
+  }
+
+ set(modelName, key, value, options, callback) {
```
eddiemonge, line 183: indentation is off here

superkhau:
>Perhaps you are asking whether to add "style-guide-es5.md" to that document? I think that's not necessary, since most back-porting work is done by us, not the community.

I agree, we don't want more burden of maintaining ES5 docs in addition to ES6. That said though, is there a way to "freeze" the old ES5 docs with LB2 @crandmck? We won't maintain the old one, but you can still reference it up to this point. We will only maintain the latest docs and not backport to the old docs.

crandmck:
> whether to add "style-guide-es5.md" to that document? I think that's not necessary, since most back-porting work is done by us, not the community.

Oh, I see, the new document replaces the old one, and you're saying that the community won't generally need the old (ES5) one.  OK.

> is there a way to "freeze" the old ES5 docs with LB2 ?

AFAIK, the only way to freeze it is to not accept any PRs to change it.  We could also add a warning at the top of the page to say that it's specific to 2.x, and is "frozen", etc.  Always need to consider that someone can land on the page (without context) from a search.

superkhau:
>AFAIK, the only way to freeze it is to not accept any PRs to change it. We could also add a warning at the top of the page to say that it's specific to 2.x, and is "frozen", etc. Always need to consider that someone can land on the page (without context) from a search.

I meant make a second copy of the .md file and add it to loopback.io as historic docs. You can git checkout to the point in time before the updates happened and copy the file directly.

superkhau:
Maybe add a footnote somewhere in the new doc that points to the old docs? Ie. If you want to know the ES5 standard we followed before and not maintaining now, see link-here-to-copied-md-file. @crandmck Thoughts?

crandmck:
> make a second copy of the .md file ...

What's the point of that?  It's already at http://loopback.io/doc/en/contrib/style-guide-es5.html.  The source file will of course live in GH as well.

> Maybe add a footnote somewhere in the new doc that points to the old docs?

Already there in the "See also" box at the top-right of http://loopback.io/doc/en/contrib/style-guide.html.

superkhau:
>What's the point of that? It's already at http://loopback.io/doc/en/contrib/style-guide-es5.html. The source file will of course live in GH as well.

You're right, there is no point when I think about it more clearly now. The correct approach is probably to link directly to the commit before the new changes.

1. We keep a reference to the pre-ES6 docs linked from loopback.io as required for users who may want to submit backports (or for us to reference easily)
2. We also get to keep the latest ES6 conventions in the current loopback.io docs

Thoughts?

bajtos:
@superkhau The current state of things is deliberate. There is a current style guide in `style-guide.{md,html}` that will keep evolving over time. Then there is a snapshot of the style guide that was in effect at the time LoopBack 2.x was in development.  I find it useful to have this old document available on the website, and also have the ability to change it if there ever becomes a need for that. (For example, add a section explaining how to backport changes from ES6 codebase to ES5 codebase.)

> We keep a reference to the pre-ES6 docs linked from loopback.io as required for users who may want to submit backports (or for us to reference easily)

We already have that now.

> We also get to keep the latest ES6 conventions in the current loopback.io docs

Ditto, we have this too.

I guess I don't understand what are the benefits of your proposal, compared to the current state? Also, is this discussion worth our time?

pages/en/contrib/style-guide.md
bajtos: Good catch, fixed in bf65d6ba

superkhau:
@bajtos You're right, I just took a closer look at loopback.io. It already looks correct the way it is -- ES6 style guide promoted and ES5 doc in a referenceable link, both updateable if needs be.

> I find it useful to have this old document available on the website, and also have the ability to change it if there ever becomes a need for that. (For example, add a section explaining how to backport changes from ES6 codebase to ES5 codebase.)

I was just asking if there was a way to "freeze"/keep a copy of the ES5 docs around so we can have both available and backport if necessary. Since we're already there, this conversation is done.

mnlbox:
Is it available now? I can't find anything with ES6 in Loopback 3 documentation :disappointed:

bajtos:
> Is it available now? I can't dind anything with ES6 in Loopback 3 documentation 😞

This pull request is adding/modifying the style guide for people contributing changes to Loopback, it's not primarily aimed at people writing applications using LoopBack as the framework.

The content of this guide is available here: http://loopback.io/doc/en/contrib/style-guide.html

PR #189, opened 2016-12-06
Updated model generator slc command

ghost:
To match the command located at; http://loopback.io/doc/en/lb2/Command-line-tools.html


slnode:

crandmck:
Thanks!  I think that was result of search/replace error... :-(

PR #193, opened 2016-12-06
Formatting and cleanup

crandmck:
@superkhau 
Apologies for not catching this in the original PR.  There are **no** content changes, just formatting and the like:
- The "Good" and "Bad" items were inconsistently styled.  Changed to always be bold text (not headings)
- Fixed page TOC so it doesn't show 3rd-level headings or lower.
- Made the "Exception" items into "Notes"
- Fixed code indenting that Eddie noted in the previous PR after it was merged.
- Other misc clean-up.


superkhau (approved):
Can you rebase before landing? PR linter is complaining.

PR #194, opened 2016-12-07
Update Use-API-Explorer.md

emckean:
straight quote in example was being magically curlified to curly left double quote, replaced with html entity (\&quot;) to make sure it stays straight


slnode:

PR #199, opened 2016-12-09
Cert based authentication addition

tjtyrrell:
Since you extend the pg plugin, that plugin supports cert based authentication to Postgres. Took a while for me to figure this out, so I thought I'd help other people if they ever run across the same issue.


slnode:

crandmck:
Thanks!

PR #201, opened 2016-12-14
Update functional-area-owners.md

superkhau:
@crandmck PTAL

### DIFF
--- a/pages/en/contrib/functional-area-owners.md
+++ b/pages/en/contrib/functional-area-owners.md
@@ -58,7 +58,6 @@ See the actual content below for detailed examples.
 - Repos
   - [loopback-boot](https://github.com/strongloop/loopback-boot)
   - [loopback-example-app-logic](https://github.com/strongloop/loopback-example-app-logic)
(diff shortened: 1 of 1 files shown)


superkhau:
cc @ritch @cgole

crandmck:
This change in and of itself is fine, but I need to understand what's happening to these example repos that you removed from the list.  The docs currently use some of these READMEs (see https://github.com/strongloop/loopback.io/blob/gh-pages/update-readmes.sh) so if they will no longer be maintained we should remove them.  But I wonder if we have replacement examples?

superkhau:
>But I wonder if we have replacement examples?

Nope, we should move most of the content to loopback.io if possible -- they will remain in strongloop-community for reference though (I'm trying to get us to take advantage of strongloop-community more instead of using it as a dumping ground for dead projects).

crandmck:
OK, I think I misunderstood.  These repos are not going away, but rather just got moved into strongloop-community org, right?  If so, we can continue to use the READMEs as we do, but just change the location where we get them from.

PR #202, opened 2016-12-14
Add preferred team style for cb moved to the next line

Amir-61:
In order to avoid inconsistency, this PR documents the preferred style guide team prefers/follows when the callback needs to go to the next line due to line length exceeding max line length defined by eslint: 

I listed all reasonable choices here, but since **option A** is the one which team has followed mostly I chose it as a good example to avoid inconsistency; what we do care is consistency; let's decide which one should be followed by LoopBack team. If you do not have objection, please lets stick to option **A**.

**A:**
```
      it('my long test description ...',
      function(done) {
        ...
      });
```
**B:**

```
      it('my long test description ...’, 
        function(done) {
          …
        });
```

**C:**

```
      it(‘my long test description ...’, 
        function(done) {
          …
        }
      );
```

/to: @strongloop/loopback-dev @superkhau PTAL, Thanks!

### DIFF
--- a/pages/en/contrib/style-guide.md
+++ b/pages/en/contrib/style-guide.md
@@ -578,6 +578,37 @@ describe('my class', () => {
   }
 });
 ```
+### Callback function moved to the next line
+
+The following examples show the preferred style when callback needs to move to the next line due to line length exceeding max line length defined by eslint:
+
+**Good**:
+
+```js
+it('my long test description ...',
+function(done) {
+  ...
+});
+```
+
+**Bad**:
+
+```js
+it('my long test description ...', 
+  function(done) {
+    …
+  });
+```
+
+**Bad**:
+
+```js
+it('my long test description ...', 
+  function(done) {
+    …
+  }
+);
+```
 
 ### Test naming
 


eddiemonge:
:( I prefer C as its clear what the ending tags are closing. I thought this was already documented somewhere though

superkhau:
@eddiemonge We're already using A on LB projects. If you want to propose a change, feel free to submit a PR and get team buy-in.

Amir-61:
@eddiemonge Yea to me all of the mentioned styles are clear; considering the fact that we need to have a team style and since we mostly used **style A** in LoopBack projects, it makes sense to have **style A** as the accepted one; however, I'm OK to change it to **style C** or any other style as the accepted style too as long as team prefers it and follows it, but other styles may need more work to change the previously used style (`style A`); with that being said please feel free to open a PR to change it and team can discuss there :-) My goal was to have consistency in terms of style :-)

bajtos:
@strongloop/loopback-dev we should really wait at least 1 business day before landing pull requests in the "contrib" section, to allow all of us to catch up and comment.

I believe we use both A and B styles in our code base. We started to use A more than B because it produces smaller diffs when fixing `max-len` violations in an existing code. AFAIK, we have never used C.

pages/en/contrib/style-guide.md
```diff
@@ -578,6 +578,37 @@ describe('my class', () => {
...
 ```
+### Callback function moved to the next line
+
+The following examples show the preferred style when callback needs to move to the next line due to line length exceeding max line length defined by eslint:
+
+**Good**:
+
+```js
+it('my long test description ...',
+function(done) {
```
bajtos, line 589: Since we are using ES6, this should be probably

```
it('my long test description ...',
() => {
  ...
});
```

(Same for the other two examples below.)

superkhau: Gotta switch to ES6 mindset still. ;)

PR #204, opened 2016-12-15
Update Authentication-authorization-and-permissions.md

haru01:
Readonly expect updateAttributes true, but false.


slnode:

PR #205, opened 2016-12-15
Update Polymorphic-relations.md

zlm919:


slnode:

crandmck:
@jannyHou Can you review this please?  The other examples seem to have the `model` key, so not sure why it wouldn't apply here?

jannyHou:
@crandmck  IMO the `model` will be ignored if `polymorphic` exists, see code https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/relation-definition.js#L1221-L1222

So it's reasonable to remove it. `polymorphic` means it will lookup the related model dynamically.
Our test case in juggler also only specify the `relationName`: https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/relations.test.js#L1693

jannyHou:
@zlm919 The change LGTM, can you sign the contribution agreement? Thanks. Then we can land this patch after CI check green.

PR #210, opened 2016-12-20
Describe our LongTermSupport plan

bajtos:
@crandmck please review. The sidebar config does not seem to work correctly for me (it does not show the new item on home page), do I need to clean any caches somehow?

@chandadharap @cgole you may want to take a look too


crandmck:
The way this is set up now, this article will appear _only_ on the home page navigation (not LB2, LB3, or Contributing navs).  I don't think this is the best way to do it, because (e.g.) someone going directly to a page in LB2 or LB3 docs will not see it. 

I think it should be in the "Contributing" nav, where all the other project-related info is.... And then  link to it from the LB2, LB3 home pages, etc.

OR we could put it in all 3 navs (via an include, so it's single-sourced).  
@raymondfeng @ritch

crandmck:
Also, perhaps we should note that we're following semver, with a link to same--so someone can infer the details of our versioning...?

ritch:
> I think it should be in the "Contributing" nav, where all the other project-related info is.... And then link to it from the LB2, LB3 home pages, etc.

Sounds fine for now...? So we can merge this and link to it from the blog post.

crandmck:
@ritch I made the changes based on my suggestion above...

PR #213, opened 2016-12-21
Add missing left bracket in the example

pauloalem:
There's missing left bracket missing in the order by two properties example


slnode:

crandmck:
Thanks!

PR #216, opened 2016-12-22
Update Repo Owners by Squad

0candy:


raymondfeng (changes requested):
Thank you for the PR. Were there any formal discussions around the new ownership? I don’t think I had the full visibility.

+1 to organize them by squads. But I still prefer to list SMEs for individual modules/repos if possible.

Let’s revisit in the new year when everyone’s back. It’s probably needs to a tribe-level discussion that involves squad leads/managers/architect.

superkhau:
@raymondfeng Yes, there was a squad leads meeting to come up with some agreement and ended up with the list at at https://ibmapim.slack.com/archives/lb-squad-leads/p1481923799000044 -- which is what I'm assuming @candy based these updates on.

Before the meeting, the squad leads got feedback from their own individual members/team's preferences on which repos they want to own as a squad. During the meeting, we simply grouped/shuffled a few repos around (because many repos happened to line up with squad preferences list by chance).

> It’s probably needs to a tribe-level discussion that involves squad leads/managers/architect.

I agree (you/@chanda + managers should've been in that meeting too), good to finish getting buy-in from all stakeholders before landing this -- let's have another meeting when everyone is back.

cc @pthieu @kjdelisle @ritch

bajtos:
I would personally like to stay more involved in strong-remoting work and therefore prefer to move the ownership of this area to squad Asteroid. Having said that, I don't know all the concerns and constraints that are involved in these decisions. If there are good reasons why strong-remoting should not be owned by Asteroid, then I am ok with that.

superkhau:
>Having said that, I don't know all the concerns and constraints that are involved in these decisions. If there are good reasons why strong-remoting should not be owned by Asteroid, then I am ok with that.

It mostly came down to freeing up our time (me/u/ritch/raymond) to work on LB4 -- @ritch would know where strong-remoting stands there though.

@ritch ^

cgole:
Once everyone (Chanda, Raymond) is back from Holidays, we'll revisit this.

crandmck:
@cgole What's status of this?

crandmck:
Ownership of https://github.com/strongloop/strong-error-handler is missing from this document.

bajtos:
https://github.com/strongloop/strong-docs should be added too, I think it should be owned by @hacksparrow (or alternatively the Asteroid squad).

hacksparrow:
@bajtos according to the discussion I had with @ritch yesterday, I would be owning it.

crandmck:
Presumably, @hacksparrow would also be owner of https://github.com/strongloop-internal/apidocs.strongloop.com (which needs to be made public and moved to strongloop org btw).

crandmck:
@superkhau and I discussed reformatting this into a table to make it easier to read.  Here's an example https://github.com/strongloop/loopback.io/wiki/Demo-of-FA-Owners-table (except we'd do it in HTML so we could set the column widths to avoid too much wrapping).

It also occurred to me that to make it easier for any tools to parse in the future, I could put the actual data into a YAML file, which Jekyll can easily layout into table format.  That would give us the best of both worlds: data that is easily machine-parseable and also easily human readable. Thoughts?

superkhau:
@crandmck :clap:, +1 for yaml

crandmck:
As a "weekend project," I implemented the page with table layout using YAML data in https://github.com/strongloop/loopback.io/tree/gh-pages/_data/fa-owners.yml.
I included both team/squad owners from @0candy and SMEs as @raymondfeng  suggested.  I suspect there will be further tweaks / iterations on the info.

Unfortunately, to see the actual table layout, you'll need to run Jekyll, so for convenience I took a screenshot to give a general idea of what it looks like: https://screencast.com/t/KJjwsiGB2NnZ.

_data/fa-owners.yml
```diff
@@ -0,0 +1,353 @@
...
+'Boot':
+  - area: Boot
+    label: boot
+    owner:
+    - sq-lb-asteroid
+    repos:
+    - loopback-boot
+    - loopback-example-app-logic
+    smes:
+    - davidcheung
```
bajtos, line 20: Can we either use a link pointing to GH profile (`https://github.com/davidcheung`), or at least prepend `@` sign and use bold font (`**@bajtos**`) to make it more obvious that this value is a GH handle?

bajtos (commented):
+1 for YAML

_data/fa-owners.yml
crandmck: In the table as displayed, all of these are rendered as links: GitHub IDs, repos, and teams.  
The links are added in https://github.com/strongloop/loopback.io/pull/216/files#diff-fab19d40a489a9c5ba42e6dfe95b57b2.  The file references are also links, where the URL is provided in the YAML.

If you run the site yourself, you'll see that, but you can get an idea from the screenshot I posted... https://www.screencast.com/t/KJjwsiGB2NnZ

superkhau: I'm also siding with @bajtos, is it not possible to render the link text to show up as @user? So instead of "loay" in your pic, it shows "@loay" instead?

crandmck: It's trivial to add a `@` but the text is **already linked**.  The text is styled as a link and has a mouseover "hint" of the URL, as you would expect.  Adding `@` serves no purpose.

superkhau: The intent is sugar to show it's a handle type link without the mouseover -- ie. superkhau vs @superkhau, which tells us that is our handle (for GitHub, Twitter, etc), even if they decide not to click on it.

crandmck: FINE, whatever!  ;-)  I added the ampersand.
While doing that, I fixed a couple bugs, so that was good. 

I suggest we merge this, and continue to iterate on it.  Actually seeing the live page will make it easier to understand what needs to change.

superkhau: TY. Lol, sorry for causing you more pain but glad you found more bugs. ;)

I've been wanting to merge it forever, but @cgole @0candy @bosco-li are on point for this now as other parties need to be involved.

I personally agree with you though on just landing this and changing it up when we receive confirmation. @cgole Is this possible?

superkhau: cc @raymondfeng ^

bajtos (approved):
LGTM.

As far as I am concerned, this can and should be landed ASAP. Any further changes/refinements can be made in follow-up pull requests. [Perfect is the enemy of good](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good).

superkhau (approved):

crandmck:
I'll ping  @cgole @0candy @bosco-li  about further iteration.

bosco-li:
@pthieu can you help take a look? https://github.com/strongloop/loopback.io/pull/216#issuecomment-276488448

PR #218, opened 2016-12-29
Fix link with example modules

fabiopagoti:
Two links inside the documentation website were broken.

From my understading there was repo called [strongloop/loopback-example](https://github.com/strongloop/loopback-example) and for some reason it does not exist anymore.

To replace the broken link I have added a new link with all loopback example apps available in strongloop profile.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -14,8 +14,8 @@ item: examples
           <div class="c-container -padded">
             <div class="c-grid">
               <div class="_col -col-2-3">
-                <h2 class="c-heading -h3"><a href="https://github.com/strongloop/loopback-example">LoopBack example module</a></h2>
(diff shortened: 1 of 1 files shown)


slnode:

crandmck:
Hi @fabiopagoti thanks for the PR.  Yes, @superkhau removed that repo (which was just a README with links to the other examples) since https://loopback.io/doc/en/lb3/Tutorials-and-examples.html now serves that purpose.

Your suggestion is fine, except that we've also recently moved some examples into the `strongloop-community` org, so the query won't find those.   So it would be better to link to https://loopback.io/doc/en/lb3/Tutorials-and-examples.html since we'll be maintaining that page with links to all examples moving forward.

Could you please make that change in your PR?

crandmck:
I went ahead and made this change.  Thanks anyway for the suggestion and the PR!

PR #226, opened 2017-01-03
Fix after remote error hook example

johnrc:
Since this is an `afterRemoteError` hook, you need to add additional properties to `ctx.error` not `ctx.result`.


slnode:

crandmck:
Thanks!

Issue #59, opened 2016-09-26
Review old doc site for any remaining misc articles to move over


crandmck:
We've migrated all the LoopBack content from docs.strongloop.com, but there are some pages that didn't get moved that are not part of LoopBack _per se_, but are related.  

Should we move any or all of these?
- [Building a real-time app using socket.io and AngularJS](https://docs.strongloop.com/display/MSG/Building+a+real-time+app+using+socket.io+and+AngularJS)
- [Pub-sub](https://docs.strongloop.com/display/MSG/Pub-sub) - (NOTE loopback-example-pubsub is alerady in `_includes/readmes`)
- [Supporting module reference](https://docs.strongloop.com/display/NODE/Supporting+module+reference)

PR #227, opened 2017-01-04
Rework current-context docs

bajtos:
Rework the documentation page "Using current context" - drop mentions of the old loopback-context module, describe the new official solution implemented in strongloop/loopback#1495

Connect to strongloop/loopback#1495

@crandmck @superkhau please review

### DIFF
--- a/pages/en/lb3/Using-current-context.md
+++ b/pages/en/lb3/Using-current-context.md
@@ -10,16 +10,6 @@ permalink: /doc/en/lb3/Using-current-context.html
 summary:
 ---
 
-{% include warning.html content="Using the current context feature is not recommended!
-
-The current implementation of loopback-context is based on the module [continuation-local-storage](https://www.npmjs.com/package/continuation-local-storage) 
-which is known to have many problems, (for example, see [issue #59](https://github.com/othiym23/node-continuation-local-storage/issues/59)).
-As a result, loopback-context does not work in many situations, as can be seen from issues reported in LoopBack's
-[issue tracker](https://github.com/strongloop/loopback/issues?utf8=%E2%9C%93&q=is%3Aissue%20getCurrentContext).
-
-See [loopback issue #1495](https://github.com/strongloop/loopback/issues/1495) updates and an alternative solution.
-" %}
-
 LoopBack applications sometimes need to access context information to implement the business logic, for example to:
 
 * Access the currently logged-in user.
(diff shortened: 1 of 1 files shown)


bajtos:
I need to land  https://github.com/strongloop/loopback/issues/3048 before I can finish this patch, as the 2.x-specific configuration may change along the way.

superkhau (approved):

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
+      if (!ctx.args.options.accessToken) return next();
+      User.findById(ctx.args.options.accessToken.userId, function(err, user) {
+        if (err) return next(err);
+        ctx.args.options.currentUser = user;
+        next();
+      });
+    });
 };
 ```
+
```
superkhau: remove empty line

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
-
 LoopBack applications sometimes need to access context information to implement the business logic, for example to:
 
 * Access the currently logged-in user.
 * Access the HTTP request (such as URL and headers).
 
 A typical request to invoke a LoopBack model method travels through multiple layers with chains of asynchronous callbacks. It's not always possible to pass all the information through method parameters. 
 
-**See also**: [Example in LoopBack repository](https://github.com/strongloop/loopback/blob/master/example/context/app.js).
+In LoopBack 2.x, we have introduced current-context APIs which used the module
```
crandmck: LoopBack 2.x introduced current-context APIs using the [continuation-local-storage](https://www.npmjs.com/package/continuation-local-storage) module ...

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
+[continuation-local-storage](https://www.npmjs.com/package/continuation-local-storage)
+to provide a context object preserved across asynchronous operations.
+Unfortunately, as we learned later, this module is not reliable and have many
+problems (for example, see [issue #59](https://github.com/othiym23/node-continuation-local-storage/issues/59)).
+As a result, our current-context feature does not work in many situations,
+as can be seen from issues reported in LoopBack's
+[issue tracker](https://github.com/strongloop/loopback/issues?utf8=%E2%9C%93&q=is%3Aissue%20getCurrentContext).
```
crandmck: Unfortunately, this module is not reliable and has many known problems (for example, see [issue #59](https://github.com/othiym23/node-continuation-local-storage/issues/59)).  As a result, the current-context feature does not work in many situations, (see [related issues](https://github.com/strongloop/loopback/issues?utf8=%E2%9C%93&q=is%3Aissue%20getCurrentContext)).

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
 
-## Error messages
+To address this problem, we have moved all current-context-related code to a
+standalone module [loopback-context](https://github.com/strongloop/loopback-context)
+and removed all current-context APIs in the version 3.0 (see
+[Release Notes](3.0-Release-Notes.html#current-context-api-and-middleware-removed) for
+more details).
```
crandmck: To address this problem, LoopBack 3.0 moves all current-context-related code to the [loopback-context](https://github.com/strongloop/loopback-context) module and removed all current-context APIs (see
 [Release Notes](3.0-Release-Notes.html#current-context-api-and-middleware-removed)).

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
 
-LoopBack 3.0 [removed current-context APIs](3.0-Release-Notes.html#current-context-api-and-middleware-removed).
-An application that uses current-context will print the following error message
-when it receives its first HTTP request:
+Having done that, we do recognize the need of accessing information like the
+currently logged-in user deep inside application logic, for example in
+[Operation Hooks](Operation-hooks.html). Until there is a reliable
+implementation of continuation-local-storage available for Node.js, we are
+recommending to explicitly pass any additional context via `options` parameter
+of (remote) methods.
```
crandmck, line 39: However, applications clearly need to acces information like the currently logged-in user in application logic, for example in [operation hooks](Operation-hooks.html). Until there is a reliable implementation of continuation-local-storage available for Node.js, explicitly pass any additional context in the `options` parameter of (remote) methods.

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
-```
-Unhandled error for request GET /api/Users:
-Error: remoting.context option was removed in version 3.0.
-For more information, see https://loopback.io/doc/en/lb3/Using-current-context.html
-for more details.
-    at restApiHandler (.../node_modules/loopback/server/middleware/rest.js:44:15)
-    at Layer.handle [as handle_request] (.../node_modules/express/lib/router/layer.js:95:5)
-    ...
-```
+All built-in methods like
```
crandmck: In LoopBack 3.0, all built-in methods ...

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
+[PersistedModel.find](http://apidocs.strongloop.com/loopback/#persistedmodel-find)
+or
+[PersistedModel.create](http://apidocs.strongloop.com/loopback/#persistedmodel-create)
+were already modified to accept an optional `options` argument.
 
-To remove this warning, disable the context middleware added by the built-in REST handler. Set the `remoting.context` property in `server/config.json` to **false**; for example:
+[Operation Hooks](Operation-hooks.html) expose the `options` argument
```
crandmck: [Operation hooks](Operation-hooks.html)

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
-  "remoting": {
-    "context": false,
-    ...
-  },
-  ...
-}
-```
+The missing piece is how to initialize the `options` parameter when a method is
+invoked via REST API, and do it in a safe manner, so that clients cannot
+override sensitive information like the currently logged-in user.
```
crandmck: You must safely initialize the `options` parameter when a method is invoked via REST API, ensuring that clients cannot override sensitive information like the currently logged-in user.  Doing so requires two steps:
- Annotate "options" parameter in remoting metadata
- Customize the value provided to "options"

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
 
-If your application relies on `loopback.getCurrentContext`, rework your code to use `loopback-context` directly, per the following instructions.
+The solution we have adopted has two parts.
 
-## Install loopback-context
+## Annotate "options" parameter in remoting metadata
 
-Add `loopback-context` to your project dependencies
+Methods accepting an `options` argument should declare this argument in their
```
crandmck: Methods accepting an `options` argument must declare ...

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
+remoting metadata and use a special value `optionsFromRequest` as the `http`
+mapping.
```
crandmck: _I think you mean_:
... set the `http` property to the special string value `"optionsFromRequest"`.

_Unless "the `http` mapping" means something else_?

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
+{
+  "arg": "options",
+  "type": "object",
+  "http": "optionsFromRequest"
+}
 ```
 
-## Configure context propagation
+Under the hood, this "magic string" value is converted by `Model.remoteMethod`
+to a function that will be called by strong-remoting for each incoming request
```
crandmck, line 71: Under the hood, `Model.remoteMethod` converts this "magic string" value ...

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
+in order to build the value for this parameter.
```
crandmck: ...to build the value for this parameter.
(_"in order" is superfluous_)

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
 
-To setup your LoopBack application to create a new context for each incoming HTTP request, configure `per-context` middleware in your `server/middleware.json` as follows:
+{% include tip.html content='
+Computed "accepts" parameters have been around for a while and they are well supported by LoopBack tooling. For example, the Swagger generator excludes computed properties from the API endpoint description. As a result, the "options" parameter will not be described in the Swagger documentation.
```
crandmck: ... For example, the [Swagger generator](Swagger-generator.html) excludes ...

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
-```javascript
-{
-  "initial": {
-    "loopback-context#per-request": {},
-  }
-  ...
-}
-```
+All built-in method have been already modified to include this new "options"
+parameter.
```
crandmck, line 79: Methods of built-in model accept this new "options" parameter.
_You already said this, but I suppose it bears repeating_.

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
 
-{% include important.html content="By default, the HTTP req/res objects are not set onto the current context. You need to set `enableHttpContext` to true to enable automatic population of req/res objects.
-" %}
+{% include note.html content='
+In LoopBack 2.x, this feature is disabled by default for compatibility reasons.  You can enable it by adding `"injectOptionsFromRemoteContext": true` to your model JSON file.
+' %}
```
crandmck, line 83: To enable, add `"injectOptionsFromRemoteContext": true` to your model JSON file.

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
 
-## Use the current context
+## Customize the value provided to "options"
 
-Once you've enabled context propagation, you can access the current context object using `LoopBackContext.getCurrentContext()`.
-The context will be available in middleware (if it is loaded after the context middleware), remoting hooks, model hooks, and custom methods.
+When strong-remoting is resolving the "options" argument, it will call model's
+method `createOptionsFromRemotingContext`. The default implementation of this
```
crandmck: When `strong-remoting` resolves the "options" argument, it calls model's `createOptionsFromRemotingContext` method.

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
+method returns an object with a single property `accessToken` containing
+the `AccessToken` instance used to authenticate the request.
 
-```javascript
-var LoopBackContext = require('loopback-context');
+There are several ways how to customize this value.
```
crandmck: There are several ways to customize this value:
- Override `createOptionsFromRemotingContext` in your model.
- Use a "beforeRemote" hook.
- Use a custom strong-remoting phase.

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
-      loopbackContext.set('currentUser', user);
-    }
+A better approach is to write a mixin that overrides this method and that can
+be shared between multiple models.
+
+### Use a "beforeRemote" hook
+
+Because the "options" parameter is a regular method parameter, it can be
+accessed from remote hooks via `ctx.args.options`.
+
```
crandmck, line 115: Because the "options" parameter is a regular method parameter, you can access it rrom remote hooks via `ctx.args.options`.

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
-    var ctx = LoopBackContext.getCurrentContext();
-    var currentUser = ctx && ctx.get('currentUser');
-    console.log('currentUser.username: ', currentUser.username); // voila!
-    ...
-    cb(null);
-  };
-  ...
+Again, a hook like this can be reused by placing the code in a mixin.
+
+Note that remote hooks are executed in order controller by the framework, which
```
crandmck: _This is confusing--what "framework"?  LoopBack?  Is the order indeterminate?_

Note that remote hooks are executed in order controlled by the framework, ...
_Note typo: controlle**d** is what you meant, not "controller", right?_

pages/en/lb3/Using-current-context.md
```diff
@@ -10,156 +10,147 @@ permalink: /doc/en/lb3/Using-current-context.html
...
+may be different from the order in which you need to modify the options
+parameter and then read the modified values. Please use the next suggestion if
+this order matters in your application.
+
+### Use a custom strong-remoting phase
+
+Internally, strong-remoting uses phases similar to [Middleware
+Phases](https://loopback.io/doc/en/lb3/Defining-middleware.html). The framework
```
crandmck: [middleware phases](https://loopback.io/doc/en/lb3/Defining-middleware.html)

crandmck:
I have a quite a few minor wording suggestions.  In general, some of the text reads more like a blog than documentation (in docs, we usually avoid using first person).  In general, my comments are the suggested revision to the text, except for italicized text which is my meta-comments/questions :-)

@bajtos If you don't have time to go through and make all these changes, I'd be glad to make the ones of which you approve (directly in your branch before merging the PR).  In some cases, I have questions though which need to be addressed.

pages/en/lb3/Using-current-context.md
bajtos: Built-in method were modified in both 3.x and 2.x. Should I explicitly mention this fact?

(IIRC, "options" parameter was added before we started this current-context rework.)

pages/en/lb3/Using-current-context.md
bajtos: > ... set the `http` property to the special string value `"optionsFromRequest"`.

yes, you are right 👍

pages/en/lb3/Using-current-context.md
bajtos: I'll improve this sentence a bit, but let's discuss more in the "main" thread.

bajtos:
@crandmck thank you for the review. I have fixed (almost) all your comments in e28628f, but there are few of them that deserve further discussion, see below.

https://github.com/strongloop/loopback.io/pull/227#discussion_r94883041
> In LoopBack 3.0, all built-in methods

Built-in method were modified in both 3.x and 2.x. Should I explicitly mention this fact?

(IIRC, "options" parameter was added before we started this current-context rework.)

https://github.com/strongloop/loopback.io/pull/227#discussion_r94884722
> _This is confusing--what "framework"?  LoopBack?  Is the order indeterminate?_
>
> Note that remote hooks are executed in order controlled by the framework, ...
_Note typo: controlle**d** is what you meant, not "controller", right?_

The order is determinate AFAIK, but it may be difficult to setup the application is such way that one gets exactly the order they want.

I don't know all the details, but it looks like the hooks are invoked from most-specific (e.g. `MyModel.myMethod`) to least specific (e.g. `MyModel.*` and then `**`), at least according to https://gist.github.com/ebarault/fdc39c5d07cc40f08664256f9e00905a#file-context-js-L242-L248:

> here, we use Model.beforeRemote hook instead of originally used RemoteObject.before hook 
> since we want remoteContext injection to perform before any other Model.beforeRemote hook
> (Model.beforeRemote hooks are called before RemoteObject.before hooks)
> NOTE: the remotingContext will be available in the Model.beforeRemote('**', function(ctx, instance, next) {...}) hook
> (/!\ only the ** one),  provided that you are able to register it after the one trigger in the iteration below
> you can still discriminate the remote method name in that hook using the ctx.mehod.name property
> (see: https://github.com/strongloop/strong-remoting/issues/340)

I think (but didn't verify), that if there are multiple hooks registered at the same level of specificness, then the order will be controlled by the order in which these hooks were registered. This order is difficult to control, because it's basically controlled by loopback-boot and how it invokes different kinds of scripts (model `.js` file, mixins, components, boot scripts).

So for example, if write a boot script that registers a `beforeRemote` hook adding a new `options` property, and you also register a `beforeRemote` hook in your model .js file (e.g. `common/models/my-model.js`) where you would like to read this added property, then you are out of luck (at least I believe so).

As far as this document goes, I don't want to dig into specific details, just advise the user that the order in which hooks are invoked may be out of their control and if the order is important for them, then they should use a different solution.

How can I express this information better than I do it now?

pages/en/lb3/Using-current-context.md
crandmck: If it applies to both v2&3, then perhaps word as follows:
> Built-in methods such as PersistedModel.find or PersistedModel.create accept an options argument.

crandmck:
> Built-in method were modified in both 3.x and 2.x. Should I explicitly mention this fact?

If it applies to both vers, then I suggest this:

Built-in methods such as `PersistedModel.find` or `PersistedModel.create` accept an options argument.

> As far as this document goes, I don't want to dig into specific details, just advise the user that the order in which hooks are invoked may be out of their control and if the order is important for them, then they should use a different solution.

How about this:

It may not always be possible to control the order in which remote hooks are executed.  If you need to control the order, then use use a custom strong-remoting phase as described in the following section.

bajtos:
@crandmck thank you, I applied your suggestions in 3994cae.

I am going to rebase & squash the commits and land the pull request now.

PR #230, opened 2017-01-05
Revised docs covering error handling

doublemarked:
Hey @crandmck, following @bajtos's suggestion in #134 I did a pass on error handling docs. I should have said something there, because I see now that you just today touched up some of the same places.

Can you give this a review and let me know if it's the direction you'd it to go or not? If it seems ok I'll resolve the conflicts introduced by your recent commits.

### DIFF
--- a/pages/en/lb3/Defining-middleware.md
+++ b/pages/en/lb3/Defining-middleware.md
@@ -600,6 +600,17 @@ In this case, since the `GET /` middleware ends the response chain, the "catch
 However, when you make a POST request to /, the "catch-all" route is triggered because it is declared **before** the post route.
 Doing a POST will show the console message from both the "catch-all" route and the `POST / `route.
 
+## Default middleware
+
+LoopBack (via strong-remoting) registers two error-handling middleware by default:
+
+  * `urlNotFound`: A catch-all handler converting all requests that reach the middleware into an Error object with status 404, so that 404 error responses are consistent with "usual" error responses.
+  * `errorHandler`: An instance of the [`strong-error-handler`](https://github.com/strongloop/strong-error-handler).
+
+The presence of these middleware is controlled by configuration in [`config.json`](config.json.html).
+
+Note: Scaffolded apps have strong-remoting's `errorHandler` disabled in favor of middleware.json config for `strong-error-handler`.
+
 ## Examples
 
 ### Static middleware
@@ -690,17 +701,10 @@ Instead, it will be ignored because the route was already matched.
 
 ### Error-handling middleware
 
-Use error-handling middleware to deal with request errors.
+Use error-handling middleware to deal with request errors. The  [`strong-error-handler`](https://github.com/strongloop/strong-error-handler/) middleware is scaffolded by default and sufficient for many projects, but additional custom error-handling middleware may be registered as necessary.
 While you are free to register any number of error-handling middleware, be sure to register them in the "final" phase.
-LoopBack registers two error-handling middleware by default:
 
-* `urlNotFound` middleware converts all requests that reach the middleware into an Error object with status 404,
-  so that 404 error responses are consistent with "usual" error responses.
-* `errorhandler` middleware is from the [errorhandler](https://github.com/expressjs/errorhandler) module,
-  previously available in Express v.3 as `express.errorHandler`. For information on customizing this error handler.
-  See [Customizing REST error handling](Environment-specific-configuration.html#customizing-rest-error-handling).
-
-Example of a custom error processing middleware:
+Example of a custom error-handling middleware:
 
 ```js
 module.exports = function() { 
@@ -725,3 +729,5 @@ To register this middleware:
 
 3.  Start the application.
 4.  Load [http://localhost:3000/url-does-not-exist](http://localhost:3000/url-does-not-exist) in your browser.
+
+Note: This middleware prints errors much like the `strong-error-handler` middleware. The `strong-error-handler` can be configured to not print errors by setting its [`log`](https://github.com/strongloop/strong-error-handler#log) parameter to `false`.
--- a/pages/en/lb3/Migrating-to-3.0.md
+++ b/pages/en/lb3/Migrating-to-3.0.md
@@ -68,42 +68,32 @@ You may need to make the following changes:
 - The new error handler's HTTP responses have a `error.statusCode` property
 instead of `error.status`.  Update any clients that read the HTTP status code
 from the response body.
-- Change any uses of the `loopback#errorhandler` short-cut to  `require('strong-error-handler')`.
-- Add the following to `config.json`:
+- Change any use of the `loopback#errorhandler` short-cut to `require('strong-error-handler')`.
+- Enable `strong-error-handler` by adding it to the `final:after` phase in your middleware config file (`server/middleware.json`):
 
 ```
   ...
-  "remoting": {
-    ...,
-    "handleErrors": false
-  }
-```
-
-This setting will not return stack traces in HTTP responses nor print errors to the console.
-
-For development, to return stack traces in HTTP responses, instead set `errorHandler.debug` to `true`;
-by default `errorHandler.log` is `true`, so the app will display errors to the console:
-
-```
-  ...
-  "remoting": {
-    ...,
-    "errorHandler": {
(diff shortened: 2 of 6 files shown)


slnode:

crandmck:
Your changes LGTM.  Sorry about the conflicts, I've been working on updating the 3.0 docs based on the changes listed in the release notes and migration guide.

Also, JFYI, we're in the middle of revising the strong-error-handler README (cf. https://github.com/strongloop/strong-error-handler/pull/36) which will be used for https://loopback.io/doc/en/lb3/Using-strong-error-handler.html

Thanks!

doublemarked:
@crandmck Ok, all sync'd up. Let me know if you'd like any changes!

And thanks for the info regarding the strong-error-handler docs. Two comments,

1. The strong-error-handler README seems like the best source of authoritative information on the module. I wonder sometimes whether the documentation contains an excess of similar but not identical information scattered throughout. The error handler info is a good example, as even in my patch, as I've replaced older documentation, there are 3-4 places where I've written similar information phrased slightly differently. Is this a problem? 

2. I enjoyed your discussion in that PR regarding use of `Use` vs. `Usage`, particularly because in my original PR I had changed a case of `use` to `usage`. It never occurred to me that this was a grammatical quirk of engineers :) I've reverted the change.

bajtos:
@slnode ok to test

pages/en/lb3/Defining-middleware.md
```diff
@@ -690,17 +690,10 @@ Instead, it will be ignored because the route was already matched.
 
 ### Error-handling middleware
 
-Use error-handling middleware to deal with request errors.
+Use error-handling middleware to deal with request errors. The  [`strong-error-handler`](https://github.com/strongloop/strong-error-handler/) middleware is scaffolded by default and sufficient for many projects, but additional custom error-handling middleware may be registered as necessary.
 While you are free to register any number of error-handling middleware, be sure to register them in the "final" phase.
-LoopBack registers two error-handling middleware by default:
 
-* `urlNotFound` middleware converts all requests that reach the middleware into an Error object with status 404,
-  so that 404 error responses are consistent with "usual" error responses.
```
bajtos, line 698: I think the information about `urlNotFound` middleware may be useful to readers and should be preserved. `urlNotFound` is a LoopBack-specific concept that's not generally known by Express developers (at least I think so).

Thoughts?

pages/en/lb3/Migrating-to-3.0.md
```diff
@@ -68,42 +68,32 @@ You may need to make the following changes:
...
+         "debug": false,
+         "log": true
+       }
     }
   }
 ```
 
-- Add the error handler as middleware in `server/middleware.json` file:
+- In your `config.json` file, disable `remoting.rest.handleErrors`. This will
+  instruct Loopback to respect your middleware-based configuration of the error handler,
```
bajtos:
```diff
- ... handler,
+ ... handler:
```

bajtos (approved):
Thank you @doublemarked for reviewing the error-handling documentation and making the improvements in all these different places!

The content changes LGTM. I'll leave review of formatting and other formal matters to @crandmck.

pages/en/lb3/Defining-middleware.md
crandmck: I agree it's worth keeping.

crandmck:
> The strong-error-handler README seems like the best source of authoritative information on the module. I wonder sometimes whether the documentation contains an excess of similar but not identical information scattered throughout. The error handler info is a good example, as even in my patch, as I've replaced older documentation, there are 3-4 places where I've written similar information phrased slightly differently. Is this a problem?

I agree this is a concern.  We need to maintain a balance having the same information duplicated in too many places and not having info "discoverable" by someone just trying to get stuff done.  Often,  upon an big release (e.g. 3.0), there will be lots of duplication, especially with release notes, migration info, etc. because there are lots of ways people can come at the software: upgrading an existing app, writing a new app, just adopting certain new features, etc.  After some time, it becomes easier to eliminate some of the duplication.  So at this point, I wouldn't worry too much; we should definitely keep this in mind, though. 

> I enjoyed your discussion in that PR regarding use of Use vs. Usage, particularly because in my original PR I had changed a case of use to usage. It never occurred to me that this was a grammatical quirk of engineers :) I've reverted the change.

I'm glad you enjoyed it. :-)  Although it's certainly not a critical issue, it's one of my pet peeves, along with "on premise" (vs. the correct "on premises"), and other malapropisms.

pages/en/lb3/Defining-middleware.md
doublemarked: You're right that this belongs somewhere. I thought it was referenced in a better place, but now see it is not. I'll bring it back into the docs, but I have a couple questions.

On investigating this again I see that there are actually two `urlNotFound` middlewares, 

  - `loopback#urlNotFound`, located at `loopback/server/middleware/url-not-found.js`. This is configured via middleware.json and is scaffolded by default.
  - `RestAdapter.urlNotFoundHandler`, located inside `strong-remoting/lib/rest-adapter.js`. This is controlled by a config.json directive, and enabled by default.

Neither of them accept any options, but they are both loaded by default in a freshly scaffolded project. The `loopback#urlNotFound` handler does not do anything because the one from strong-remoting catches everything first. 

Two questions:

1. To remain consistent with the errorHandler behavior, should we update the scaffolding and advice in docs to disable strog-remoting's `urlNotFound`, making the middleware configured `loopback#urlNotFound` the active catchall? This would make it more straightforward for an implementer to create their own catchall, by just replacing `loopback#urlNotFound` in their middleware config with their own middleware.

2. Maybe it's a mistake for there to be two of these? Is the plan to extract `urlNotFound`, as was done with strong-error-handler? Should it perhaps even be part of strong-error-handler? Any plans here?

bajtos: I think the `urlNotFound` middleware is different from the error handler. The reason why the REST adapters adds its own urlNotFound middleware is to prevent the top-level loopback app to add routes into the route space managed by strong-remoting.

Consider the following example:

```js
var loopback = require('loopback');
var app = loopback();

app.dataSource('db', { connector: 'memory' });
app.model(loopback.User, { dataSource: 'db' });

app.use('/api', loopback.rest());
app.get('/api/Users/foobar', (req, res) => res.send({ hello: 'world' }));

app.listen(3000, () => console.log('http://localhost:3000/'));
```

Currently, `GET /api/Users/foobar` returns 404 error `Unknown "User" id "foobar".` With the built-in `urlNotFound` handler disabled, the custom route would be run instead.

Honestly, I cannot tell whether that's a good or a bad thing, considering that the following code would call the custom route too:

```js
app.get('/api/Users/foobar', (req, res) => res.send({ hello: 'world' }));
app.use('/api', loopback.rest());
```

I am proposing to move this discussion to a standalone issue, so that other LoopBack users can get involved. As for this pull request, I let's keep the mention of `urlNotFound` middleware as it is now.

Thoughts?

> Maybe it's a mistake for there to be two of these? Is the plan to extract urlNotFound, as was done with strong-error-handler? Should it perhaps even be part of strong-error-handler? Any plans here?

We don't have any plans for that, but now that you mentioned it, I think it may make sense to provide this useful middleware in a place where other users can reuse it and strong-error-handler looks like the perfect vehicle for that!

crandmck:
AFAICT this is ready to be merged... Are there any other changes to make at this point?  As always, we can continue to iterate and refine the docs afterward.

pages/en/lb3/Defining-middleware.md
doublemarked: TYVM @bajtos. I agree, sounds like we need to move this into a separate Issue/discussion. I'm not convinced that it's imperative for strong-remoting to defend its route space, but I understand your logic and see that this is getting out of scope for this PR.

> We don't have any plans for that, but now that you mentioned it, I think it may make sense to provide this useful middleware in a place where other users can reuse it and strong-error-handler looks like the perfect vehicle for that!

I am attracted to the consistency of this.

> As for this pull request, I let's keep the mention of urlNotFound middleware as it is now.

Agreed. @crandmck - I'll update this shortly to bring this back and then this should be ready to merge.

PR #233, opened 2017-01-10
Corrected overlapped text

harikrishnanp:
Some text in this page is overlapped and unreadable. Fixed it by adding line break.

Eg:

image


slnode:

crandmck:
Thanks!

PR #241, opened 2017-01-13
Add env-based datasource config section

kjdelisle:
connected to https://github.com/strongloop-internal/scrum-loopback/issues/1396

If there's a section for this configuration already, I had great difficulty finding it.

Just adding a small bit about environment-based configuration, based on a customer question.

### DIFF
--- a/pages/en/lb3/Environment-specific-configuration.md
+++ b/pages/en/lb3/Environment-specific-configuration.md
@@ -187,6 +187,24 @@ Example data sources:
 }
 ```
 
+You can also configure your datasource._\_env\__.js file to use environment variables:
+
+{% include code-caption.html content="datasources.production.json" %}
+```javascript
+module.exports = {
+  db: {
+    connector: 'mongodb',
+    hostname: process.env.DB_HOST,
+    port: process.env.DB_PORT || 27017,
+    user: process.env.DB_USER,
+    password: process.env.DB_PASSWORD,
+    database: 'myapp',
+  }
+```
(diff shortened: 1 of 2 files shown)


crandmck:
Thanks for the PR.... We do have https://loopback.io/doc/en/lb2/Environment-specific-configuration.html#data-source-configuration

I think your addition would be better placed there.  Could you please move it there?  You could leave a link in `Advanced-topics-data-sources.html` if you like.

Also, since you had a hard time finding the, I'd also suggest adding cross-references to it wherever it  makes sense.

crandmck (changes requested):
Please see my comments above.

pages/en/lb3/Environment-specific-configuration.md
```diff
@@ -187,6 +187,22 @@ Example data sources:
...
+    connector: 'mongodb',
+    hostname: process.env.DB_HOST,
+    port: process.env.DB_PORT || 27017,
+    user: process.env.DB_USER,
+    password: process.env.DB_PASSWORD,
+    database: 'myapp',
+  }
+```
+
+
```
kjdelisle, line 207: Should I add that blurb about setting `export PRODUCTION=true` or `set PRODUCTION=true` here to make it clearer?

crandmck: Yes, I think that makes sense.

crandmck:
Thanks!

PR #249, opened 2017-01-16
describe component-passport user overriding

davidcheung:
To resolve issue where uniqueID by provider is in email format,
in loopback we sufix `@loopback.provider.com` to all uniqueIDs
see strongloop/loopback-component-passport#158


crandmck:
LGTM

PR #262, opened 2017-01-30
Update Exposing-models-over-REST.md

masseelch:
I think this was just a typo!

### DIFF
--- a/pages/en/lb3/Exposing-models-over-REST.md
+++ b/pages/en/lb3/Exposing-models-over-REST.md
@@ -399,7 +399,7 @@ Product.disableMethodByName('createChangeStream');         // Removes (GET|POST)
 
 ### Hiding endpoints for related models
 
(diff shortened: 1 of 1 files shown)


slnode:

masseelch:
I just saw, that there are a lot more places in the text, where the `disableMethodByName()` is used instead of the correct `disableRemoteMethodByName.`

crandmck:
Thanks for the PR.  I'll land your change and then go ahead and correct the rest of the places this occurs.

I think the source of this mistake is that the method in `strong-remoting` is [disableMethodByName](http://apidocs.strongloop.com/strong-remoting/#sharedclass-prototype-disablemethodbyname), but here and in general, one uses the Model implementation which--as you correctly point out--is [disableRemoteMethodByName](http://apidocs.strongloop.com/loopback/#model-disableremotemethodbyname).

PR #263, opened 2017-01-31
Add FA owner redirects

superkhau:
Allows us to find the FA owners doc via `loopback.io/fa-owner[s]` or `loopback.io/functional-area-owner[s]`.

cc @strongloop/loopback-devs


crandmck:
LGTM

PR #267, opened 2017-02-03
started => booted [WAITING FOR npm PUBLISH OF loopback-boot]

AshKyd:
The changelog indicates the `booted` event was added in 2015-01-08, Version 2.6.0

For reference, the emit is called here https://github.com/strongloop/loopback-boot/blob/master/lib/plugins/application.js#L131

### DIFF
--- a/pages/en/lb3/Events.md
+++ b/pages/en/lb3/Events.md
@@ -18,7 +18,7 @@ In addition to the [standard Node events](http://nodejs.org/api/events.html), Lo
 
 ## Application events
 
-By default, an application created with the [application generator](Application-generator.html) emits a 'started' event when it starts up, after running [boot scripts](Defining-boot-scripts.html).
+By default, an application created with the [application generator](Application-generator.html) emits a 'booted' event when it starts up, after running [boot scripts](Defining-boot-scripts.html).
 
 ## Model events
 


slnode:

crandmck:
This change looks correct to me.   @superkhau can you pls confirm...?
Bummed that the docs have been wrong for so long!

superkhau (approved):

superkhau:
@crandmck I believe this is correct. @bajtos Can you confirm? I can't recall if it has both a `started` and a `booted` event -- simple grep of lb + lb-boot doesn't show `emit('started'`.

bajtos:
The source line points to code that's part of the upcoming 3.0 release of loopback-boot, which didn't happen yet (cc @raymondfeng @davidcheung). As far as I understand the new loopback-boot code, it's true the app will emit `booted` when loopback-boot finishes its work.

However, the `started` event will be emitted too, as this is part of loopback scaffolding - see [app.start()](https://github.com/strongloop/loopback-workspace/blob/4ffa3928e51535e8095ae94c61419b12bba8be83/templates/projects/empty-server/files/server/server.js#L11) in `server/server.js` template.

I am proposing to keep the current text without changes and add a new paragraph mentioning the new `booted` event produced by the upcoming loopback-boot version.

bajtos (changes requested):
☝️

crandmck (approved):

crandmck:
Oops, I accidentally approved this, but IIUC, according to @bajtos we should leave the existing text "as is" and then add the `booted` event when loopback-boot v3.0 is published.  

I don't think adding this information _before_ it's published to npm makes sense.  So if that's going to happen soon, we can leave this PR open (with the suggested changes).  If not, then we should close this and maybe just open a tracking issue to add that when it does occur.

crandmck:
@AshKyd Do you want to make the changes [suggested above](#issuecomment-277733820) ?
If not, I'll close this....

bajtos:
I am afraid `loopback-boot@3` was still not released yet - see https://github.com/strongloop/loopback-boot/issues/241

crandmck:
Yes, it still has label `waiting`. But I thought we could go ahead and do this...

> keep the current text without changes and add a new paragraph mentioning the new booted event produced by the upcoming loopback-boot version.

Unless I'm misunderstanding...?

AshKyd:
No probs, will make the change shortly.

AshKyd:
After trying to write it, I don't think I that suggestion really works. Would something like the following be better?

> By default, an application emits the 'booted' event when it starts up, after running boot scripts. In addition, an application created with the application generator emits a 'started' event.

I'm not 100% sure this is right, but the reason I originally opened this ticket was because my app wasn't emitting `started`. I think this version text it clearer where that event comes from. Can someone please double-check before I make the changes?

crandmck:
@bajtos It looks like `loopback-boot` was published to npm some time ago.  Should we land this now?

bajtos:
@crandmck 
> It looks like `loopback-boot` was published to npm some time ago.  Should we land this now?

In my understanding, we are waiting specifically for the version 3.0.0 to be published. AFAICT, that did not happen yet, see https://github.com/strongloop/loopback-boot/releases and https://www.npmjs.com/package/loopback-boot.

We have an issue to track the task of releasing loopback-boot@3.0.0, see https://github.com/strongloop/loopback-boot/issues/241

AshKyd:
Closing this because it's likely no longer relevant and I don't want it in my list.

PR #270, opened 2017-02-07
Tutorial code breaking

guptakvgaurav:
While following this documentation for learning purpose, i fall into the issue where `mongoDs.automigrate` was raising an exception (`automigrate is not a function of undefined`). Bit of investigation leads to following solution:

  var mongoDs = app.dataSources['mongo-connector'];

Here, `mongo-connector` is the name of datasource which i created while defining the datasource. It could be mapped to datasource.json as well. 
Note: Their should be a generic method which should return a connected datasource. This will add some level of independence.

### DIFF
--- a/pages/en/lb3/Create-new-models.md
+++ b/pages/en/lb3/Create-new-models.md
@@ -133,7 +133,7 @@ See [Creating a database schema from models](Creating-a-database-schema-from-mo
 var async = require('async');
 module.exports = function(app) {
   //data sources
-  var mongoDs = app.dataSources['mongo-connector']; // 'name' of your mongo connector, you can find it in datasource.json
+  var mongoDs = app.dataSources.mongoDs; // 'name' of your mongo connector, you can find it in datasource.json
   var mysqlDs = app.dataSources.mysqlDs;
   //create all models
   async.parallel({
--- a/pages/en/lb3/Create-new-models.md
+++ b/pages/en/lb3/Create-new-models.md
@@ -133,7 +133,7 @@ See [Creating a database schema from models](Creating-a-database-schema-from-mo
 var async = require('async');
 module.exports = function(app) {
   //data sources
-  var mongoDs = app.dataSources.mongoDs;
+  var mongoDs = app.dataSources['mongo-connector']; // 'name' of your mongo connector, you can find it in datasource.json
   var mysqlDs = app.dataSources.mysqlDs;
   //create all models
   async.parallel({


slnode:

crandmck:
Thanks for the suggestion.  

In general, the "Getting Started" tutorials assume that you're following along, and thus in the previous article [Create new data source](https://loopback.io/doc/en/lb3/Create-new-data-source.html), you would have a data source named `mongoDs`.  While your change would in fact work and enable people to name their data source something else, we also have a repo `loopback-getting-started-intermediate` that has a number of [tagged releases](https://github.com/strongloop/loopback-getting-started-intermediate/releases) (to enable people to pick up the tutorial at various points if they haven't followed from the start).  

So, we would also need to make the corresponding change in that repo, and ensure it gets into all those tagged releases.  If you're willing to do that work, we could consider landing this change.

guptakvgaurav:
@crandmck Yes you are right. I renamed the connector and then expected it to work. So that my bad.

However, I still feel below is much cleaner implementation

> app.dataSources.getConnectedDS("mongoDs"); // return supplied DS.


where name should be optional.

**If name is supplied**
 1. Should return supplied DS if exists/connected.

**If name is not supplied**

1.  If application has single DS configured, then return the configured DS. 
>    app.datasources.getConnectedDS(); // return configured DS

2.  If application has multiple DS configured, then return error.
>    app.datasources.getConnectedDS(); // return error


I can work on it if you believe it will be good to have feature.

crandmck:
That makes sense.  As I said, we would also need to make the corresponding change in that repo, and ensure it gets into all those tagged releases. If you're willing to do that work, we could consider landing this change.

@superkhau I don't have time to make the change(s) to loopback-getting-started-intermediate.  Do you think we should do this?  If so, can you find someone to change the example app?  Perhaps @siddhipai ?

superkhau:
@guptakvgaurav Your proposal sounds like sugar -- IMO, we should not bother for now. That said, we are working on loopback-next, and I will add your thoughts to the roadmap there intead. I think we need a simplier API such as `app.dataSources.getDefault()` and auto set the first configured datasource as default (or return the one set as default by user in the configs such as `datasources.json`).

tldr;
I wouldn't bother.

@bajtos Thoughts? ^

superkhau:
>@superkhau I don't have time to make the change(s) to loopback-getting-started-intermediate. Do you think we should do this? If so, can you find someone to change the example app? Perhaps @siddhipai ?

I think we should just tell user's reading the docs they need to complete the previous steps before moving on instead of changing it all up.

bajtos:
We used to have default datasources in LoopBack 1.x (look for `autoAttach`) and it wasn't working well, it was deprecated in 2.x and removed in 3.x. I don't think adding something similar back is a good idea.

> I think we should just tell user's reading the docs they need to complete the previous steps before moving on instead of changing it all up.

+1

crandmck:
> I think we should just tell user's reading the docs they need to complete the previous steps before moving on instead of changing it all up.

Every page in the Getting Started tutorials has that already, e.g. 
https://screencast.com/t/nRDcuJfK

pages/en/lb3/Create-new-models.md
```diff
@@ -133,7 +133,7 @@ See [Creating a database schema from models](Creating-a-database-schema-from-mo
 var async = require('async');
 module.exports = function(app) {
   //data sources
-  var mongoDs = app.dataSources.mongoDs;
+  var mongoDs = app.dataSources['mongo-connector']; // 'name' of your mongo connector, you can find it in datasource.json
```
superkhau: How about:

```
 var mongoDs = app.dataSources.mongoDs; // where mongoDs is the name of your datasource in datasources.json
```

superkhau (changes requested):

pages/en/lb3/Create-new-models.md
crandmck: LGTM

crandmck:
@guptakvgaurav Would you like to make this change in your PR?   Then we will land it.

crandmck:
Thanks!

PR #278, opened 2017-02-10
auto-update lb3 readmes as well as lb2

rmg:
As a side effect, since the lb2 docs weren't being updated the build was
failing.


crandmck:
Aha! Thx @rmg

PR #288, opened 2017-02-20
Allow logged in users to read coffeeshops

p31d3ng:
Users won't be allowed to view any coffeshops (and therefore the coffeshop list will be empty) if not explicitly allowing users to read coffeshops, because all endpoints are denied explicitly at the first hand.


slnode:

crandmck:
@superkhau PTAL.  This is for the getting-started-intermediate app.  Does this look right to you?

superkhau (approved):

superkhau:
Commit message is too long by 3 characters though.

PR #294, opened 2017-02-24
Fix broken example links

0candy:
Fix https://github.com/strongloop-internal/scrum-loopback/issues/1443


0candy:
@crandmck PTAL

crandmck:
Hi @0candy thanks for the PR, but these README files are copied from their home repos, as explained in http://loopback.io/doc/en/contrib/Including-READMEs.html.  

In general, we should never be editing files in the `readmes` directory, but rather edit them directly in the source repo and rely on the script to update them in the docs.  Sometimes, to ensure that it displays properly in Jekyll, I do edit it here, but then I copy it to its home repo and check it in there.

I hope you didn't spend a bunch of time on this! :-(

crandmck:
In theory, you could use some of this work by copying the markdown files into the corresponding README.md file and then opening a PR in that repo.

PR #298, opened 2017-03-01
Updated create-angularjs-client.md

brentpayton:
In my experience with just one installation, "lb-ng" will not be installed by following the tutorial as written.  I added instructions to install the "apiconnect" NPM module which provides "lb-ng".  I also found that the folder containing "lb-ng" was not added to the path so I added a step to include the necessary folder in the path environment variable.

### DIFF
--- a/pages/en/lb3/Create-AngularJS-client.md
+++ b/pages/en/lb3/Create-AngularJS-client.md
@@ -44,10 +44,8 @@ For more information, see [AngularJS JavaScript SDK](AngularJS-JavaScript-SDK).
 To generate the Angular services for a LoopBack application, use the AngularJS SDK `lb-ng` command-line tool.  You may need to install `lb-ng` with the following command.  
 
 ```
-$ npm install -g apiconnect
+$ npm install -g loopback-sdk-angular-cli
 ```
-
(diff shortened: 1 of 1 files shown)


slnode:

crandmck:
Thanks for the PR @brentpayton.

Rather than installing API Connect, the best way to install the Angular CLI is given in http://loopback.io/doc/en/lb3/AngularJS-JavaScript-SDK.html#installation: 
Just do `npm install -g loopback-sdk-angular-cli`.

We don't want to tell people to install API Connect here for several reasons:
-  API Connect does not yet support LoopBack v 3.  
- It installs the `apic` CLI, which shares some of the same generator commands as `lb`, but has some differences and a bunch of additional stuff specific to API Connect.  
- This site is focused on the open-source LoopBack vs. using it with a commercial product such as API Connect.

So, would you please change this PR to just say to do `npm install -g loopback-sdk-angular-cli`....?

brentpayton:
Thanks for getting back to me so soon, @crandmck!  

I made the change you suggested and removed content about updating the path environment variable.  Updating the path was necessary with the way this PR was previously worded.

Preferences differ from project to project but would you have preferred that I open a bug first or otherwise communicate before creating the PR?  In retrospect, it seems like I jumped to the Pull 
Request track quickly.  This is my first PR.

crandmck:
No worries! 
In general, opening an issue is preferred for substantive changes (not for trivial changes), but we also won't reject a PR just because there isn't a corresponding issue.  We appreciate the contribution!

PR #311, opened 2017-03-07
Add "npm start" as a shortcut for "bundle exec..."

bajtos:
I never remember what's the full bundle command to run. OTOH, "npm start" is a commonly used command in Node.js land.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Bundler will look in the Gemfile for which gems to install. The `github-pages` g
 Run Jekyll using the following command:
 
 ```
-$ bundle exec jekyll serve
+$ npm start
 ```
 
 Then, load [http://localhost:4001/](http://localhost:4001/) on your browser.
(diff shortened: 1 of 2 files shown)


crandmck (approved):
Makes sense.  I created an alias for `bundle exec jekyll serve` a long time ago for myself, but this is a fine alternative.

ebarault:
@crandmck : would have been nice to add the `--incremental` flag, don't you think?

crandmck:
Sure, that makes sense.  I usually use `-i` myself, but I already have cmd aliases set up, so I hadn't actually needed this.  

Maybe add a `start_i` script option (or whatever) so you can do either.  Sometimes you need to do a clean rebuild (e.g. if you modify the sidebar.yml or other stuff that incremental doesn't rebuild).  Feel free to open a PR if you'd like to add that.

PR #313, opened 2017-03-08
Please process my merged commit

breadfit:
I did the following

1. Created Installation.md and translate to Korean.
2. Modified index.md


slnode:

pages/ko/lb3/index.md
```diff
@@ -1,6 +1,6 @@
 ---
 lang: ko
-title: "LoopBack 3.0 - 한국어"
+title: "LoopBack 3.0"
```
crandmck: I think it's important to keep the name of the language in this title, since this is the home page for Korean LB docs. It will make it immediately clear what the page is if (for example) someone lands there from a search.
You'll notice this is the convention followed by all other languages, and I'd like to be consistent.

pages/ko/lb3/Installation.md
```diff
@@ -0,0 +1,42 @@
+---
+title: 설치하기
+lang: ko
+keywords: LoopBack
+tags: [installation]
+sidebar: lb3_sidebar
+permalink: /doc/ko/lb3/Installation.html
+summary: Install the LoopBack CLI tool to create and develop LoopBack 3.0 apps.
+---
```
crandmck: Please add `layout: translation` so the page uses the proper layout, as described in http://loopback.io/doc/en/contrib/Translating_articles.html#procedure.

Also, for a partially-translated page such as this please add `trans_complete: false` to the frontmatter, as explained in http://loopback.io/doc/en/contrib/Translating_articles.html#front-matter-example.

crandmck (changes requested):

crandmck:
@breadfit Would you please make the suggested changes so we can land this?

pages/ko/lb3/index.md
breadfit: Okay. I will leave that part.

pages/ko/lb3/Installation.md
breadfit: Sorry. I'll revise what you said and then pass it back.

crandmck:
Thanks!  Keep up the translation work!!

PR #317, opened 2017-03-09
Explaining defining datasources programmatically

timlind:
There are many different APIs and therefore different examples, which lead to errors when trying to associate with a model.

There needs to be a clear explanation here of the differences between the APIs, how to define datasources programmatically and still use the defined name in model-config.js.

It should be recommended not to store credentials within source code.

### DIFF
--- a/pages/en/lb3/Attaching-models-to-data-sources.md
+++ b/pages/en/lb3/Attaching-models-to-data-sources.md
@@ -56,7 +56,11 @@ The tool will then add an entry such as the following to `datasources.json`:
 
 This example creates a MySQL data source called "corp1". The identifier determines the name by which you refer to the data source and can be any string.
 
+If you wish to hide data source credentials outside the source code, and instead use environment variables, you can define a datasource programmatically. In order for a programmatically defined data source to be registered for use in attaching to a model ... TBC
+
 ## Add data source credentials
+
+For security reasons, it is better not to include credentials in the source code, and instead extract them from environment variables when defining the datasource.
 
 Edit `datasources.json` to add the necessary authentication credentials for the data source; typically hostname, username, password, and database name.
 
@@ -93,3 +97,9 @@ and set the dataSource property for the model. For example, to make myModel use 
 ```
 
 By default, the model generator creates models to use the `db` data source.
+
+In order for programmatically defined datasources to be associated with a model without encountering errors ... TBC
+
+The API documentation includes a Registry class which includes a function Register.createDataSource the purpose of which is .... TBC ...
+
+The difference between loopback.createDataSource vs app.dataSource vs new DataSource() is ... TBC


slnode:

crandmck:
@timlind Thanks for the PR.  Before landing this, I'd like to get all the `TBC` sections filled in.  Would you will be able to do that?

ebarault:
@crandmck : see [my comment here](https://github.com/strongloop/loopback/issues/3246#issuecomment-285611092)
[this](https://loopback.io/doc/en/lb3/Advanced-topics-data-sources.html) is already explaining how to set up datasources programmatically.
Now there's indeed a potential issue with the way credentials are meant to be stored in the datasources.json file ATM, but i don't think recommending as a best practice to define all the datasources by code is a good idea. 

We could add a warning in the doc while a proper solution is defined though

timlind:
@crandmck nope, I actually don't know the answer to that info, it's why I've done the PR. When I try add datasources programatically I get an error when trying to associate them via model-config. I'm hoping someone can give an example that works and explain the different options and design decisions here.

timlind:
@ebarault I agree, the issue here is that the example of defining datasources programatically doesn't have any explanation of how the usage of that scenario compares to defining in the json, i.e as soon as I define it programatically I can't associate models using model-config.js. I'm actually also struggling to get the Model.attachTo method working with a named datasource that is defined programatically. I'd like the docs to help me understand what the intended use and design decisions are here. There is also more than one way to do these things (i.e several similar APIs) and sometimes they don't all work, so some explanation of the timing of when these datasources are accessible would be useful.

ebarault:
ok, i get your concern now. Never tried to do this, but would be interested to see how it goes. The best way to progress is trying to set up a sandbox that attempts doing this. I'm rather busy right now, would you mind posting a [sandbox](https://github.com/strongloop/loopback-sandbox) using the memory connector that would show your issue, this might be quicker to go with.

timlind:
@ebarault ok done, this repo is breaking. I realised the boot scripts are run after models, so trying to attaching data sources in a model.js is only going to work with datasources.json, still not sure where the datasources should be defined / attached, and if the datasource host is defined in json and password in a bootscript, whether that would work.

https://github.com/timlind/loopback-sandbox-programmatic-datasource/commit/979b02eb989c9db0563859dd722ec2b62309fa7b

crandmck:
> I actually don't know the answer to that info, it's why I've done the PR. 

OK, I see.  We'll keep this PR open until we can sort it out and fill in those sections.  Thanks for getting the ball rolling @timlind.

Thanks @ebarault for your help.  If needed, we can ask for other help to resolve this.

ebarault:
a solution was provided by @superkhau [here](https://github.com/strongloop/loopback/issues/3246#issuecomment-285740867)

the doc could be improved though.
plus : the solution provided differs from [that one](https://loopback.io/doc/en/lb3/Advanced-topics-data-sources.html) and @timlind could not make something out of the latter

crandmck:
OK, I hadn't seen https://github.com/strongloop/loopback/issues/3246 until just now... I think we can make some simple clarifications and add cross-references as you say.  I'll try to work on it asap.

crandmck:
Please see #319

timlind:
@crandmck thanks for looking into this, those notes on the doc will be super useful for DB creds (which was my use case), I think you might want to update the suggestion to define using datasources.js instead of programmatically though, right?

If we focus this issue just on defining data sources programatically, I'm pointing out that getting everything hooked up and working is really tricky.

If you play around with my sandbox (https://github.com/timlind/loopback.io/commit/be726363b9c5aeee019f9d6b261ebe72a28be299), just trying to follow the docs that use `new DataSource()`, it's not clear where to define datasources (bootscript? how does it integrate with a custom migration.js bootscript?), where to associate models (why does model-config.js not work?), and what the interaction is if the same datasource is partially defined in datasources.json (can model-config.js be used then?), if attaching models programatically, this should be done somewhere other than model.js where bootscripts have not been run and datasources have not been defined.

crandmck:
Closing in favor of #319.

PR #319, opened 2017-03-11
Db creds security

crandmck:
@timlind @ebarault This supercedes #317 and should help to clarify matters.
Thank you for bringing it to my attention.

BTW, I made a few unrelated formatting improvements along the way :-)
You don't need to worry about those.

### DIFF
--- a/pages/en/lb3/Advanced-topics-data-sources.md
+++ b/pages/en/lb3/Advanced-topics-data-sources.md
@@ -17,12 +17,9 @@ The diagram illustrates the relationship between LoopBack `Model`, `DataSource
 
 1.  Define the model.
 
-2.  Create an instance of `ModelBuilder` or `DataSource`. `DataSource` extends from `ModelBuilder`.
-    ModelBuilder is responsible for compiling model definitions to JavaScript constructors representing model classes.
-    DataSource inherits that function from ModelBuilder.
+2.  Create an instance of `ModelBuilder` or `DataSource`. `DataSource` extends from `ModelBuilder`. `ModelBuilder` is responsible for compiling model definitions to JavaScript constructors representing model classes.  `DataSource` inherits that function from ModelBuilder.
 
-3.  Use `ModelBuilder` or `DataSource` to build a JavaScript constructor (i.e. the model class) from the model definition.
-    Model classes built from `ModelBuilder` can be later attached to a DataSource to receive the mixin of data access functions.
+3.  Use `ModelBuilder` or `DataSource` to build a JavaScript constructor (i.e. the model class) from the model definition.  Model classes built from `ModelBuilder` can be later attached to a DataSource to receive the mixin of data access functions.
 
 4.  As part of step 2, `DataSource` initializes the underlying `Connector` with a settings object which provides configurations to the connector instance. 
     `Connector` collaborates with `DataSource` to define the functions as `DataAccessObject` to be mixed into the model class.
@@ -34,6 +31,9 @@ It's a factory for data access logic around model classes. With the ability to p
 `DataSource` provides the necessary abstraction to interact with databases or services to decouple the business logic from plumbing technologies.
 
 ## Creating a DataSource programmatically
+
+{% include warning.html content="Don't put production database credentials explicitly in JSON or JavaScript files, where they could be a security vulnerability.  Instead, load the values from environment variables.  For more information, see [Specifying database credentials with environment variables](Attaching-models-to-data-sources.html#specifying-database-credentials-with-environment-variables)
+"%}
 
 The [`DataSource` constructor](http://apidocs.strongloop.com/loopback-datasource-juggler/#datasource) accepts the following arguments:
 
--- a/pages/en/lb3/Attaching-models-to-data-sources.md
+++ b/pages/en/lb3/Attaching-models-to-data-sources.md
@@ -56,7 +56,11 @@ The tool will then add an entry such as the following to `datasources.json`:
 
 This example creates a MySQL data source called "corp1". The identifier determines the name by which you refer to the data source and can be any string.
 
-## Add data source credentials
+## Add database credentials
+
+{% include tip.html content="Putting production database credentials in a JSON file is not recommended for security reasons.
+Instead, create the data source programmatically and load the credentials from environment variables, as explained [below](#specifying-database-credentials-with-environment-variables).
+"%}
 
 Edit `datasources.json` to add the necessary authentication credentials for the data source; typically hostname, username, password, and database name.
 
@@ -73,6 +77,31 @@ For example:
     "database": "your-db-name"
   }
 ```
+
+### Specifying database credentials with environment variables
+
+Best practice is not to put production database credentials explicitly in JSON or JavaScript files, where they could be  a security vulnerability.  Instead, define datasources in a JavaScript file then load the values from environment variables.
+
+For example, assuming you have set a valid MongoDB username and password in the
+environment variables MONGO_USER and MONGO_PASS, respectively:
+
+{% include code-caption.html content="datasources.production.js" %}
+```javascript
+var DataSource = require('loopback-datasource-juggler').DataSource;
+
+var dataSource = new DataSource({
+   connector: require('loopback-connector-mongodb'),
+   host: 'localhost',
+   port: 27017,
+   database: 'mydb',
+   username: process.env.MONGO_USER,
+   password: process.env.MONGO_PASS
+});
+ ```
+
+You can use different credentials for development and production by using the NODE_ENV 
+environment variable, as explained in [Environment-specific configuration](Environment-specific-configuration.html#data-source-configuration).  
+
 
 ## Make the model use the data source
 
--- a/pages/en/lb3/Attaching-models-to-data-sources.md
+++ b/pages/en/lb3/Attaching-models-to-data-sources.md
@@ -41,26 +41,22 @@ $ apic create --type datasource
 You can also add and modify data sources using the API Designer tool.
 </div>
 
-The tool will prompt you for the name of the new data source and the connector to use; for example, MySQL, Oracle, REST, and so on.
+The tool will prompt you for the name of the new data source and the connector to use; for example, MongoDB, MySQL, Oracle, REST, and so on.
 The tool will then add an entry such as the following to `datasources.json`:
 
 {% include code-caption.html content="/server/datasources.json" %}
 ```javascript
   ...
   "corp1": {
-    "name": "corp1",
-    "connector": "mysql"
+    "name": "mongo1",
+    "connector": "mongodb"
   }
   ...
 ```
 
-This example creates a MySQL data source called "corp1". The identifier determines the name by which you refer to the data source and can be any string.
+This example creates a MongoDB data source called "mongo1". The identifier determines the name by which you refer to the data source and can be any string.
 
 ## Add database credentials
-
-{% include tip.html content="Putting production database credentials in a JSON file is not recommended for security reasons.
-Instead, load the credentials from environment variables, as explained [below](#specifying-database-credentials-with-environment-variables).
-"%}
 
 Edit `datasources.json` to add the necessary authentication credentials for the data source; typically hostname, username, password, and database name.
 
@@ -70,13 +66,17 @@ For example:
 ```javascript
 "corp1": {
     "name": "corp1",
-    "connector": "mysql",
-    "host": "your-mysql-server.foo.com",
+    "connector": "mongodb",
+    "host": "your-mongodb-server.foo.com",
     "user": "db-username",
     "password": "db-password",
     "database": "your-db-name"
   }
 ```
+
+{% include warning.html content="Putting production database credentials in a JSON file is not recommended for security reasons.
+Instead, load the credentials from environment variables, as explained [below](#specifying-database-credentials-with-environment-variables).
+"%}
 
 ### Specifying database credentials with environment variables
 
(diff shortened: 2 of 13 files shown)


ebarault:
@crandmck : please see this comment https://github.com/strongloop/loopback/issues/3246#issuecomment-286071328

this information should complement [this section](https://loopback.io/doc/en/lb3/Advanced-topics-data-sources.html)

bajtos:
`datasources.*.js` files are expected to export only the configuration of datasources, not full datasource objects.

```js
module.exports = {
  db: {
    connector: 'mongodb',
    host: 'localhost',
    port: 27017,
    database: 'mydb',
    username: process.env.MONGO_USER,
    password: process.env.MONGO_PASS
  }
};
```

However, the example is still unnecessary complex and can be greatly simplified per https://github.com/strongloop/loopback/issues/3246#issuecomment-286076113:

```js
// server/datasources.json
{
  "db": {
    "connector": "mongodb",
    "host": "localhost",
    "port": 27017,
    "database": "mydb",
    "username": "${MONGO_USER}",
    "password": "${MONGO_PASS}"
  }
}
```

IMPORTANT NOTE:

The way how datasource configurations are processed by loopback-boot, environment-specific configuration is merged with the baseline config. As a result, it's difficult to clear extra config options in environment-specific files (see e.g. https://github.com/strongloop/loopback-boot/issues/93):

```js
// datasources.json
{
  "db": {
    "connector": "mongodb",
    "database": "mydb",
    // the line below was added later by a developer
    "loggerLevel": "debug"
  }
}

// datasources.production.json
{
  "db": {
    "username": "${MONGO_USER}",
    "password": "${MONGO_PASS}"
  }
}

// oops, now we are using loggerLevel = debug in production,
// because the developer forgot to update datasources.production.json
```

To avoid the situation described above, I would personally recommend to use a connection string in production mode (see https://github.com/strongloop/loopback/issues/2089):

```js
// server/datasources.json
{
  "db": {
    "connector": "mongodb",
    "url": "${MONGODB_URL}",
    // configuration below is for Development/Test environments only
    "host": "localhost",
    "database": "my-database-name",
    // etc.
  }
}
```

bajtos (changes requested):
☝️

ebarault:
>
```
    "username": "${MONGO_USER}",
    "password": "${MONGO_PASS}"
```

ahh, much better indeed... **thank you** for bringing this to our attention

> connection string

@bajtos in your example, why mixing both a connection string and host/database params? would the connection string if provided override the latter? (due to some implementation details)

> environment-specific configuration is merged with the baseline config

yes... this is troublesome and can lead people to unexpected issues. your IMPORTANT NOTE should be somewhere in the doc (where the whole file.myenv.json/js is documented, not specifically in the context of this PR)

bajtos:
> in your example, why mixing both a connection string and host/database params? would the connection string if provided override the latter? (due to some implementation details)

Yes, connection string is overriding all other parameters, at least in the connectors maintained by us. See https://github.com/strongloop/loopback/issues/2089#issuecomment-186189300

ebarault:
thanks for clarifying

then it sort of superseeds what i understood was the best practice so far: 
> defining as many `datasource.env.json/js` files as there are environments, including development

do you consider the doc should reflect this? (mention it as best practice and keep mentioning other alternatives)

crandmck:
I'll update this PR to reflect @bajtos input above (asap--been slammed with other stuff).  I wasn't aware of this capability, and it definitely needs to be documented.

crandmck:
OK, I updated the PR with the input from @bajtos.

My apologies, I accidentally committed `pages/en/lb2/Basic-model-object.md` to this `db-creds-security` branch.  Please ignore it--It is completely unrelated to this PR and just has some formatting improvements.  I could back out the changes and recommit elsewhere, but if you don't mind ignoring, it can piggyback on this PR.... :-)

crandmck:
@bajtos @ebarault Could you please take another look.

ebarault:
@crandmck : sure, it'll try tonight or tomorrow

pages/en/lb3/Attaching-models-to-data-sources.md
```diff
@@ -56,7 +56,11 @@ The tool will then add an entry such as the following to `datasources.json`:
 
 This example creates a MySQL data source called "corp1". The identifier determines the name by which you refer to the data source and can be any string.
 
-## Add data source credentials
+## Add database credentials
+
+{% include tip.html content="Putting production database credentials in a JSON file is not recommended for security reasons.
+Instead, load the credentials from environment variables, as explained [below](#specifying-database-credentials-with-environment-variables).
+"%}
```
ebarault: i'd put this tip block after the /server/datasources.json example (L80). This will serve as a junction with the following section "Specifying database credentials with environment variables"

Also the display is a bit inconsistent between Advanced-topics-data-sources.md where it's a warning, and here where it's a tip.

pages/en/lb3/Attaching-models-to-data-sources.md
```diff
@@ -74,6 +78,44 @@ For example:
...
+
+{% include code-caption.html content="datasources.json" %}
+```javascript
+{
+  "db": {
+    "username": "${MONGO_USER}",
+    "password": "${MONGO_PASS}"
+  }
+}
+ ```
```
ebarault: i'd suggest providing a valid datasource config here (including a connector, a host, etc.)
shall we be consistent through the different examples regarding the datasources definition? for now we go from mysql to mongodb

pages/en/lb3/Attaching-models-to-data-sources.md
```diff
@@ -74,6 +78,44 @@ For example:
...
+
+You can use different credentials for development and production by using the NODE_ENV
```
ebarault: _Rewording suggestion (**bold** differs)_
"you can use different **datasources definitions, including** credentials by using..."

pages/en/lb3/Attaching-models-to-data-sources.md
```diff
@@ -74,6 +78,44 @@ For example:
...
+environment variable, as explained in [Environment-specific configuration](Environment-specific-configuration.html#data-source-configuration),
+for example `datasources.production.json` for production environment when NODE_ENV is 'production'.
+
+{% include note.html content="LooopBack merges configurations environment-specific configuration (for example in `datasources.production.json`) with the baseline configuration in `datasources.json`. As a result, it's difficult to clear extra settings in environment-specific files, so best practice is to use the single `url` connection string setting (where available) that overrides the other settings.  See the example below. 
+" %}
```
ebarault: - one typo: Loo**o**pback

- Also, this statement on how the merge between top-level file and environment files is operating and the inherent limitation should be mentioned at the top of the  **Environment-specific-configuration.html** file, shared for all applicable files, and combined with the content of the important note : "The additional files can override the top-level keys with value-types (strings, numbers) only. Nested objects and arrays are not supported." (which BTW is duplicated in the file, with two different wordings)

- The db credentials warning is missing in Environment-specific-configuration.html.

- for the record: the last example in http://loopback.io/doc/en/lb3/Environment-specific-configuration.html#data-source-configuration states a **datasources.production.json** file, but it's a **.js**

- _Rewording suggestion (**bold** differs)_
"As a result, **it might be difficult in some setups** to clear extra **datasource** settings **inherited from the top-level datasources.json file** in environment-specific files. **To ease the juggling between different datasources when switching environments it is recommended when applicable to use a single datasources.json file mixing both a individual settings such as `host`, `port`, `database` for local environment and a `url` connection string setting that will override the formers for environment such as staging and production.** See the example below."

pages/en/lb3/datasources.json.md
```diff
@@ -168,12 +171,11 @@ The additional files can override the top-level data-source options with string
...
+
 {% include code-caption.html content="datasources.json" %}
 ```javascript
 {
-  // the key is the datasource name
-  // the value is the config object to pass to
-  // app.dataSource(name, config).
   db: {
     connector: 'memory'
   }
```
ebarault: the content of this file is greatly duplicated with content in Environment-specific-configuration.html#data-source-configuration, which will make the whole difficult to maintain. Hence, cross-linking would be preferable.

ebarault (commented):
In addition to my inlined comments, i'd add that the accessibility of this information could be improved.
As shown below there are two sections discussing the subject: "Attaching models to datasources" and "Understanding datasources", in addition "Understanding datasources" is nested in a section named "Connecting models to datasources", which is much too close to "Attaching models to datasources".

* Defining models
** Attaching models to datasources

* Connecting models to datasources
** Understanding datasources

image

Ideally datasources would require their own top section, and would fit just before Defining models.
The content of **Defining datasources** would be extracted from the two sub-sections Attaching models to datasources and Understanding datasources.
The **Connecting models to datasources** section's name should change. It is as of now a melting pot of advanced details and functionalities.
The applicable content would fit in a new top-level **Attached models to datasources**

So the logical sequence would look like

* Defining datasources
** Datasources: Advanced topics
* Defining models
* Attaching models to datasources

ebarault:
further comments:

@bajtos : does loopback-boot supports template literals in config.json?

(i could only find this in changes.md)
```
2015-08-28, Version 2.12.0
==========================
 * Resolve ${var} values in component-config.json (Hage Yaapa)
 * Resolve ${var} values in middleware.json (Hack Sparrow)
```

if so, the example in Environment-specific-configuration.html#application-wide-configuration page should be also reviewed to make use of the ${MY_ENV_VAR} ability.

As of now it's 
```
module.exports = {
  host: process.env.CUSTOM_HOST,
  port: process.env.CUSTOM_PORT
};
```
it could be a config.json file 
```
```

pages/en/lb3/Advanced-topics-data-sources.md
```diff
@@ -35,6 +32,9 @@ It's a factory for data access logic around model classes. With the ability to p
 
 ## Creating a DataSource programmatically
 
+{% include warning.html content="Don't put production database credentials explicitly in JSON or JavaScript files, where they could be a security vulnerability.  Instead, load the values from environment variables.  For more information, see [Specifying database credentials with environment variables](Attaching-models-to-data-sources.html#specifying-database-credentials-with-environment-variables)
+"%}
+
```
ebarault, line 37: ok for the warning but it stands that the problem expressed [in this comment](https://github.com/strongloop/loopback.io/pull/317#issuecomment-285854277) is still not covered by the doc.

The doc explains how to create a datasource programmatically but some info should be added on where to use this snippet, and how to connect models on these datasources (programmatically I suppose, but one could assume that they could reference datasources created programmatically in the model-config.json

crandmck:
Thanks for all the input.  This is very helpful.

Re your comment

> there are two sections discussing the subject: "Attaching models to datasources" and "Understanding datasources", in addition "Understanding datasources" is nested in a section named "Connecting models to datasources", which is much too close to "Attaching models to datasources".

I totally agree, and I've been aware of this for quite a while, but I never got around to addressing it.  Now you'll guilt me into it ;-)  

But seriously, I agree we should fix this.  Unfortunately, I'm taken up with work on other projects at the moment.  However, I'll make sure it's put at the top of the queue once the other work is done.

pages/en/lb3/Advanced-topics-data-sources.md
crandmck: I opened a new issue for this: #331

pages/en/lb3/Attaching-models-to-data-sources.md
crandmck: Done, and made it warning.

pages/en/lb3/Attaching-models-to-data-sources.md
crandmck: Done; changed all examples to MongoDB.

pages/en/lb3/Attaching-models-to-data-sources.md
crandmck: Changed wording (didn't use your specific suggestion, but clarified, I hope).

pages/en/lb3/Attaching-models-to-data-sources.md
crandmck: Incorporated all these suggestions.

pages/en/lb3/datasources.json.md
crandmck: I edited `datasources.json.md` page extensively to replace a lot of the content with cross-references.  The procedural info was really inappropriate for a reference article (which it is), and is provided elsewhere, so I removed it.

crandmck:
I'd like to keep this issue focused on the original task (specifying database credentials securely).  I opened #332 for the broader task of consolidating the information that's now in a couple different places.  I do want to address this, but don't have time at the moment to do it justice.

So @ebarault @bajtos if you would please review my latest changes and (hopefully) approve the PR.  I'd like to land it if possible.  We can carry on related work in #332 and #331.  Thanks again for your help.

pages/en/lb3/Attaching-models-to-data-sources.md
```diff
@@ -66,14 +66,66 @@ For example:
...
+ ```
+
+You can use different data source definitions (including database credentials) for development and production by using the NODE_ENV
+environment variable, as explained in [Environment-specific configuration](Environment-specific-configuration.html#data-source-configuration),
+for example `datasources.production.json` for production environment when NODE_ENV is 'production'.
+
+### Using multiple data source configurations
+
+LoopBack merges environment-specific configurations (for example in `datasources.production.json`) with the baseline configuration in `datasources.json`.
+Environment-specific configurations can override the top-level values with string and number values only. You cannot use objects or array values.
```
bajtos, line 110: This is not true - AFAIK, one can override object values with object values, array values with array values too.

bajtos (approved):
LGTM.

There is one claim that I think is not correct, but that can be fixed in a follow-up pull request.

pages/en/lb3/Attaching-models-to-data-sources.md
ebarault: well this is copy/paste from the current doc

crandmck:
> There is one claim that I think is not correct, but that can be fixed in a follow-up pull request.

OK, thanks.  I'm going to land this and open a separate issue to clear this up.

PR #328, opened 2017-03-19
Explicitly declare User variable

jesperbruunhansen:
Copy/pasting this snippet creates throws: `ReferenceError: User is not defined`.
Fix by explicitly define where to find variable.


slnode:

crandmck:
Thanks!

PR #330, opened 2017-03-21
Update Define-a-remote-hook.md

neeleshprat-camelot:
Typos


slnode:

crandmck:
@neeleshprat-camelot Thanks for catching this!

> a string that matches the remote method which you want to "hook" your function

Correct grammar would actually be:

"...a string that matches the remote method **to** which you want to "hook" your function..."

(which is presumably what I was trying to say in the first place, but I messed it up  :-(

Could you please make this change in your PR?

crandmck:
Thanks!

PR #338, opened 2017-03-29
Describe Authorization scopes

bajtos:
Documentation for the new feature to be added by https://github.com/strongloop/loopback/pull/3313.

Connect to strongloop/loopback#382

### DIFF
--- a/pages/en/lb3/Controlling-data-access.md
+++ b/pages/en/lb3/Controlling-data-access.md
@@ -204,10 +204,10 @@ It calculates the specifics by checking the access request against each ACL rule
 At each level, the matching yields three points:
 
 * 3: exact match
-* 2: wildcard match ('*')
+* 2: wildcard match (`'*'`)
 * -1: no match
 
-Higher-level matches take precedence over lower-level matches. For example, the exact match at model level will overweight the wildcard match.
+Higher-level matches take precedence over lower-level matches. For example, the exact match at model level will override the wildcard match.
 
 For example, consider the following access request:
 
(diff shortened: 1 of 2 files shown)


bajtos:
(This pull request must not be landed before the change in LoopBack is finished.)

superkhau (approved):

crandmck:
Actually, the doc PR should not be landed until the code is landed AND published to npm.

bajtos:
I have reworked the pull request to match what has been landed by strongloop/loopback#3313.

@ebarault @crandmck could you please review again?

crandmck (changes requested):
I went ahead and made wording and grammar edits directly in the branch, since that seemed most efficient.

IIUC, we should also update https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#methods since this feature will now be implemented.  Is that right?
If so, please add to this PR--or if you prefer I can take a stab at it next week.
Again, if I'm understanding, this reference info should largely duplicate what's in https://loopback.io/doc/en/lb3/Remote-methods.html#options.   If so, then instead of 
actually duplicating these tables, we can move them into the `_includes/content` and
single-source them so we only have to maintain them there.

pages/en/lb3/Controlling-data-access.md
```diff
@@ -253,45 +253,41 @@ Assuming the following ACL rules are defined:
...
-LoopBack provides a single built-in scope `DEFAULT` which is used whenever a method does not define any scopes or an access token does not contain any allowed scopes. This preserves backwards compatibility for existing applications.
+LoopBack provides the built-in scope `DEFAULT` that is used whenever a method does not define any scopes or an access token does not contain any allowed scopes. This preserves backwards compatibility for existing applications.
 " %}
 
 ### Define scopes required by remote methods
 
-The scopes are defined together with other remoting metadata via the new
-setting `accessScopes`. The user can invoke the remote method as long as their access token
-is granted at least one of the scopes listed in `accessScopes` array.
+The new model property `accessScopes` defines scopes as part of remoting metadata.
```
bajtos: @crandmck 
`accessScopes` is not a model property, it is a new property in remote method metadata. My proposal:

```diff
- The new model property `accessScopes` defines scopes as part of remoting metadata.
+ The new setting `accessScopes` defines scopes as part of remoting metadata.
```

bajtos:
@crandmck thank you for the edits.

> https://loopback.io/doc/en/lb3/Remote-methods.html#options

Good catch, `accessScopes` should be added to that table. I'll do that ASAP.

> IIUC, we should also update https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#methods since this feature will now be implemented. Is that right?

I am confused. The current doc pages says that remote methods cannot be configured via JSON files yet, but that hasn't been true for months. So yes, we should fix that section and I think that fix is independent of the changes I am making here. I am proposing to leave that out of scope of this pull request.

> Again, if I'm understanding, this reference info should largely duplicate what's in https://loopback.io/doc/en/lb3/Remote-methods.html#options. If so, then instead of
actually duplicating these tables, we can move them into the _includes/content and
single-source them so we only have to maintain them there.

Ditto, I am proposing to leave `_includes/content` out of scope of this patch.

bajtos:
Updated, PTAL.

pages/en/lb3/Controlling-data-access.md
crandmck: IIUC, it's a value set in the model JSON file, which is what I meant by a "model property".  Apparently that is confusing, but "setting" doesn't really convey any specific meaning.  

Perhaps it would be clearer to say 
> The `methods.accessScopes` property in the model JSON file defines...

crandmck:
> The current doc pages says that remote methods cannot be configured via JSON files yet, but that hasn't been true for months. So yes, we should fix that section and I think that fix is independent of the changes I am making here. I am proposing to leave that out of scope of this pull request.

No one told me, created an issue, or opened a doc PR, so apparently it was not documented (unless I'm missing something).  It's closely related to this PR, but we can do it separately if you prefer.
Added #352.

With [one small edit](https://github.com/strongloop/loopback.io/pull/338#discussion_r110962379) noted above I think this is ready to land, WHEN the change is published to npm.  We shouldn't land this until that time, though.

crandmck (approved):

crandmck:
Note: https://github.com/strongloop/loopback.io/commit/c3059946e7673214dc97f2487aa09ed83414c849 will require this to be rebased, because I added the doc for remote method config via JSON per #352.

pages/en/lb3/Controlling-data-access.md
bajtos: Done in ac92730

bajtos:
> Note: c305994 will require this to be rebased, because I added the doc for remote method config via JSON per #352.

Rebased, I moved the newly-added remote method option "accessScopes" to the new place you created.

>  I think this is ready to land, WHEN the change is published to npm. We shouldn't land this until that time, though.

Sounds good. I'd like to land https://github.com/strongloop/loopback/pull/3350 before making a new loopback 3.x release.

bajtos:
All required changes were landed and released in `loopback@3.7.0`. I'm going to rebase and land this patch.

PR #340, opened 2017-03-31
fixing hidden/protected examples

### LINKED ISSUE #339: modify examples for model hidden/protected properties
## URL of the page 

http://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#hidden-properties

## Nature of the issue 

examples show the hidden/protected properties nested inside the properties object, but according to the code datasource-juggler/model.js hidden/protected should be defined as top level properties

ebarault:
fixes #339 
see further inlined comments

### DIFF
--- a/pages/en/lb2/Model-definition-JSON-file.md
+++ b/pages/en/lb2/Model-definition-JSON-file.md
@@ -551,16 +551,19 @@ An example of a hidden property is User.password:
 
 {% include code-caption.html content="common/models/user.json" %}
 ```javascript
-...
+{
+  ...
   "properties": {
     ...
     "password": {
       "type": "string",
       "required": true
     },
-...
-   "hidden": ["password"],
-...
+    ...
+  },
+  "hidden": ["password", "verificationToken"],
+  ...
+}
 ```
 
 If you want to white-list the fields returned instead of black-listing them, consider:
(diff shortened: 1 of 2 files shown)


pages/en/lb2/Model-definition-JSON-file.md
```diff
@@ -580,20 +583,23 @@ Author will have personal information such as social security number etc, and th
...
     "email": {
       "type": "string",
       "required": true
     },
-...
-   "protected": ["email"],
-...
+    ...
+  },
+  "protected": ["email"],
```
ebarault, line 600: well actually user.json has not protected property : should we fix this in loopback or change the example? 
cc: @bajtos

bajtos: >  well actually user.json has not protected property : should we fix this in loopback or change the example?

I don't fully understand the purpose of `protected` properties. @raymondfeng what's your opinion?

If I had to decide, I'd rather fix the example. E.g.

```diff
- An example of a protected property is User.email:
+ An example configuring `email` as a protected property:
```

bajtos (approved):
Nice, thank you for the pull request @ebarault!

ebarault:
the difference between hidden and protected properties is quite subtle and confuses quite a lot of people i'm afraid
- **hidden**: whether the property will be returned when the query directly targets the model
- **protected** : whether the property will be returned when the model is targeted is an included relation

password could be protected for sure, but alone. I'll go for your proposal @bajtos in the meantime

crandmck:
I'll go ahead and merge this, but I think it's worth clarifying the diff. btw hidden & protected properties since it's a source of confusion.  I'll do that separately.

ebarault:
@crandmck : thanks
please make sure to modify this
```diff
- An example of a protected property is User.email:
+ An example configuring `email` as a protected property:
```

crandmck:
ah ok, thx for the heads up..  I'll do that

PR #346, opened 2017-04-05
Link to example of discovery function usage

kjdelisle:
The existing message doesn't make it clear that the first example is _not_ for dumping model definitions to text, but rather for leveraging discovered definitions at runtime.


crandmck (approved):

crandmck:
Thanks!

PR #348, opened 2017-04-06
Fix typo in function name

kjdelisle:
Just a small fix as part of our doc review for discovery and migration.


crandmck:
Thanks @kjdelisle 

I see you updated `readmes/loopback-connector-cloudant.md` to include recent changes to its README.  While this doesn't do any harm, it's typically unnecessary, because we have Jenkins pipeline set up that runs a script to pull down the various READMEs used in the doc site once they are published to npm.  Of course, adding it manually means we get it before the connector is published to npm (which is fine if it's just a doc improvement, but can be a problem for example if it's info on a new feature which wouldn't generally be available until the repo is republished to npm).

More details in http://loopback.io/doc/en/contrib/Including-READMEs.html.

PR #350, opened 2017-04-09
Documents primaryKey option in relations

garrows:
This was only documented in the `hasOne` relation documentation even though it's supported in `belongsTo` & `hasMany` as well.

Source code support for this can be found here 
https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/relation-definition.js#L1237
https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/relation-definition.js#L599


slnode:

crandmck:
Thanks!

PR #355, opened 2017-04-11
Documented 'filter' mode on model.strict

garrows:
Implemented here https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model.js#L254


slnode:

crandmck:
Thanks!

PR #358, opened 2017-04-14
update for soap ds.

crandmck:
Based on https://github.com/strongloop/loopback-cli/pull/29


rashmihunt (commented):
'how you can start with a SOAP WSDL document' should be changed to 'how you can start with a SOAP Web Service datasource'

rashmihunt (commented):
Remove the section:

Install strong-soap

Install the strong-soap Node module. This is a workaround until the ability to use the SOAP datasource instead of WSDL is supported.

npm install strong-soap --save

rashmihunt (changes requested):
@crandmck Added above review comments

PR #372, opened 2017-05-01
Update LoopBack-FAQ.md

wheel0ck:
How do you perform a password reset for a registered user?

When you start the project from scratch and follow the steps one by one, I think it is necessary to remember to configure the body parser and the loopback token otherwise we remain blocked for a while ...


slnode:

crandmck:
Thanks!

PR #375, opened 2017-05-02
Warn the reader that .json files should be kept

sanderboom:
The docpage is not very clear in that the `.js` files are only parsed when the corresponding `.json` file is available (eventhough it can be empty).

### DIFF
--- a/pages/en/lb3/Environment-specific-configuration.md
+++ b/pages/en/lb3/Environment-specific-configuration.md
@@ -21,6 +21,12 @@ LoopBack applications have the following types of configuration files:
 * **Middleware configuration files**, by default `server/middleware.json`.
 * **Configuration files for LoopBack components**, by default `server/component-config.json`.
 
+{% include note.html content="Using `*.js` files to configure a LoopBack application will only work
+when the original `.json` files are kept in place. Rather than replacing `.json` files with `.js`
+files, you should be overriding values from `.json` files in `.js` files. That said, by leaving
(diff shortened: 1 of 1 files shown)


slnode:

pages/en/lb3/Environment-specific-configuration.md
```diff
@@ -21,6 +21,8 @@ LoopBack applications have the following types of configuration files:
 * **Middleware configuration files**, by default `server/middleware.json`.
 * **Configuration files for LoopBack components**, by default `server/component-config.json`.
 
+{% include note.html content="Using `*.js` files to configure the LoopBack application will only work when the original `.json` files are kept in place. Rather than replacing `.json` files with `.js` files, you should be overriding values from `.json` files in `.js` files. That said, by leaving the default `.json` files empty, all configuration can be done in `.js` files." %}
```
crandmck: Due to a quirk with the "note" Jekyll include, you need to put the ending characters on a separate line, otherwise the note doesn't terminate and consumes the remainder of the page! 
e.g. like this:
```
...can be done in `.js` files.
" %}
```

crandmck (changes requested):

pages/en/lb3/Environment-specific-configuration.md
sanderboom: Thanks. I've taken the effort to install ruby/bundler and now actually see the result. *Much* better this way ;)

crandmck (approved):

PR #381, opened 2017-05-15
Lb3/update polymorphic relation

jannyHou:
connect to strongloop/loopback-datasource-juggler#1298

- Update 'hasMany' 'hasOne' 'belongsTo' docs according to https://github.com/strongloop/loopback-datasource-juggler/issues/1298#issuecomment-298596768
- Adjust relation sequences to make it more understandable. E.g. User may want to see how to define belongsTo polymorphic relation right after reading the doc of hasMany polymorphic section.
- Update `hasManyThrough` and `hasAndBelongsToMany`:
  - hasAndBelongsToMany: according to test cases in juggler: https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/relations.test.js#L2650
  - hasManyThrough: we currently lack test cases in juggler, I verified by the example repo(will provide link below)

- I created an example repo(branch, to be more accurate) to demo `hasMany` `hasManyThrough` and `hasAndBelongsToMany`: 
  - link: https://github.com/strongloop/loopback-example-relations/tree/example/polymorphic-relations
  - that branch is still in progress but contains enough models scaffolded for us to play with polymorphic relations: verify, diagnose, find bugs or try something new.

I understand reviewers may have questions like why something is optional, or is there any easier way to define a complete relation, etc.. We will investigate/improve those things in the future, well at this moment, if there is no critical error in this doc, or any concept you feel is wrong, I prefer we can update the doc asap since the juggler PR is released. I verified my changed parts with the example repo, so at least those docs work for sure. We will continue improving relations in the future.
I list some of my questions to be investigated, feel free to add more, questions and answers are welcomed, appreciate any feedback :)

- Why we have an optional invert hasMany relation for `hasManyThrough` and `hasAndBelongsToMany`.
- Does user have to define a through model in `hasAndBelongsToMany`? looks like our code can automatically setup one.


jannyHou:
@ebarault PTAL thanks. Details provided in https://github.com/strongloop/loopback.io/pull/381#issue-228865439

pages/en/lb3/Polymorphic-relations.md
```diff
@@ -34,9 +34,9 @@ This documentation is still a work in progress.
 " %}
 
 LoopBack supports _polymorphic relations_ in which a model can belong to more than one other model, on a single association.
-For example, you might have a Picture model that belongs to either an Author model or a Reader model. 
+For example, you might have a Picture model that belongs to either an Author model or an Reader model.
```
kjdelisle: Not a fix. `an Reader` is not correct.

pages/en/lb3/Polymorphic-relations.md
```diff
@@ -45,7 +45,34 @@ The examples below use three example models: Picture, Author, and Reader, where
 
 ## HasMany polymorphic relations
 
-The usual options apply, for example: `as: 'photos'` to specify a different relation name/accessor.
+Take the following scenario as example:
+
+> A Picture model belongs to either an Author model or a Reader model 
+
+A hasMany polymorphic relation means: there are two properties created in model Picture, one serves as a discriminator, which tells which model the picture actually belongs to, and the other one servers as a foreignKey, which is the id(or primaryKey) value of either the author or reader.
```
kjdelisle: `which tells` is not clear, or grammatically correct. I'd rephrase this:
`A hasMany polymorphic relation means: there are two properties created in model Picture, one serves as a discriminator, which states what model the picture belongs to`

Additionally, you're using things like `belongs to` in a description about hasMany; it could be confusing for users that might think they can just use the two interchangably from either direction, rather than needing both for a two-way relationship.

pages/en/lb3/Polymorphic-relations.md
```diff
@@ -45,7 +45,34 @@ The examples below use three example models: Picture, Author, and Reader, where
...
+
+This section explains how to define a hasMany polymorphic relation in modelFrom(Author or Reader in example). You need to also define the corresponding belongsTo relation described in the next section "BelongsTo polymorphic relations" to establish the relation completely.
```
kjdelisle: This part is redundant: `This section explains how to define a hasMany polymorphic relation`

"If you intend to create a two-way representation of a one-to-many relationship, you will also need to create the `belongsTo` relation on the Model that is the "many" within that relationship. In this example, the Picture model will have two `belongsTo` relationships that reference the Author and Reader models by their primary keys."

pages/en/lb3/Polymorphic-relations.md
```diff
@@ -96,47 +138,140 @@ And:
...
   } 
 });
 ```
 
-## HasManyThrough polymorphic relations
+## BelongsTo polymorphic relations
 
-To define a hasMany polymorphic relation, there must be a "through" model, similarly to a standard [HasManyThrough relation](HasManyThrough-relations.html).
+Because you define the related model dynamically, you cannot declare it up front.
+So instead of passing in the related model (name), you specify the name of the polymorphic relation.
```
kjdelisle: Missing comma after "so":

`So, instead of passing in the related model (name)...`

pages/en/lb3/Polymorphic-relations.md
```diff
@@ -96,47 +138,140 @@ And:
...
+  polymorphic: {
+    foreignKey: 'imageableId',
+    discriminator: 'imageableType'
+  }
+});
+```
+
+## HasOne polymorphic relations
+
+The relation definitions in 'HasOne' is almost same as 'HasMany', the only difference is `type` should be 'hasOne', for details, refer to HasMany polymorphic relations section.
```
kjdelisle: I strongly disagree with the way this is phrased. `hasOne` is similar to `hasMany` only in how you define it on the Model; saying that they're almost the same could easily confuse someone who was seeing it for what it was, which is a one-to-one relationship.

`refer to HasMany polymorphic relations section`
Give them a link to that section using Markdown Syntax:
"Refer to \[HasMany polymorphic relations\]\(#HasMany-polymorphic-relations\) section"

pages/en/lb3/Polymorphic-relations.md
kjdelisle: `described in the next section "BelongsTo polymorphic relations"`
Give them a link:
"...described in the next section \[BelongsTo polymorphic relations\]\(#BelongsTo-polymorphic-relations\) to establish the relationship completely."

pages/en/lb3/Polymorphic-relations.md
```diff
@@ -147,80 +282,77 @@ Then here's an example of a polymorphic hasManyThrough relation:
...
+Picture.hasMany(Author, {through: ImageLink, polymorphic: 'imageable', invert: true});
+Picture.hasMany(Reader, {through: ImageLink, polymorphic: 'imageable', invert: true});
 ```
 
 ## HasAndBelongsToMany polymorphic relations
 
-This requires an explicit 'through' model, in this case: PictureLink
-
-The relations `Picture.belongsTo(PictureLink)` and `Picture.belongsTo('imageable', { polymorphic: true });` will be setup automatically.
+hasAndBelongsToMany polymorphic relation is similar to hasManyThrough relation.
```
kjdelisle: "...similar to hasManyThrough relation."
Should be:
"similar to \[hasManyThrough relation\]\(#HasManyThrough-polymorphic-relations\)."

kjdelisle (changes requested):

pages/en/lb3/Polymorphic-relations.md
jannyHou: good catch. At first I use `article` and `employee` as example models, left some `an` before model name, will clean up those places.

pages/en/lb3/Polymorphic-relations.md
jannyHou: I thought when this pages are displayed, `*.md` files are loaded by other code to generate .html file, so markdown hyperlink may not work. 
I will try it, and there are a few more places need the hyperlink as well.

jannyHou: > The relation definitions in 'HasOne'

@kjdelisle I will rephrase it since it causes your confusion, while actually I explicitly describe it as **The relation definitions** is similar.
I am trying to emphasize user can use same definition parameters in the relation object(just change `type: hasMany` to `type: hasOne`), so I don't need to copy paste the whole definitions again.

Or maybe I can add hasMany & hasOne relation hyperlink here(those would be on a different page) to explain the essential difference between them:`one-to-many` vs `one-to-one`.

Any thought?

jannyHou:
@kjdelisle Applied feedback, thanks!
PTAL again.

pages/en/lb3/Polymorphic-relations.md
```diff
@@ -96,47 +142,143 @@ And:
...
+    foreignKey: 'imageableId',
+    discriminator: 'imageableType'
+  }
+});
+```
+
+## HasOne polymorphic relations
+
+The difference between hasOne and hasMany is, hasOne represents a "one-to-one" relation while hasMany represents a "one-to-many" relation.
+(I am going to add hyperlink of hasOne and hasMany here, haven't figured out how to link to another page :P)
```
kjdelisle: Make sure this is fixed before merging. I'm sure you knew, but we all forget stuff sometimes. :P

kjdelisle (approved):

pages/en/lb3/Polymorphic-relations.md
```diff
@@ -45,7 +45,38 @@ The examples below use three example models: Picture, Author, and Reader, where
...
 ## HasMany polymorphic relations
 
-The usual options apply, for example: `as: 'photos'` to specify a different relation name/accessor.
+Take the following scenario as example:
+
+> A Picture model belongs to either an Author model or a Reader model 
+
+A hasMany polymorphic relation means: 
+
+- There are two properties created in model Picture, one serves as a discriminator, which states what model the picture belongs to, and the other one servers as a foreignKey, which is the id(or primaryKey) value of either the author or reader.
```
b-admike: _serves_ instead of _servers_ here

pages/en/lb3/Polymorphic-relations.md
```diff
@@ -45,7 +45,38 @@ The examples below use three example models: Picture, Author, and Reader, where
...
-The usual options apply, for example: `as: 'photos'` to specify a different relation name/accessor.
+Take the following scenario as example:
+
+> A Picture model belongs to either an Author model or a Reader model 
+
+A hasMany polymorphic relation means: 
+
+- There are two properties created in model Picture, one serves as a discriminator, which states what model the picture belongs to, and the other one servers as a foreignKey, which is the id(or primaryKey) value of either the author or reader. 
+
+- There are CRUD apis added to modelFrom(Author and Reader), by which you can create/modify/delete modelTo instance that belongsTo it. For example, `author.pictures.create()` create a new picture belongsTo author.
```
b-admike: I think rephrasing `For example, 'author.pictures.create()' create a new picture belongsTo author.` to `For example, 'author.pictures.create()' creates a new picture model instance which belongs to the author model.` or something similar would be good.

pages/en/lb3/Polymorphic-relations.md
```diff
@@ -147,80 +289,77 @@ Then here's an example of a polymorphic hasManyThrough relation:
...
 ## HasAndBelongsToMany polymorphic relations
 
-This requires an explicit 'through' model, in this case: PictureLink
-
-The relations `Picture.belongsTo(PictureLink)` and `Picture.belongsTo('imageable', { polymorphic: true });` will be setup automatically.
+hasAndBelongsToMany polymorphic relation is similar to [hasManyThrough polymorphic relation](#HasManyThrough-polymorphic-relations).
+It also requires an explicit 'through' model, in our example: ImageLink
 
-The same is true for the needed properties on PictureLink.
+The difference between them is, hasAndBelongsToMany will automatically setup belongsTo relations in through model: `ImageLink.belongsTo(Picture, {})` and `ImageLink.belongsTo(ImageLink, {polymorphic: true})`.
```
b-admike: `difference between them is,` to `difference between them is that` and `automatically setup` to `automatically set up`

pages/en/lb3/Polymorphic-relations.md
```diff
@@ -96,47 +142,143 @@ And:
...
+
+## HasManyThrough polymorphic relations
+
+To define a hasManyThrough polymorphic relation, there must be a "through" model, for example:
+
+> Author hasMany Picture through ImageLink polymorphically
+>
+> Reader hasMany Picture through ImageLink polymorphically
+
+A hasManyThrough polymorphic relation means the through model has three properties created. The first two are same as those created in the toModel in a hasMany polymorphic relation: one is discriminator(the value is either `Author` or `Reader`) and the other one is foreignKey(the id(primarykey) value of either an author or a reader). The third property is a foreignKey property reference model Picture.
```
b-admike, line 259: It'd help users understand the last sentence more if we specify the type of model it references i.e. `The third property is a foreignKey property which references the modelFrom(?), in this case the Picture model.`

pages/en/lb3/Polymorphic-relations.md
jannyHou: @b-admike lower case `picture` and `author` are already instance :p  Camel case `Picture` and `Author` means model, and I think here a model instance belongs to another model instance(both of them are instantiated), not a model.

pages/en/lb3/Polymorphic-relations.md
```diff
@@ -147,80 +289,77 @@ Then here's an example of a polymorphic hasManyThrough relation:
...
-      }
+      "through": "ImageLink",
+      "invert": true,
+      "polymorphic": "imageable"
+    },
+    "readers": {
+      "type": "hasMany",
+      "model": "Reader",
+      "through": "ImageLink",
+      "invert": true,
```
b-admike, line 326: Ignore if I'm wrong, but I don't see `invert:true` explained to the user above or below the example code.

pages/en/lb3/Polymorphic-relations.md
b-admike: Ah I didn't know that, thanks Janny. `create` to `creates` in that case.

pages/en/lb3/Polymorphic-relations.md
jannyHou: @b-admike I explained it in https://github.com/strongloop/loopback.io/pull/381#issue-228865439

My current understanding is the invert hasMany relation is optional(I think this is explained in doc, I also verified it), but regarding why do we had the optional relation in old doc(and juggler test), I could not say for sure, probably because by doing that you can have the api on model `Picture`

pages/en/lb3/Polymorphic-relations.md
jannyHou: @b-admike Here I could not find a proper word to describe model `Picture` so I use the name, in a hasManyThrough relation, `Author` `Reader` `Picture` are **ALL** `modelFrom`, the through model is actually the  `modelTo`, but I could not think of a descriptive word to distinguish between different `modelFrom`(one is `Author`&`Reader`, the other one is `Picture`)

jannyHou:
@b-admike feedback applied and replied, thanks! PTAL again and take your time if you are still reviewing it.

b-admike (approved):

pages/en/lb3/Polymorphic-relations.md
b-admike: Ok I'm fine with leaving as is in that case.

jannyHou:
All feedback applied and also add the cross file hypelink. I am landing it. Thanks a lot everyone really appreciate your feedback :)

crandmck:
Thanks @jannyHou.  I still find the struck out entries with **NOT EXPECTED** to be confusing.
What are they trying to convey?  If these properties are not supported, then either we should simply not document them, or (if there is a reason to) say not supported.  But why would we document something that's not supported?

Also, I think the info in the lengthy nested bullet lists could be presented more clearly.  Would you rather I do it in this PR/branch or separately after you land this?

In general, it's not a good idea to make "predictions" in docs.  The docs should reflect the way things are, not the way we want or expect things to be.  So, suggest you remove:

> `as` will be deprecated in LoopBack4

Finally.. Since this clarifies polymorphic relations substantially, do you think we can remove the 

>Warning: This documentation is still a work in progress." 

at top of the doc?

jannyHou:
@crandmck Our old doc kind of mixed the definition in `hasManyRelation.polymorphic` object and `belongsToRelation.polymorphic` object. 
E.g. When I saw hasMany.polymorphic has `selector` `discriminator` and `foreignKey`, and belongsTo.polymorphic has `discriminator` and `foreignKey`, I would want to try add a `selector` to belongsTo as well since they look so similar. 
It was misleading, and gave user an impression that they can define `model`  or `polymorphic.selector/as`in belongsTo polymorphic relation, while that's not correct, by providing them, in old code either you had a bug or juggler just ignored that field. In the new code we add validation for them.

crandmck:
I see.  In that case, Instead of the current formatting, I would simply add a note that these properties are not supported for this case, and if provided will throw an error.  That would be appropriate to put in a [warning alert](http://loopback.io/doc/en/contrib/alerts.html#types-of-alerts). 

Also, if you could please respond to my other questions above (I edited the original comment, so perhaps you didn't see them).

ebarault:
Hi all, I'll be able to review the PR end of week for beginning of next week. 

À general comment is that I'd like we ajust the examples to fit my refactor in the unit tests: use Products and Employee as in Rail's polymorphic relations doc, instead of Author and Reader which better apply to standard hasMany relations examples .

jannyHou:
> I would simply add a note that these properties are not supported for this case, and if provided will throw an error.

Cool I will do that now.

> In general, it's not a good idea to make "predictions" in docs. The docs should reflect the way things are, not the way we want or expect things to be. So, suggest you remove:

The new code in juggler has a deprecation warning when user define polymorphic with `as`, I am ok to remove it, or maybe I can declare the deprecation as a note afterwards.

> Also, I think the info in the lengthy nested bullet lists could be presented more clearly. Would you rather I do it in this PR/branch or separately after you land this?

@crandmck There will be more review from @ebarault next week, if you don't mind, how about I check in this PR first(since the relation logic looks good) and create another PR to apply @ebarault 's feedback, and in the second PR, we can discuss more about how to improve it. AFAIK, we are still missing embedsMany/embedsOne polymorphic in doc.

crandmck:
> The new code in juggler has a deprecation warning when user define polymorphic with as, I am ok to remove it, or maybe I can declare the deprecation as a note afterwards.

Is it deprecated now?  Or it will be in v4?  If now, then just say it's deprecated; if not deprecated until v4, then I wouldn't say anything.

> ...how about I check in this PR first(since the relation logic looks good) and create another PR to apply @ebarault 's feedback, and in the second PR, we can discuss more about how to improve it. AFAIK, we are still missing embedsMany/embedsOne polymorphic in doc.

Yes, that's fine.  Thanks.  AFAIC, you can land this when your done with the above.

jannyHou:
@crandmck Now we still accept `as`, but print a warning that LoopBack 4 will stop using it. 
Thanks for feedback :) I will reorganize the definition parameter part and change the example model names in the next coming PR.

PR #385, opened 2017-05-24
Update Using-MongoLab.md

atherdon:
Small update, 'cause we don't have previous link url - mongolab.com
Maybe it'll be necessary to update title and permalink, but not sure.

### DIFF
--- a/pages/en/lb3/Using-MongoLab.md
+++ b/pages/en/lb3/Using-MongoLab.md
@@ -17,7 +17,6 @@ For example, the entry in `datasources.json` might look like this: 
 {% include code-caption.html content="/server/datasources.json" %}
 ```javascript
 "mongodb": {
-  "defaultForType": "mongodb",
   "connector": "loopback-connector-mongodb",
   "url": "mongodb://localhost:27017/mydb"
 }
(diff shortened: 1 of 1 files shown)


slnode:

crandmck:
Thanks!  It would make sense to change the title.  Could you please do this?  Then, I'll land the PR.

The permalink doesn't matter as much, but you can change it if you wish.

atherdon:
Hi @crandmck, ok will do!
I think change permalink is not necessary, because it can ruin google index(maybe).
I propose to add some clarification. On title I'll add "*" sign to make things clear - MongoLab == mLab

atherdon:
Question: I tried to copy paste code from this page to my datasource file and got an issue "Cannot create data source "nameDS": DataSource option "defaultForType" is no longer supported".
So I came to this link: https://stackoverflow.com/a/21655979
I test this solution and it works well for me. Can I change this code sample on readme page too?

crandmck:
Yes, please do remove `defaultforType`.  It was removed in v3 (see https://loopback.io/doc/en/lb3/3.0-Release-Notes.html#loopbackautoattach-removed).

Thanks for testing it and catching this.

atherdon:
Done, please check @crandmck

crandmck:
Thanks!

PR #386, opened 2017-05-25
Doc for excludeBaseProperties

rashmihunt:
I have added documentation detailing excludeBaseProperties. 
@crandmck @raymondfeng PTAL

### DIFF
--- a/pages/en/lb3/Model-definition-JSON-file.md
+++ b/pages/en/lb3/Model-definition-JSON-file.md
@@ -584,7 +584,8 @@ For example, to map a property to a column in an Oracle database table, use the 
 
 ## Exclude properties from base model
 
-By default, a model inherits all properties from the base. To exclude some base properties from being visible, you need to set excludeBaseProperties = []
+By default, a model inherits all properties from the base. To exclude some base properties from being visible, you need to set `excludeBaseProperties = ['property-be-excluded-from-base-model']`.
+`excludeBaseProperties` is recommended approach over previous approach of setting the base property to 'null' or 'false'. 
 
 For example,
 {% include code-caption.html content="common/models/customer.json" %}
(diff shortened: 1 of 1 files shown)


raymondfeng:
Do we know why this PR includes `pages/en/lb3/readmes/loopback-connector-cassandra.md`?

pages/en/lb3/Model-definition-JSON-file.md
```diff
@@ -575,9 +584,34 @@ For example, to map a property to a column in an Oracle database table, use the
 
 ## Exclude properties from base model
 
-By default, a model inherits all properties from the base. To exclude some base properties from being visible, you need to set the property to `false` or `null`.
+By default, a model inherits all properties from the base. To exclude some base properties from being visible, you need to set excludeBaseProperties = []
```
raymondfeng: Maybe `set excludeBaseProperties = ['property-be-excluded-from-base-model']`.

I guess we should not remove **set the property to `false` or `null`**. We should state that `excludeBaseProperties` is now the recommended way.

crandmck:
> Do we know why this PR includes pages/en/lb3/readmes/loopback-connector-cassandra.md?

I noticed that, too.  

Normally, READMEs from other repos such as this are automatically pulled down from GitHub by the Jenkins pipeline.  You can see that the commit is from  "slnode" https://github.com/strongloop/loopback.io/pull/386/commits/85a91ca22377df98ef40863e86c0eb2a45675b4d

I'm not sure how it got into this PR, but I don't think it will cause a problem.  However, if you want to remove it from the PR, that would be OK, too.  Jenkins should eventually sync it with the `loopback-connector-cassandra` repo.

> I guess we should not remove set the property to false or null. We should state that excludeBaseProperties is now the recommended way.

I think this PR does still include the doc for "old" way, and has this:

" Below way of excluding base properties by setting the base property to 'null' or 'false' has been deprecated. Instead, use excludeBaseProperties as shown above."

So, it LGTM.

rashmihunt:
@crandmck @raymondfeng I have no idea how the loopback-connector-cassandra.md got committed. I had committed only one file  Model-definition-JSON-file.md file for this PR. Some magic has happened internally looks like.

@raymondfeng I added old way as deprecated based on the conversation we had couple of days ago where I had asked what do we do with the old way?

rashmihunt:
@raymondfeng Fixed as per what you mentioned
1) excludeBaseProperties = ['property-be-excluded-from-base-model']
2) Removed the word 'deprecate' as used "recommended" and also mentioned the old approach at the beginning of the section as you wanted.

raymondfeng (approved):
LGTM

Issue #332, opened 2017-03-21
Clean up datasource configuration docs


crandmck:
From comment by @ebarault in #319:

.. the accessibility of this information could be improved.
As shown below there are two sections discussing the subject: "Attaching models to datasources" and "Understanding datasources", in addition "Understanding datasources" is nested in a section named "Connecting models to datasources", which is much too close to "Attaching models to datasources".

* Defining models
** Attaching models to datasources

* Connecting models to datasources
** Understanding datasources

![image](https://cloud.githubusercontent.com/assets/12000935/24022525/516fca34-0aa7-11e7-876d-94ccefdab6ea.png)

Ideally datasources would require their own top section, and would fit just before Defining models.
The content of **Defining datasources** would be extracted from the two sub-sections Attaching models to datasources and Understanding datasources.
The **Connecting models to datasources** section's name should change. It is as of now a melting pot of advanced details and functionalities.
The applicable content would fit in a new top-level **Attached models to datasources**

So the logical sequence would look like

* Defining datasources
** Datasources: Advanced topics
* Defining models
* Attaching models to datasources

PR #395, opened 2017-06-07
Reorg of data source and some model docs

crandmck:
Addresses #332.   Eliminates a lot of duplication between "Attaching models to datasources" and  "Connecting models to datasources", and generally improves organization, etc.

Small difference in what's suggested there, the nav now looks like this:
- Defining data sources
- Defining models
- Using built-in models
- Attaching models to data sources

I also consolidated "Creating models" and "Defining models" and made a number of other small improvements.


crandmck:
Landing this, since it does not require technical review.  It's just reorg and streamlining.

PR #396, opened 2017-06-08
Add info on debug log for users in production

ssh24:
connect to https://github.com/strongloop-internal/scrum-loopback/issues/1119

### DIFF
--- a/pages/en/contrib/Reporting-issues.md
+++ b/pages/en/contrib/Reporting-issues.md
@@ -43,7 +43,7 @@ There are three types of issues:
 
 We take security vulnerabilities in LoopBack very seriously for our users.
 
-It is recommended to never debug log anything in production that may contain sensitive data. For example, do not log the data payload for `PersistedModel.create` requests as it may contain sensitive data sent from the client.
+In production, do not enable debug logs that may contain sensitive data; for example, the payload of `PersistedModel.create` should not be exposed in production. Logging this type of information is acceptable in development, but not in production.
 
 Do not report security vulnerabilities using GitHub issues. Please send an email to `callback@ibm.com` with:
 
--- a/pages/en/contrib/Reporting-issues.md
+++ b/pages/en/contrib/Reporting-issues.md
@@ -41,6 +41,10 @@ There are three types of issues:
 
 ### Security issues
 
+We take security vulnerabilities in LoopBack very seriously for our users.
+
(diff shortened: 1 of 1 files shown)


pages/en/lb3/readmes/loopback-connector-cloudant.md
```diff
@@ -561,6 +585,71 @@ module.exports = function(server) {
 };
 ```
 
+# Bulk replace
+
+Given an array of data to be updated, Cloudant supports the idea of performing bulk replace on a model instance.
```
bajtos: I am proposing to add a note that bulk replace is not invoking any operation hooks.

pages/en/contrib/Reporting-issues.md
```diff
@@ -41,6 +41,10 @@ There are three types of issues:
 
 ### Security issues
 
+We take security vulnerabilities in LoopBack very seriously for our users.
+
+It is recommended to never debug log anything in production that may contain sensitive data. For example, do not log the data payload for `PersistedModel.create` requests as it may contain sensitive data sent from the client.
```
bajtos: This is not the advice I had in mind.

We are recommending to never turn on debug logs on production, because debug logs often contain sensitive data sent by the client.

It is ok to debug-log the payload of `PersistedModel.create` if debug logs are used only at development time.

bajtos (changes requested):

pages/en/contrib/Reporting-issues.md
ssh24: Reworded the sentence to sound better.

pages/en/contrib/Reporting-issues.md
```diff
@@ -41,6 +41,10 @@ There are three types of issues:
 
 ### Security issues
 
+We take security vulnerabilities in LoopBack very seriously for our users.
+
+It is recommended to never turn on debug logs on production that may contain sensitive data sent by the client. For example, it is ok to debug-log the payload of `PersistedModel.create` if debug logs are used only at development time.
```
crandmck: Suggest rewording:

In production, do not enable debug logs that may contain sensitive data; for example, the payload of `PersistedModel.create` might contain passwords and thus should not be exposed in production.  Logging this type of information is acceptable in development, but not production.

ssh24: Fixed.

bajtos (approved):

crandmck (approved):

PR #399, opened 2017-06-14
Document relations options property

frankorama:
Documenting the options property that can be defined in a model relation.  Thus far I have only found one property for options.

### DIFF
--- a/pages/en/lb2/Model-definition-JSON-file.md
+++ b/pages/en/lb2/Model-definition-JSON-file.md
@@ -741,24 +741,9 @@ For example:
       <td>Name of model creating hasManyThrough relation. See example below.</td>
     </tr>
     <tr>
-      <td>options</td>
-      <td>object</td>
-      <td>
-        <table>
-	  <tbody>
-	    <tr>
-	      <th>Key</th>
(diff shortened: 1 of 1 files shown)


slnode:

crandmck:
@frankorama Thanks for the PR.  

Did you mean to add this only for v2?  I wouldn't think it's specific to v2.
To also add it to LoopBack v3 docs, edit `pages/en/lb3/Model-definition-JSON-file.md`.

Also, rather than putting the options in a "table within a table" which is awkward and can be hard to read, since there's only one options property, I suggest just documenting the property as `options.disableInclude`.  

And one small nitpick: Boolean is always capitalized :-)

frankorama:
@crandmck Thanks for the feedback.  My first pull-request.  I'll make the changes to v2 (and change the case on 'Boolean')

Cheers.

crandmck:
Thanks!  Could you please also address this:
> rather than putting the options in a "table within a table" which is awkward and can be hard to read, since there's only one options property, I suggest just documenting the property as options.disableInclude.

crandmck:
Great, LGTM!

PR #403, opened 2017-06-21
Update Creating-a-database-schema-from-models.md

ehtasham:


slnode:

crandmck:
Thanks for the PR, but I think that code is in fact JavaScript and not JSON due to the `var schema_v1 = ... ` part.

But I think it does need some changes.  The first code block (with `var schema_v1 = { ... }` ) should probaby just be in the same JS file as the second one, and it should be called `/common/models/create-schema.js` to clarify what it does.

Would you be willing to change your PR to have something like that?

crandmck:
@ehtasham Could you please address my comment above?  If not, then I'll close this PR.

PR #413, opened 2017-07-05
Update Bluemix docs

hacksparrow:
Added new content or updated existing content about Bluemix integration.

PR #421, opened 2017-07-18
Add Bluemix integration docs

hacksparrow:
Added new content or updated existing content about Bluemix integration.

Opened in place of #413 (messed up history).


hacksparrow:
@bschrammIBM @crandmck this one can be landed.

PR #422, opened 2017-07-18
Correct where filter docs

loay:
connect to https://github.com/strongloop/loopback/issues/2196


dhmlau:
@loay, would this change applicable to lb2/Where-filter.md as well?

superkhau (approved):

PR #423, opened 2017-07-21
Fixed logic error in Extend your API page

thescientist13:
### Overview
We need to use `currentHour >= OPEN_HOUR` so that it says the shop is open
when the currentHour is between 06:00 and 07:00.

Authored by @sumnerevans on behalf of Kenzan (Labs)


slnode:

PR #433, opened 2017-08-05
Documentation fix

TwistedLogic:
That looks like a copy/paste from another page.


slnode:

crandmck:
Thanks!

For future reference, it's best practice to align v3 and v2 docs, with the exception that updating v3 docs doesn't require updating v2 docs (since v3 is the current release).  In general, we don't spend a lot of time updating the v2 docs, but fixes are welcome.

I'll update https://loopback.io/doc/en/lb3/HasMany-relations.html to have this change, too.

TwistedLogic:
Roger that. Thanks.

PR #442, opened 2017-08-28
Shortened install on win

buokae:
Replaced manual compiler tools installation instructions with windows-build-tools package.


slnode:

buokae:
I also need to download and install Win64 OpenSSL v1.0.2L from https://slproweb.com/products/Win32OpenSSL.html, so I thought about adding that to the docs too but I don't know how the admins feel about that. After this I am left with just three warnings when running 'npm install -g loopback-cli', which is good enough I think.

crandmck:
Thanks!  This looks like an easier process....

PR #445, opened 2017-08-30
Fix typo

osk2:
Model path should be `common/models` not `common/model`


slnode:

crandmck:
Good catch, thanks!

PR #447, opened 2017-09-04
Update Create-a-simple-API.md

clearnz:
Line return after first sentence for consistency.


slnode:

crandmck:
Thanks for catching! Now that I look at this, I'm going to remove the colon after the first sentence, because it doesn't work as is.

PR #448, opened 2017-09-07
Affix sidebar, add scroll

christophherr:
Iteration to affix the sidebar vertically and to show a scrollbar when the sidebar content is longer than the window/viewport height.
Related issue: #425

PR #451, opened 2017-09-11
Fix sidebar and add scrollbar

crandmck:
@christophherr This is largely based on your PR to this branch: #448.  I've tested it locally and it seems to work quite well, but thought I'd open a PR for your review and any other tweaks that might be needed.  PTAL!

### DIFF
--- a/css/lavish-bootstrap.css
+++ b/css/lavish-bootstrap.css
@@ -717,7 +717,7 @@ pre code {
 .col-sm-12,
 .col-md-1,
 .col-md-2,
-/* .col-md-3, To make sidebar fixed */
+.col-md-3
 .col-md-4,
 .col-md-5,
 .col-md-6,
@@ -5895,13 +5895,4 @@ td.visible-print {
   td.hidden-print {
     display: none !important;
   }
-}
-
-/* To make sidebar fixed */
-.col-md-3
-{
-  position: fixed;
-  min-height: 1px;
-  padding-left: 15px;
-  padding-right: 15px;
 }
--- a/css/lavish-bootstrap.css
+++ b/css/lavish-bootstrap.css
@@ -717,7 +717,7 @@ pre code {
 .col-sm-12,
 .col-md-1,
 .col-md-2,
-.col-md-3,
+/* .col-md-3, To make sidebar fixed */
 .col-md-4,
 .col-md-5,
 .col-md-6,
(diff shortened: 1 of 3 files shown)


crandmck:
I'm going to go ahead and land this.  I don't expect any problems.... Thanks for your help @christophherr !

css/lavish-bootstrap.css
```diff
@@ -717,7 +717,7 @@ pre code {
 .col-sm-12,
 .col-md-1,
 .col-md-2,
-.col-md-3,
+.col-md-3
```
damache, line 720: missing comma, think this is causing the sidebar to over extend into the page
screenshot from 2017-09-12 13-01-03

just happen within the hour.

crandmck: I checked in a correction for this.... See if that fixes it.

damache:
no.  same thing.  I tried it in chrome too.  
screenshot from 2017-09-12 15-10-44

damache:
I have a page open still before the change was pushed.  the computed size of the nav bar before is
screenshot from 2017-09-12 15-19-06

and after
screenshot from 2017-09-12 15-19-40

crandmck:
Unfortunately, I can't reproduce the problem, even using FireFox. :-(

christophherr:
@damache What is your screen size / screen resolution?
I get a similar result with zoom set to 80% (in Chrome)
http://prntscr.com/gkfpg4
We might just need to add more media queries, e.g.
```css

@media (min-width: 1410px) {
    .col-md-3 {
        width: 19%;
    }
}

@media (min-width: 1680px) {
    .col-md-3 {
        width: 17%;
    }
}

@media (min-width: 1880px) {
    .col-md-3 {
        width: 14%;
    }
}
```

damache:
@christophherr @crandmck it's still happening and my browser is at 100%.  my screen is 1920 x 1080.  it happens when the sidebar is expanded and the slider appears.  
screenshot from 2017-09-13 09-49-23

notice in the image the sidebar is extended beyond the view

today I'm on my laptop and yesterday I was on my monitor at work.  think the same resolution

crandmck:
OK, I can now reproduce this problem.  It occurs on FF only when you make your browser window REALLY wide, like 1800px+ ....  For me, it does not occur for widths less than that.  As a workaround, just make your browser window narrower.

PR #453, opened 2017-09-11
Google cloud

crandmck:
Add Google Cloud provider info per https://github.com/strongloop/loopback-component-storage/issues/199

### DIFF
--- a/pages/en/lb3/Storage-component.md
+++ b/pages/en/lb3/Storage-component.md
@@ -17,9 +17,10 @@ makes it easy to upload and download files to cloud storage providers and the lo
 It has Node.js and REST APIs for managing binary content in cloud providers, including:
 
 * Amazon
-* Rackspace
-* Openstack
 * Azure
+* Google Cloud
+* Openstack
+* Rackspace
 
 You use the storage component like any other LoopBack data source such as a database.
 Like other data sources, it supports create, read, update, and delete (CRUD) operations with exactly the same LoopBack and REST APIs.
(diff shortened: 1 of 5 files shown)


crandmck:
@overdrivr PTAL

pages/en/lb3/Storage-component.md
```diff
@@ -247,7 +249,31 @@ as shown in the following table.
...
+    <tr>
+      <td>keyFilename</td>
+      <td>Path to key file.</td>
+    </tr>
+    <tr>
+      <td>projectID</td>
+      <td>Google Cloud project ID.</td>
+    </tr>
+    <tr>
+      <td>nameConflict</td>
```
Overdrivr: This option is available to all providers, it is implemented [here](https://github.com/strongloop/loopback-component-storage/blob/98ded2eed24b3dd4af5a2fbc80f90d7cfa8318be/lib/storage-handler.js#L89) but I don't recall seeing it mentioned it the docs. So maybe this option should be mentioned in a more general manner.

Overdrivr:
I've made a quick feedback, there is also a list of providers on the README, it could be nice to add gcloud here too.

crandmck:
@Overdrivr I don't see your feedback....?  

I opened https://github.com/strongloop/loopback-component-storage/pull/222 to update the README.

Also, sorry, it looks like my commits for #452 got into this PR, too... You can ignore those.

Overdrivr:
Sorry I wrote the review but forgot to submit, should be good now ;)

pages/en/lb3/Storage-component.md
crandmck: Ah, right.  It is actually documented here https://loopback.io/doc/en/lb3/Storage-component.html#automatic-unique-filenames.

Since it's not unique to Google, I'll remove it from the table entry.

PR #454, opened 2017-09-12
Fix guideline mistake, "connector" should be "loopback-component-stor…

haotangio:
…age" instead of "storage"


slnode:

crandmck:
Thanks!

PR #459, opened 2017-09-13
initialize page directory for lb4 zh

haoxins:


crandmck:
Thanks for the translation!  

I'll land this, but I _strongly_ suggest you hold off translating any other LB4/LB-Next docs because we are still writing them (in English) and the content is likely to change substantially up til GA release next year.

haoxins:
>  we are still writing them (in English) and the content is likely to change substantially up til GA release next year

got it :)

PR #460, opened 2017-09-13
Change links to API docs to apidocs.loopback.io

crandmck:
Updates all links to API docs to the new URL http://apidocs.loopback.io.

Although straightforward, this touches a large number of files.


slnode:

PR #463, opened 2017-09-18
Decorator doc

jannyHou:
connect to strongloop/loopback-next#552

### DIFF
--- a/pages/en/lb4/Decorators.md
+++ b/pages/en/lb4/Decorators.md
@@ -1,0 +1,397 @@
+---
+lang: en
+title: 'Decorators'
+keywords: LoopBack 4.0, LoopBack-Next
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Decorators.html
+summary:
+---
+
+A decorator allows you to annotate or modify your class declarations and members with metadata.
+
+## Introduction
+
+*If you're new to Decorators in TypeScript, see [here](https://www.typescriptlang.org/docs/handbook/decorators.html) for more info.*
+
+Decorators give LoopBack the flexibility to modify your plain TypeScript classes 
+and properties in a way that allows the framework to better understand how to 
+make use of them, without the need to inherit base classes or add functions 
+that tie into an API.
+
+As a default, LoopBack comes with some pre-defined decorators:
+
+- [Route Decorators](#route-decorators)
+- [Dependency Injection](#dependency-injection)
+- [Authentication Decorator](#authentication-decorators)
+- [Repository Decorators](#repository-decorators)
+
+## Route Decorators
+
+Route decorators are used to expose controller methods as REST API operations.
+If you are not familiar with the concept Route or Controller, please see [LoopBack Route](routes.htm)
+and [LoopBack Controller](controllers.htm) to learn more about them.
+
+By calling a route decorator, you provide OpenAPI specification to describe the 
+endpoint which the decorated method maps to. You can choose different decorators 
+accordingly or do a composition of them:
+
+### API Decorator
+
+  Syntax: [`@api(spec: ControllerSpec)`](http://apidocs.loopback.io/@loopback%2fcore/#783)
+
+  `@api` is a decorator for controller constructor, it's called before a controller
+  class. `@api` is used when you have a base path and a Paths Object, which
+  contains all path definitions of your controller. Please note the api specs defined 
+  with `@api` will override other api specs defined inside the controller. For example:
+
+  ```ts
+    @api({
+      basePath: '/',
+      paths: {
+        '/greet': {
+          get: {
+            'x-operation-name': 'greet',
+            'x-controller-name': 'MyController',
+            parameters: [{name: 'name', type: 'string', in: 'query'}],
+            responses: {
+              '200': {
+                description: 'greeting text',
+                schema: {type: 'string'},
+              }
+            }
+          }
+        }
+      }
+    })
+    class MyController {
+      // The operation endpoint defined here will be overriden!
+      @get('/foo')
+      @param.query.number('limit')
+      greet(name) {
+      }
+    }
+    app.controller(MyController);
+  ```
+
+  A more detailed explanation can be found in [Specifying Controller APIs](controllers.htm#specifying-controller-apis)
+
+### Operation Decorator
+
+  Syntax: [`@operation(verb: string, path: string, spec?: OperationObject)`](http://apidocs.loopback.io/@loopback%2fcore/#818)
+
+  `@operation` is a controller method decorator. It exposes a Controller method as 
+  a REST API operation. You can specify the verb, path, parameters and response 
+  as specification of your endpoint, for example:
+
+  ```ts
+  const spec = {
+    parameters: [{name: 'name', type: 'string', in: 'query'}],
+    responses: {
+      '200': {
+        description: 'greeting text',
+        schema: {type: 'boolean'},
+      }
+    }
+  };
+  class MyController {
+    @operation('HEAD', '/checkExist', spec)
+    checkExist(name) {
+    }
+  }
+  ```
+  
+### Commonly-used Operation Decorators
+  
+  Syntax: [`@get(path: string, spec?: OperationObject)`](http://apidocs.loopback.io/@loopback%2fcore/#798)
+  
+  Same Syntax for decorators [`@post`](http://apidocs.loopback.io/@loopback%2fcore/#802)
+  , [`@put`](http://apidocs.loopback.io/@loopback%2fcore/#806)
+  , [`@patch`](http://apidocs.loopback.io/@loopback%2fcore/#810)
+  , [`@del`](http://apidocs.loopback.io/@loopback%2fcore/#814)
+
+  You can call these sugar operation decorators as a shortcut of `@operation`, for example:
+  ```ts
+  class MyController {
+    @get('/greet', spec)
+    greet(name) {
+    }
+  }
+  ```
+
+  is equivalent to
+
+  ```ts
+  class MyController {
+    @operation('GET', '/greet', spec)
+    greet(name) {
+    }
+  }
+  ```
+  
+  For more usage, refer to [Routing to Controllers](controllers.htm#routing-to-controllers)
+
+### Parameter Decorator
+
+  Syntax: [`@param(paramSpec: ParameterObject)`]()
+
+  Optional Pattern: `@param.${in}.${type}(${name})`
+
+  `@param` can be applied to method itself or specific parameters, it describes
+  an input parameter of a Controller method.
+
+  For example:
+
+  ```ts
+    // example 1: decorator `@param ` applied on method level
+    class MyController {
+      @get('/')
+      @param(offsetSpec)
+      @param(pageSizeSpec)
+      list(offset?: number, pageSize?: number) {}
+    }
+  ```
+
+  or
+  
+  ```ts
+    // example 2: decorator `@param` applied on parameter level
+    class MyController {
+      @get('/')
+      list(
+        @param(offsetSpec) offset?: number,
+        @param(pageSizeSpec) pageSize?: number,
+      ) {}
+    }
+  ```
+
+  In the first example, we apply multiple decorators to a single declaration. 
+  The order in which the decorators are called is important because a `@param` 
+  decorator must be applied after an operation decorator. 
+  To learn more about TypeScript decorator composition, 
+  refer to [TypeScript Decorator Documentation](https://www.typescriptlang.org/docs/handbook/decorators.html)
+
+  Please note method level `@param` and parameter level `@param` are mutually exclusive,
+  you can not mix and apply them to the same parameter.
+  
+  You can also use this pattern to make it quicker to define params: `@param.${in}.${type}(${name})`
+  
+  - in: one of the following values: `query`, `header`, `path`, `formData`, `body`
+  - type: one of the following values:  `string`, `number`, `boolean`, `integer`
+  - name: a `string`, name of the parameter
+
+  So an example would be `@param.query.number('offset')`. 
+  You can find the specific usage in [Writing Controller methods](controller.htm#writing-controller-methods)
+  
+## Dependency Injection
+
+`@inject` is a decorator to annotate method arguments for automatic injection by 
+LoopBack's IoC container. 
+
+The injected values are applied to a constructed instance, so it can only be used on 
+non-static properties or constructor parameters of a Class.
+
+The `@inject` decorator allows you to inject dependencies bound to any implementation 
+of the [Context](#context) object, such as an Application instance or a request context instance.
+You can bind values, class definitions and provider functions to those contexts and 
+then resolve values (or the results of functions that return those values!) in other 
+areas of your code.
+
+```ts
+// application.ts
+import {Application} from '@loopback/core';
+import 'fs-extra';
+class MyApp extends Application {
+  constructor() {
+    super();
+    const app = this;
+    const widgetConf = JSON.parse(fs.readSync('./widget-config.json'));
+    function logInfo(info) {
+      console.log(info);
+    }
+    app.bind('config.widget').to(widgetConf);
+    app.bind('logger.widget').to(logInfo);
+  }
+}
+```
+
+Now that we've bound the 'config.widget' key to our configuration object, 
+and 'logger.widget' key to the function `logInfo()`,
+we can inject them in our WidgetController:
+
+```ts
+// widget-controller.ts
+import {widgetSpec} from '../apispec';
+@api(widgetSpec)
+class WidgetController {
+  // injection for property
+  @inject('logger.widget')
+  private logger: Function
+  
+  // injection for constructor parameter
+  constructor(
+    @inject('config.widget') protected widget: any
+    // This will be resolved at runtime!
+  ) {}
+  // etc...
+}
+```
+
+For more information, see the [Dependency Injection](Dependency-Injection.htm) section under [LoopBack Core Concepts](Concepts.htm)
+
+## Authentication Decorator
+
+  Syntax: `@authenticate(strategyName: string, options?: Object)`
+
+  Marks a controller method as needing an authenticated user. 
+  This decorator requires a strategy name as a parameter.
+
+  Here's an example using 'BasicStrategy': to authenticate user in function `whoAmI`:
+
+  ```ts
+  // my-controller.ts
+  import { authenticate } from '@loopback/authentication';
+  import { inject } from '@loopback/context';
+
+  class MyController {
+    constructor(
+      @inject(BindingKeys.Authentication.CURRENT_USER)
+      private user: UserProfile,
+    ) {}
+
+    @authenticate('BasicStrategy')
+    async whoAmI() : Promise<string> {
+      return this.user.id;
+    }
+  }
+  ```
+
+## Repository Decorators
+
+  As a Domain-driven design concept, 
+  the repository is a layer between your domain object and data mapping layers 
+  using a collection-like interface for accessing domain objects. 
+  
+  In LoopBack, a domain object is usually a TypeScript/JavaScript Class instance, 
+  and a typical example of a data mappting layer module could be a database's node.js driver.
+  
+  LoopBack repository encapsulates your TypeScript/JavaScript Class instance, 
+  and its methods that communicate with your database.
+  It is an interface to implement data persistence. 
+
+  Repository decorators are used for defining models(domain objects) for use with your chosen datasources, 
+  and the navigation strategies among models.
+
+  If you are not familiar with repository related concepts like `Model`, `Entity` and `Datasource`,  
+  please see LoopBack concept [Repositories](#Repositories.htm) to learn more. 
+  
+### Model Decorators
+
+  Model is a Class that LoopBack builds for you to organize the data that 
+  share same configurations and properties.
+  You can use model decorators to define a model and its properties.
+
+#### Model Decorator
+
+  Syntax: `@model(definition?: ModelDefinitionSyntax)`
+
+  Model decorator is a Class decorator.
+  In LoopBack 4, we inherit the model definition format from LoopBack 3,
+  you can find it in [Model definition JSON file](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html).
+  For usage examples, see [Define Models](Repositories.html#define-models)
+
+  *Please note we will elaborate more about model and model definition in #Model.htm,*
+  *and replace the link above with LB4 link*
+
+  By using a model decorator, you can define a model as your repository's metadata,
+  then you have two choices to create the repository instance:
+  
+  One is to inject your repository and resolve it with Legacy Juggler that complete 
+  with CRUD operations for accessing the model's data.
+  A use case can be found in section [Repository decorator](#repository-decorator)
+
+  The other one is defining your own repository without using legacy juggler, 
+  and use an ORM/ODM of your choice.
+
+   ```ts
+  // Missing example here
+  // Will be provided in Model.htm
+  // refer to [example code](https://github.com/strongloop/loopback-next-example/blob/master/services/account-without-juggler/repositories/account/models/Account.ts)
+  ```
+
+#### Property Decorator
+
+  Syntax: `@property(definition: PropertyDefinition)`
+
+  The property decorator defines metadata for a property on a Model definition.
+  The format of property definitions can be found in [Property definitions](https://loopback.io/doc/en/lb2/Model-definition-JSON-file.html#properties) 
+
+  For usage examples, see [Define Models](Repositories.html#define-models)
+
+### Relation Decorators
+
+  The relation decorator defines the nature of a relationship between two models.
+
+  *This feature has not yet been released in alpha form. Documentation will be* 
+  *added here as this feature progresses.*
+
+#### Relation Decorator
+
+  Syntax: `@relation`
+  
+  Register a general relation.
+
+#### Specfic Relation Decorator
+  
+  Syntax:
+
+  - `@belongsTo` 
+  - `@hasOne` 
+  - `@hasMany` 
+  - `@embedsOne` 
+  - `@embedsMany` 
+  - `@referencesOne` 
+  - `@referencesMany`
+
+  Register a specific relation
+
+### Repository Decorator
+
+  Syntax: `@repository(model: string | typeof Entity, dataSource?: string | juggler.DataSource)`
+
+  This decorator either injects an existing repository or creates a repository 
+  from a model and a datasource.
+
+  The injection example can be found in [Repository#controller-configuration](Repositories.html#controller-configuration)
+
+  To create a repository in a controller, you can define your model and datasource 
+  first, then import them in your controller file:
+
+  *To learn more about creating model and datasource, please see the example in [Thinking in LoopBack](Thinking-in-LoopBack.htm#define-product-model-repository-and-data-source)*
+  
+  ```ts
+  // my-controller.ts
+  import { Todo } from '{path_of_Todo_model}.ts';
+  import { datasource } from '{path_of_datasource}.ts';
+
+  export class TodoController {
+    @repository(Todo, datasource)
+    repository: EntityCrudRepository<Todo, number>;
+    ... ...
+  }
+  ```
+  
+  If the model or datasource is already bound to the app, you can create the 
+  repository by providing their names instead of the classes. For example:
+
+  ```ts
+  // with `datasource` and `Todo` already defined.
+  app.bind('datasources.ds').to(datasource);
+  app.bind('repositories.todo').to(Todo);
+
+  export class TodoController {
+    @repository('todo', 'ds')
+    repository: EntityCrudRepository<Todo, number>;
+    // etc
+  }
+  ```
--- a/_data/sidebars/lb4_sidebar.yml
+++ b/_data/sidebars/lb4_sidebar.yml
@@ -47,6 +47,10 @@ children:
   - title: 'Repositories'
     url: /doc/en/lb4/Repositories.html
     output: 'web, pdf'
+  
+  - title: 'Decorators'
+    url: /doc/en/lb4/Decorators.html
+    output: 'web, pdf'
 
 - title: 'Getting started'
   url: /doc/en/lb4/Getting-started.html
--- a/pages/en/lb4/Controllers.md
+++ b/pages/en/lb4/Controllers.md
@@ -154,5 +154,5 @@ class HelloController {
 ```
 
 - A `Repository` is LoopBack's database abstraction. See [Repositories](Repositories.html) for more.
-- `@post('/messages')` creates the `Route` for the Operation using `app.route()`.
+- `@get('/messages')` creates the `Route` for the Operation using `app.route()`.
 - `@param.query.number` adds a `number` param with a source of `query`.


pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Decorators.html
+summary:
+---
+
+A decorator allows you annotate or modify your class declarations and members with metadata.
+
+## Introduction
+
+*To get familiar with this TypeScript feature, try read [Understand TypeScript Decorator](#understand-typescript-decorator)*
```
kjdelisle: `If you're new to Decorators in TypeScript, see [here](https://www.typescriptlang.org/docs/handbook/decorators.html) for more info.`

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+
+Decorator is a good strategy to keep your code loosely coupled and modifies metadata at runtime. Currently LoopBack core provides built-in decorator functions you can use directly by `@myDecorator`, e.g. Route decorators that introduces in section Route-Decorators, and implements a dependency injection system that can apply a decorator by `@inject('myDecorator')`.
```
kjdelisle: `Decorators give LoopBack the flexibility to modify your plain TypeScript classes and properties in a way that allows the framework to better understand how to make use of them, without you needing to inherit base classes or add functions that tie into an API.`

kjdelisle: My suggestion is that we don't plug the virtues of decorators by themselves, but rather, how _we_ use them to make your life easier.

kjdelisle: I would also split the `Currently, LoopBack core provides...` section into its own line.
(It also makes it easier to review it in smaller pieces!)

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+
+A decorator allows you annotate or modify your class declarations and members with metadata.
+
+## Introduction
+
+*To get familiar with this TypeScript feature, try read [Understand TypeScript Decorator](#understand-typescript-decorator)*
+
+Decorator is a good strategy to keep your code loosely coupled and modifies metadata at runtime. Currently LoopBack core provides built-in decorator functions you can use directly by `@myDecorator`, e.g. Route decorators that introduces in section Route-Decorators, and implements a dependency injection system that can apply a decorator by `@inject('myDecorator')`. 
+
+Some components also exports decorators for you:
```
kjdelisle: `As a default, LoopBack comes with some pre-defined decorators for your use:`

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+- [authentication](#authentication-decorators)
+- [repository](#repository-decorators)
+
```
kjdelisle: There should be a bullet point at the top of this list pointing to `@loopback/context`, which is what provides
the `@inject` decorator (arguably the most important one!)

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+*Janny: Most examples are documented in a certain concept that a decorator works closely with, I am not sure should I turn this page to something like a reference:*
+
+- some usage are well documented, I linked to certain section
+- some usage can be found in unit test, while integration example is missing, should I create on this page?
+- some have integration test, modify it and add on this page?
+- Do I need to document what metadata each decorator adds to the reflector and document in the reference?
+- I am confused with some decorator can be used by either @mydecorator or @inject(‘mydecorator’), are they equal?
```
kjdelisle: _some usage can be found in unit test, while integration example is missing, should I create on this page?_

No, if we're going to create the decorator usage docs for particular decorators, they belong under the relevant sections for those libraries.

_some have integration test, modify it and add on this page?_

The examples you create here should be simple, and should be relatable. Creating an example for injection should be a mixture of code snippets and explanation:

> The `@inject` decorator allows you inject dependencies bound to the application context, or
> to any implementation of the [Context](#context) object. You can bind values, class definitions and 
> provider functions to those contexts and then resolve values (or the results of functions that return those 
> values!) in other areas of your code.
> 
> ```ts
> // application.ts
> import {Application} from '@loopback/core';
> import 'fs-extra';
> class MyApp extends Application {
>   constructor() {
>     super();
>     const app = this;
>     const widgetConf = JSON.parse(fs.readSync('./widget-config.json'));
>     app.bind('config.widget').to(widgetConf);
>   }
> }
> ```
> Now that we've bound the 'config.widget' key to our configuration object, we can inject it in our
> WidgetController:
> 
> ```ts
> // widget-controller.ts
> import {widgetSpec} from '../apispec';
> @api(widgetSpec)
> class WidgetController {
>   @inject('config.widget')
>   protected config: JSON; // This will be resolved at runtime!
>   
>   // etc...
> }
> ```

_I am confused with some decorator can be used by either `@mydecorator` or `@inject(‘mydecorator’)`, are they equal?_

They are not equivalent; the first format only modifies the target being decorated (an implicit parameter). The second format will take one more additional arguments, and return a function whose format is exactly like the first decorator's design.

```ts
/**
 * Make a class produce readonly instances.
 * @param {T} The class definition to modify.
 * @returns {T} The decorated definition of the class of type T.
 */
function readonly<T extends {new (...args: any[]): {}}>(target: T) {
   // Note that we just get the results of the decorator from this function!
   const val = class extends target {
      constructor(...args: any[]) {
        super(args);
        Object.freeze(this);
      }
   };
   Object.freeze(val);
   Object.freeze(val.prototype);
   return val;
}

/**
 * Make the property readonly
 */
function readonlyProperty(name: string) {
  // We're returning the function that will then be applied to the target, much
  // like how the above example does so *directly*!
  return (target: {}) => {
     const val = target[name];
     if (val) {
       Object.freeze(name);
     }
  };
}
```

Example usage is here:
```ts
@readonly
class Foo {
}

@readonlyProperty("name")
class Bar {
  public name: string;
}
```

_Do I need to document what metadata each decorator adds to the reflector and document in the reference?_

Personally, I wouldn't expect that at this time. We'll need to make more in-depth documentation for these sections as we go, but for now, explaining how LoopBack uses them, what they're called and what they do is sufficient.

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+
+- @api
+
+  Register route for a controller by api decorator, more details and example refer to [Specifying Controller APIs](controllers.htm#Specifying Controller APIs)
+
+- @operation
+
+  Expose a Controller method as a REST API operation. Usage refer to [code](https://github.com/strongloop/loopback-next/blob/debe221ee7eb07b6d6962416b79592c8f13d1e91/packages/core/src/router/metadata.ts#L230-L234)
+
+  *Janny: Need example in doc?*
```
kjdelisle: Nope, it's documented in several areas. But I wouldn't use the link you have here, as it's tied to a specific commit. `/blob/debe221ee7eb07b6d6962416b79592c8f13d1e91/packages...` should be `/blob/master/packages...`

kjdelisle: Additionally, we're linking to a document called metadata. If there isn't anything that describes how to do this in the Controller section, then that's a problem to fix later.

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+
+- @param
+
+  Describe an input parameter of a Controller method. An example can be found in [Writing Controller methods](controller.htm#Writing-Controller-methods), more usage refer to [tsDoc](https://github.com/strongloop/loopback-next/blob/debe221ee7eb07b6d6962416b79592c8f13d1e91/packages/core/src/router/metadata.ts#L252-L272)
```
kjdelisle: Same issues as above.

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+
+  *Janny: Change link to tsDoc page when it's done*
+  
+## Dependency Injection
+
+@inject is a decorator to annotate method arguments for automatic injection by LoopBack IoC container. It can only be applied on a **method parameter** or a **property**. Basiclly it defines a reflector metadata for the target class.
```
kjdelisle: Several items:
- Don't use the word "basically"
- "It can only be applied on a method parameter or a property" doesn't explain that it doesn't mean you can't _inject_ objects/classes, or that those properties can't be complex objects themselves.
- "It defines a reflector metadata for the target class" tells users that they have to target a class (which is not true) and is too vague. "reflector metadata" doesn't tell us anything about how it works.

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+
+A quick example can be found in [test case](https://github.com/strongloop/loopback-next/blob/9769e85306a8a866d71f133b1d0c013dde0d43c2/packages/context/test/unit/inject.test.ts#L21-L28)
```
kjdelisle: Again, hash-based linking is bad. Switch it to use `/blob/master/...`

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+
+It's been documented as a LoopBack core concept in [Dependency Injection](Dependency-Injection.htm)
```
kjdelisle: `For more information, see the [Dependency Injection](Dependency-Injection.htm) section under [LoopBack Core Concepts]({I don't know where the link to that is})`

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+
+*Janny: we miss a list of all metadata keys we can inject*
```
kjdelisle: The _keys_ should be listed under the Dependency Injection section, or in a glossary later.

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Authentication Decorators
```
kjdelisle: `## Authentication Decorator` -- shouldn't be plural.

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+
+- @authenticate
+
+Mark a controller method as requiring authenticated user, it takes in a strategy name. Example refer to [test case](https://github.com/strongloop/loopback-next/blob/9769e85306a8a866d71f133b1d0c013dde0d43c2/packages/authentication/test/acceptance/basic-auth.ts#L104)
```
kjdelisle: fix link hash to be master.

kjdelisle: Better yet, don't link users to tests, but to the documentation for those plugins. If they don't have any, that's fine.

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+
+Define a model in a repository, see [define-models](http://loopback.io/doc/en/lb4/Repositories.html#define-models)
+
+- @property
+
+Define a property in a model, see [define-models](http://loopback.io/doc/en/lb4/Repositories.html#define-models)
+
+### Relation decorators
+
+*Since relation is still in design phase, we only have unit test but no integration example. Refer to [test case](https://github.com/strongloop/loopback-next/blob/9769e85306a8a866d71f133b1d0c013dde0d43c2/packages/repository/src/decorators/relation.ts) for usage*
```
kjdelisle: fix hash

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+
+- @belongsTo @hasOne @hasMany @embedsOne @embedsMany @referencesOne @referencesMany
+
+Register a specific relation
+
+### Repository decorators
+
+- @repository
+
+A Repository is a type of _Service_ that represents a collection of data within a DataSource, the decorator takes in a model class(or its name) and a datasource instance(or its name). For usage, refer to [test case](https://github.com/strongloop/loopback-next/blob/9769e85306a8a866d71f133b1d0c013dde0d43c2/packages/repository/test/unit/decorator/repository.ts#L90)
```
kjdelisle: fix hash

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Compose and customize decorators
+
+TBD 
+
+- Use chitchat as an example of compose decorators
```
kjdelisle: Example snippets are still sufficient.

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+- for customize decorators, I expect to create some validation or hook functions as example
```
kjdelisle: We don't need to show users how to make their own decorators, that's the TypeScript documentation's focus.

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Understand TypeScript Decorator
```
kjdelisle: We don't need this section, remove it.

kjdelisle (changes requested):

pages/en/lb4/Decorator.md
jannyHou: @kjdelisle I feel you may misunderstand the layout here(or git shows your comment in wrong line), `authentication` and `repository` are components, not in loopback core.

So how i organize the content here is:
- First introduce two decorators loopback core provides: 
  - decorator functions
  - injection
- Second introduce component decorators

See: 
```
Currently LoopBack core provides built-in decorator functions you can use directly by `@myDecorator`, e.g. Route decorators that introduces in section Route-Decorators, and implements a dependency injection system that can apply a decorator by `@inject('myDecorator')`. 

Some components also exports decorators for you:
[authentication](#authentication-decorators)
[repository](#repository-decorators)

```

pages/en/lb4/Decorator.md
jannyHou: UPDATE:
Currently the examples in concept [controllers](http://loopback.io/doc/en/lb4/Controllers.html#routing-to-controllers) does NOT contain `@operation`, so I will create another PR to add it to controllers section.

pages/en/lb4/Decorator.md
jannyHou: @kjdelisle 

> "It can only be applied on a method parameter or a property" doesn't explain that it doesn't mean you can't inject objects/classes, or that those properties can't be complex objects themselves.

Sorry I am not sure what do you mean here, according to our code you **cannot** inject a class or a method, see: https://github.com/strongloop/loopback-next/blob/master/packages/context/src/inject.ts#L99


> tells users that they have to target a class (which is not true)

From my understanding the `target` has to be a class or to be more accurate,  an instance of class, you can check the [function]( https://github.com/strongloop/loopback-next/blob/master/packages/context/src/inject.ts#L99), and could you give me an example of `@inject` of which the `target` is not an instance?

pages/en/lb4/Decorator.md
jannyHou: @kjdelisle it is a requirement I copied from original wiki page, not sure who put it there but seems someone wants it: https://github.com/strongloop/loopback-next/wiki/Using-decorators

pages/en/lb4/Decorator.md
kjdelisle: What I'm saying is that the current phrasing could be misread as saying that you can't resolve class instances with `@inject`, which you can.
The `@inject` decorator itself must be applied to references, as you say, but this isn't what you're saying in the above text.

kjdelisle: Also, `method arguments` implies a parameter passed into a function/constructor, rather than direct assignment of the target property to the injected value/result.

crandmck (commented):
I'll leave it for you and Kevin to hash out the technical aspects.  Once you land this PR, I will review and edit the article for English usage, grammar, and spelling.

jannyHou:
Thank you @kjdelisle appreciate your detailed comments :) Feedback applied, PTAL.

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,196 @@
...
+
+A decorator allows you annotate or modify your class declarations and members with metadata.
+
+## Introduction
+
+*If you're new to Decorators in TypeScript, see [here](https://www.typescriptlang.org/docs/handbook/decorators.html) for more info.*
+
+Decorators give LoopBack the flexibility to modify your plain TypeScript classes and properties in a way that allows the framework to better understand how to make use of them, without you needing to inherit base classes or add functions that tie into an API.
+
+Currently LoopBack core provides built-in decorator functions you can use directly by `@myDecorator`, e.g. Route decorators that introduces in section Route-Decorators, and implements a dependency injection system that can apply a decorator by `@inject('myDecorator')`.
```
kjdelisle: `Currently, ...` (needs a comma)

kjdelisle: Line length should be under 80. It's not enforced, but it makes reviewing docs much easier.

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,196 @@
...
+*If you're new to Decorators in TypeScript, see [here](https://www.typescriptlang.org/docs/handbook/decorators.html) for more info.*
+
+Decorators give LoopBack the flexibility to modify your plain TypeScript classes and properties in a way that allows the framework to better understand how to make use of them, without you needing to inherit base classes or add functions that tie into an API.
+
+Currently LoopBack core provides built-in decorator functions you can use directly by `@myDecorator`, e.g. Route decorators that introduces in section Route-Decorators, and implements a dependency injection system that can apply a decorator by `@inject('myDecorator')`. 
+
+As a default, LoopBack comes with some pre-defined component decorators for your use:
+- [authentication](#authentication-decorators)
+- [repository](#repository-decorators)
+
```
kjdelisle: Still missing the link to context decorators.

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,196 @@
...
+@api(widgetSpec)
+class WidgetController {
+  @inject('config.widget')
+  protected config: JSON; // This will be resolved at runtime!
+
+  // etc...
+}
+```
+
+It's been documented as a LoopBack core concept in [Dependency Injection](Dependency-Injection.htm)
```
kjdelisle: Remove this; you wrote its replacement one line below it!

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,196 @@
...
+For more information, see the [Dependency Injection](Dependency-Injection.htm) section under [LoopBack Core Concepts](Concepts.htm)
+
+
+## Authentication Decorator
+
+- @authenticate
+
+  Mark a controller method as requiring authenticated user, it takes in a strategy name. An example is using BasicStrategy to authenticate user in function `whoAmI`:
```
kjdelisle: "It takes in a strategy name" should be reworded to be explicit; mention that the decorator takes a strategy name as a parameter, and make sure there is an explanation for what that is (or a link to somewhere else that does explain it)

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,196 @@
...
+
+  Define a model in a repository, see [define-models](http://loopback.io/doc/en/lb4/Repositories.html#define-models)
+
+- @property
+
+  Define a property in a model, see [define-models](http://loopback.io/doc/en/lb4/Repositories.html#define-models)
+
+### Relation decorators
+
+  *Since relation is still in design phase, examples TBD.*
```
kjdelisle: Even the placeholders need formatting.
"This feature has not yet been released in alpha form. Documentation will be added here as this feature progresses."

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,196 @@
...
+
+- @belongsTo @hasOne @hasMany @embedsOne @embedsMany @referencesOne @referencesMany
+
+  Register a specific relation
+
+### Repository decorators
+
+- @repository
+
+  A Repository is a type of _Service_ that represents a collection of data within a DataSource, the decorator either injects an existing repository or creates a repository from a model and a datasource.
```
kjdelisle: A repository represents a collection of data *and* the means to access that data.

kjdelisle: In general, I'd rephrase this.

kjdelisle (changes requested):
Still a bit more to go, but looking good. :)

pages/en/lb4/Decorator.md
jannyHou: let me confirm...when you say context decorators you mean `inject` only right? I didn't find any other decorators in context....

pages/en/lb4/Decorator.md
jannyHou: @kjdelisle sorry I am still confused. This is the error msg from code:
> @inject can only be used on properties or method parameters.

If "It can only be applied on a method parameter or a property" is not saying the same thing, I really don't understand but I can change it to exactly "@inject can only be used on properties or method parameters.", is that ok for you?

pages/en/lb4/Decorator.md
jannyHou: And I also explained the layout in https://github.com/strongloop/loopback.io/pull/463#discussion_r139484961

pages/en/lb4/Decorator.md
jannyHou: yep I think the snippet below explains it

pages/en/lb4/Decorator.md
jannyHou: ok, actually I copied this line from old wiki....

jannyHou:
@kjdelisle only one comment need to clarify: https://github.com/strongloop/loopback.io/pull/463#discussion_r139803903, thanks!

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,209 @@
...
+*If you're new to Decorators in TypeScript, see [here](https://www.typescriptlang.org/docs/handbook/decorators.html) for more info.*
+
+Decorators give LoopBack the flexibility to modify your plain TypeScript classes 
+and properties in a way that allows the framework to better understand how to 
+make use of them, without you needing to inherit base classes or add functions 
+that tie into an API.
+
+Currently, LoopBack core provides built-in decorator functions you can use directly
+by `@myDecorator`, e.g. Route decorators that introduces in section Route-Decorators, 
+and implements a dependency injection system that can apply a decorator by `@inject('myDecorator')`.
```
bajtos: I don't think this is a valid usage: `@inject('myDecorator')`

`@inject` is a decorator for annotating dependencies to be injected, the parameter is the binding key that should be used to resolve the dependency. For example, `@inject('sequence.actions.findRoute')`.

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,209 @@
...
+      ) {}
+    }
+  ```
+  
+  More example can be found in [Writing Controller methods](controller.htm#Writing-Controller-methods)
+  
+## Dependency Injection
+
+@inject is a decorator to annotate method arguments for automatic injection by 
+LoopBack IoC container. @inject can only be used on properties or method parameters.
```
bajtos: I think you should to wrap `@inject` in back-ticks, so that it renders as code. I think GitHub's syntax highlighter thinks `@inject` is a GitHub handle now.

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,209 @@
...
+
+The `@inject` decorator allows you inject dependencies bound to the application context, or
```
bajtos: `...to the application context, to the per-request context, or to any implementation...`

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,209 @@
...
+Now that we've bound the 'config.widget' key to our configuration object, we can
+inject it in our WidgetController:
+
+```ts
+// widget-controller.ts
+import {widgetSpec} from '../apispec';
+@api(widgetSpec)
+class WidgetController {
+  @inject('config.widget')
+  protected config: JSON; // This will be resolved at runtime!
```
bajtos: Property injection should be used for things that have reasonable default that can be used by the class when the caller did not provide/inject any value. Required dependencies should be injected via constructor arguments. 

This is important to make classes easy to use directly from code, with no IoC framework resolving the dependencies. A typical use case is a unit-test:

```ts
class controller = new WidgetController();
const result = await controller.doSomething();
// oops, config was not initialized!
```

I am proposing to rework this example to use constructor-argument injection.

```ts
@api(widgetSpec)
class WidgetController {
  constructor(
    @inject('config.widget') protected config: JSON
  ) {}

  // etc...
}
```

pages/en/lb4/Decorator.md
```diff
@@ -0,0 +1,209 @@
...
+    host: 'localhost',
+    port: 3306,
+    database: 'test',
+    password: 'pass',
+    user: 'root',
+  });
+
+  export class TodoController {
+    @repository(Todo, datasource)
+    repository: EntityCrudRepository<Todo, number>;
```
bajtos: While this is a valid usage, we are advocating for a slightly different usage in Thinking in LoopBack - see [Define Product model, repository, and data source](http://loopback.io/doc/en/lb4/Thinking-in-LoopBack.html#define-product-model-repository-and-data-source) and [Decouple Controller from Repository](http://loopback.io/doc/en/lb4/Thinking-in-LoopBack.html#decouple-controller-from-repository).

I'd like us to follow the approach described in Thinking in LoopBack in the examples shown on this page too.

bajtos (dismissed):
Great content!

crandmck (changes requested):
Please also add this page to the sidebar navigation: https://github.com/strongloop/loopback.io/blob/gh-pages/_data/sidebars/lb4_sidebar.yml#L46

It should be obvious how to add it, but LMK if you have any questions.

pages/en/lb4/Decorator.md
jannyHou: True, this part is from my old (wrong) understanding and I guess it also causes Kevin confused. Will remove it.

jannyHou:
@bajtos @crandmck Thanks! feedbacks applied and add new page to side bar.

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+lang: en
+title: 'Decorators'
+keywords: LoopBack 4.0, LoopBack-Next
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Decorators.html
+summary:
+---
+
+A decorator allows you annotate or modify your class declarations and members with metadata.
```
dhmlau: ... allows you `to` annotate or.....

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Introduction
+
+*If you're new to Decorators in TypeScript, see [here](https://www.typescriptlang.org/docs/handbook/decorators.html) for more info.*
+
+Decorators give LoopBack the flexibility to modify your plain TypeScript classes 
+and properties in a way that allows the framework to better understand how to 
+make use of them, without you needing to inherit base classes or add functions
```
dhmlau: suggested change:
from:
> without you needing to inherit 

to:
 
> without the need to inherit...

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+make use of them, without you needing to inherit base classes or add functions 
+that tie into an API.
+
+Currently, LoopBack core provides built-in decorator functions you can use directly
+by `@myDecorator`:
+
+- [route](#route-decorators)
+- [inject](#dependency-injection)
+
+As a default, LoopBack also comes with some pre-defined component decorators for your use:
```
dhmlau: I think without `for your use` is fine. i.e.
> As a default, LoopBack also comes with some pre-defined component decorators

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+
+- [authentication](#authentication-decorators)
+- [repository](#repository-decorators)
+
+## Route Decorators
+- `@get`, `@post`, `@put`, `@patch`, `@del`
+
+  Register route for a controller by rest decorator, for their usage, refer to
```
dhmlau: `REST` instead of `rest`?

jannyHou:
@dhmlau Thanks! Feedback applied.

crandmck (approved):

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+
+*If you're new to Decorators in TypeScript, see [here](https://www.typescriptlang.org/docs/handbook/decorators.html) for more info.*
+
+Decorators give LoopBack the flexibility to modify your plain TypeScript classes 
+and properties in a way that allows the framework to better understand how to 
+make use of them, without the need to inherit base classes or add functions 
+that tie into an API.
+
+Currently, LoopBack core provides built-in decorator functions you can use directly
+by `@myDecorator`:
```
kjdelisle: Is this supposed to be here?
```
Currently, LoopBack core provides built-in decorator functions for use
in your application and controllers:
 - [@route](#route-decorators)
 - [@inject](#inject-decorators)
 // etc
```

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+
+- [route](#route-decorators)
+- [inject](#dependency-injection)
+
+As a default, LoopBack also comes with some pre-defined component decorators:
+
+- [authentication](#authentication-decorators)
+- [repository](#repository-decorators)
```
kjdelisle:
```
- [@authentication](#authentication-decorators)
- [@repository](#repository-decorators)
```

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Route Decorators
+- `@get`, `@post`, `@put`, `@patch`, `@del`
```
kjdelisle: Needs a brief overview before this.
```
## Route Decorators
`@loopback/rest` provides some default routing decorators that map
to the five most commonly used REST verbs: `@get`, `@post`, `@put`,
`@patch`, and `@del`.
```

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+
+  Register route for a controller by REST decorator, for their usage, refer to
```
kjdelisle:
```
You can use these decorators to setup routing to individual functions within your controllers.
For usage examples, see [Routing to Controllers](controllers.htm#Routing-to-Controllers)
```

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+- [authentication](#authentication-decorators)
+- [repository](#repository-decorators)
+
+## Route Decorators
+- `@get`, `@post`, `@put`, `@patch`, `@del`
+
+  Register route for a controller by REST decorator, for their usage, refer to 
+  [Routing to Controllers](controllers.htm#Routing-to-Controllers)
+
+- `@api`
```
kjdelisle:
```
## OpenAPI Spec Decorator
You can define the API behaviour for some or all of the functions on a
controller simultaneously by using the `@api` decorator. For usage examples,
see [Specifying Controller APIs](controllers.htm#Specifying-Controller-APIs).
```

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+
+  Register route for a controller by api decorator, more details and example refer 
+  to [Specifying Controller APIs](controllers.htm#Specifying-Controller-APIs)
+
+- `@operation`
```
kjdelisle:
```
## Operation Decorator
The more open-ended version of the pre-defined REST verb decorators,
this decorator allows you to define less commonly used verbs yourself,
such as `OPTIONS` or `HEAD`. For usage examples, see
[TBD in Controllers](controllers.htm#)
```

kjdelisle: This is also meant to replace the text just below it.

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+
+  Expose a Controller method as a REST API operation. Usage refer to [TBD in Controllers](controllers.htm#)
+
+- `@param`
+
+  Describe an input parameter of a Controller method.
```
kjdelisle: `Describe input parameters for your controller functions.`

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+    class MyController {
+      @get('/')
+      list(
+        @param(offsetSpec) offset?: number,
+        @param(pageSizeSpec) pageSize?: number,
+      ) {}
+    }
+  ```
+  
+  More example can be found in [Writing Controller methods](controller.htm#Writing-Controller-methods)
```
kjdelisle: `More examples can be found...`
Missing the "s" on examples.

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+  }
+  ```
+
+## Repository Decorators
+
+### Model decorators
+
+- `@model`
+
+  Define a model in a repository, see [define-models](Repositories.html#define-models)
```
kjdelisle:
```
Define the decorated class as a Model for use with the Legacy Juggler.
For usage examples, see [Define Models](Repositories.html#define-models)
```

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+
+- `@property`
+
+  Define a property in a model, see [define-models](Repositories.html#define-models)
```
kjdelisle:
```
Define the metadata for a Model property.
For usage examples, see [Define Models](Repositories.html#define-models)
```

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+    @authenticate('BasicStrategy')
+    async whoAmI() : Promise<string> {
+      return this.user.id;
+    }
+  }
+  ```
+
+## Repository Decorators
+
+### Model decorators
```
kjdelisle: `### Model Decorators`

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+
+- `@model`
+
+  Define a model in a repository, see [define-models](Repositories.html#define-models)
+
+- `@property`
+
+  Define a property in a model, see [define-models](Repositories.html#define-models)
+
+### Relation decorators
```
kjdelisle: `### Relation Decorators`

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+
+- `@relation`
+
+  Register a general relation
+
+- `@belongsTo` `@hasOne` `@hasMany` `@embedsOne` `@embedsMany` `@referencesOne` `@referencesMany`
+
+  Register a specific relation
+
+### Repository decorator
```
kjdelisle: `### Repository Decorator`

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+
+- `@repository`
+
+  A repository represents a collection of data and the means to access that data,
```
kjdelisle: Comma should be a period.

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+- `@belongsTo` `@hasOne` `@hasMany` `@embedsOne` `@embedsMany` `@referencesOne` `@referencesMany`
+
+  Register a specific relation
+
+### Repository decorator
+
+- `@repository`
+
+  A repository represents a collection of data and the means to access that data, 
+  the decorator either injects an existing repository or creates a repository
```
kjdelisle:
```
The `@repository` decorator either injects an existing repository instance
or instantiates one for you, using a given Model and a DataSource.

For detailed usage examples, see
[Repository#controller-configuration](Repositories.html#controller-configuration)
```

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+  A repository represents a collection of data and the means to access that data, 
+  the decorator either injects an existing repository or creates a repository 
+  from a model and a datasource.
+
+  The injection example can be found in [Repository#controller-configuration](Repositories.html#controller-configuration)
+
+  ```ts
+  // my-controller.ts
+  // Know more about how to create model 'Todo' and datasource 'datasource', 
+  // please see example in [Thinking in LoopBack](Thinking-in-LoopBack.htm#define-product-model-repository-and-data-source)
```
kjdelisle: Remove these comments.

kjdelisle: If the link is useful, it should be outside the code snippet, and properly formatted for use.

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+
+  ```ts
+  // with `datasource` and `Todo` already defined.
+  app.bind('datasources.ds').to(datasource);
+  app.bind('repositories.todo').to(Todo);
+
+  export class TodoController {
+    @repository('todo', 'ds')
+    repository: EntityCrudRepository<Todo, number>;
+    ... ...
```
kjdelisle: Replace with:
```
 // etc
```

This may seem trivial, but it lets people paste this into an editor with less friction.

They're still going to see errors due to things not being defined, such as `app`, but if they're following along with our examples, they might already have the rest of the groundwork laid out, which would mean the snippet could be useful.

kjdelisle (changes requested):
Some more technical and grammar cleanup is needed, IMO, but it's definitely shaping up nicely, thanks @jannyHou !

pages/en/lb4/Decorators.md
kjdelisle: Alternatively, you can bullet-list them like the previous example
```
- `@get`
- `@post`
- `@put`
- `@patch`
- `@del`
```

pages/en/lb4/Decorators.md
kjdelisle: The primary reason I'd like to see the decorators replaced with proper titles is that it makes it easier to reference these sections from elsewhere (either in the doc or external to it).

The secondary reason is that it's easier to find when you're scrolling through it. Code-blocked text isn't large with heavy-weighted font like subtitle headers.

pages/en/lb4/Decorators.md
kjdelisle: Don't forget the header!
```
## Param Decorator
```

pages/en/lb4/Decorators.md
jannyHou: @kjdelisle actually 'route' 'inject' 'authentication' 'repository' are not decorator names, but decorator categories, for example route category has N decorators, authentication category has one decorator.

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,207 @@
...
+- `@api`
+
+  Register route for a controller by api decorator, more details and example refer 
+  to [Specifying Controller APIs](controllers.htm#Specifying-Controller-APIs)
+
+- `@operation`
+
+  Expose a Controller method as a REST API operation. Usage refer to [TBD in Controllers](controllers.htm#)
+
+- `@param`
```
virkt25: `@param` has another set of operations that make it even easier to use that I think might be worthwhile documenting. 

It follows this pattern to make it quicker to define params: `@param.${in}.${type}(${name})`
in - one of: query, header, path, formData, body
type - one of: string, number, boolean, integer
name - name of the parameter

So an example would be `@param.query.number('offset')` or `@param.query.number('pageSize')`

Looks great otherwise! :)

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,316 @@
...
+If you are not familiar with the concept Route or Controller, please see [LoopBack Route](routes.htm)
+and [LoopBack Controller](controllers.htm) to learn more about them.
+
+By calling a route decorator, you provide OpenAPI specification to describe the 
+endpoint which the decorated method maps to. You can choose different decorators 
+accordingly or do a composition of them:
+
+### API Decorator
+
+  Syntax: `@api`
```
raymondfeng: Should it be `@api(swaggerSpecObject)`?

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,316 @@
...
+      }
+    }
+    app.controller(MyController);
+  ```
+
+  A more detailed explanation can be found in [Specifying Controller APIs](controllers.htm#specifying-controller-apis)
+
+### Operation Decorator
+
+  Syntax: `@operation`
```
raymondfeng: `@operation(swaggerSpecOperationObject)`?

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,316 @@
...
+  class MyController {
+    @operation('HEAD', '/checkExist', spec)
+    checkExist(name) {
+    }
+  }
+  ```
+  
+### Commonly-used Operations Decorator
+  
+  Syntax: `@get`, `@post`, `@put`, `@patch`, `@del`
```
raymondfeng: `@get(path, swaggerSpecOperationObject)`?

pages/en/lb4/Decorators.md
raymondfeng: Maybe a link to the decorator apidocs?

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,316 @@
...
+  an input parameter of a Controller method.
+
+  For example:
+
+  ```ts
+    class MyController {
+      @get('/')
+      @param(offsetSpec)
+      @param(pageSizeSpec)
+      list(offset?: number, pageSize?: number) {}
```
raymondfeng, line 151: Add a note that method level `@param` is positional and it's exclusive with parameter level `@param`

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,316 @@
...
+Now that we've bound the 'config.widget' key to our configuration object, we can
+inject it in our WidgetController:
+
+```ts
+// widget-controller.ts
+import {widgetSpec} from '../apispec';
+@api(widgetSpec)
+class WidgetController {
+  constructor(
+    @inject('config.widget') protected widget: JSON
```
raymondfeng: The type `JSON` is suspicious.

raymondfeng: Can we also add an example for `property injection` to apply `@inject` to a property?

pages/en/lb4/Decorators.md
crandmck: IIUC, these should show up in http://apidocs.loopback.io/@loopback%2fauthentication/ but I don't see them.  Do we require an npm publish to see them or is something else going on?

raymondfeng (changes requested):
Great work. Please address my comments.

pages/en/lb4/Decorators.md
kjdelisle: Can you expand on what that means? Personally, I _think_ I know what you mean, but I wouldn't act on the feedback if I didn't know what it meant.

pages/en/lb4/Decorators.md
kjdelisle: >The type JSON is suspicious.

I wouldn't saturate the example with detail that might obscure the essential meaning in lead-chasing.
`@inject('config.widget') protected widgetConfig: any` should do the trick.

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,316 @@
...
+}
+```
+
+For more information, see the [Dependency Injection](Dependency-Injection.htm) section under [LoopBack Core Concepts](Concepts.htm)
+
+## Authentication Decorator
+
+  Syntax: `@authenticate`
+
+  Mark a controller method as requiring authenticated user, takes a strategy name
```
kjdelisle:
```
Marks a controller method as needing an authenticated user.
This decorator requires a strategy name as a parameter.

Here's an example using 'BasicStrategy':
// etc
```

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,316 @@
...
+    async whoAmI() : Promise<string> {
+      return this.user.id;
+    }
+  }
+  ```
+
+## Repository Decorators
+
+### Model Decorators
+
```
kjdelisle:
```
Model decorators provide the schematic metadata for the legacy juggler to validate given inputs
against expected types, required properties and more.
```

kjdelisle (changes requested):
I'll take another look later, need to submit what I've got for now and jump onto a different PR.

pages/en/lb4/Decorators.md
jannyHou: @raymondfeng nice to know we have the apidocs already,will link the function there and add parameters. 
While as Rand says, ~seems most Route decorators are currently missing~ in http://apidocs.loopback.io/@loopback%2fcore/#1666
Never mind I find them.

pages/en/lb4/Decorators.md
jannyHou: @raymondfeng @kjdelisle Let me confirm I understand it correctly... "positional" means the sequence of composed decorator must be
```
@operationDecorator
@param
```

> exclusive with parameter level @param

Means "Mixed usage of @param at method/parameter level is not allowed and results in error"

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,316 @@
...
+- [Authentication Decorator](#authentication-decorators)
+- [Repository Decorators](#repository-decorators)
+
+## Route Decorators
+
+Route decorators are used to expose controller methods as REST API operations.
+If you are not familiar with the concept Route or Controller, please see [LoopBack Route](routes.htm)
+and [LoopBack Controller](controllers.htm) to learn more about them.
+
+By calling a route decorator, you provide OpenAPI specification to describe the
```
b-admike, line 35: Can we re-word it to `OpenAPI PathsObject specification`, so it's clear that we don't mean the full spec?

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,316 @@
...
+    }
+  };
+  class MyController {
+    @operation('HEAD', '/checkExist', spec)
+    checkExist(name) {
+    }
+  }
+  ```
+  
+### Commonly-used Operations Decorator
```
b-admike: `Decorator` -> `Decorators`

pages/en/lb4/Decorators.md
jannyHou: @kjdelisle @raymondfeng Sorry I am confused, should the type of `widget` be `any` instead of `JSON`? and could you elaborate more about why `JSON` type is suspicious?

pages/en/lb4/Decorators.md
jannyHou: @b-admike I choose the more general description because for some decorator the input is not a `PathsObject`:
- @api: it takes in `basePath` as a string
- @param: it takes in a `paramSpec`, even though under the hood the decorator converts it to a part of `PathsObject`.

pages/en/lb4/Decorators.md
kjdelisle: @jannyHou It doesn't actually exist (in the sense it's used here, anyway). When I wrote that example, I "paraphrased" the type of the object so that it would be immediately clear what that item was supposed to be, but it's technically an invalid type and wouldn't _actually_ work.

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,379 @@
...
+  }
+  ```
+
+## Repository Decorators
+
+  As a Domain-driven design concept, 
+  repository means a mediate between your domain object and data mapping layers
+  using a collection-like interface for accessing domain objects. 
+  In framework LoopBack, a domain object is usually a TypeScript/JavaScript Class instance, 
+  and a typical example of data mappting layer module could be a database's node.js driver.
```
kjdelisle: `...a typical example of data mappting...` -> `...a typical example of a data mapping...`

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,379 @@
...
+  
+  LoopBack repository encapsulates your TypeScript/JavaScript Class instance, 
+  and its methods that communicate with your database.
+  It is an interface to implement data persistency.
```
kjdelisle: `persistency` -> `persistence`

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,379 @@
...
+    async whoAmI() : Promise<string> {
+      return this.user.id;
+    }
+  }
+  ```
+
+## Repository Decorators
+
+  As a Domain-driven design concept, 
+  repository means a mediate between your domain object and data mapping layers
```
kjdelisle: `the repository is a layer between your domain object and data mapping layers`

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,379 @@
...
+  and a typical example of data mappting layer module could be a database's node.js driver.
+  
+  LoopBack repository encapsulates your TypeScript/JavaScript Class instance, 
+  and its methods that communicate with your database.
+  It is an interface to implement data persistency. 
+
+  Repository decorators are used for defining models(domain objects) for use with your chosen datasources, 
+  and the navigation strategies among models.
+
+  If you are not familiar with repository related concepts like Model, Entity, Datasource,
```
kjdelisle: `Model, Entity, and Datasource`

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,379 @@
...
+  Model decorator is a Class decorator.
+  In LoopBack 4, we inherit the model definition format from LoopBack 3,
+  you can find it in [Model definition JSON file](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html).
+  For usage examples, see [Define Models](Repositories.html#define-models)
+
+  *Please note we will elaborate more about model and model definition in #Model.htm,*
+  *and replace the link above with LB4 link*
+
+  By using a model decorator, you can define a model as your repository's metadata,
+  and resolve your repository with Legacy Juggler that defautly comes with a set of CRUD operations bound to your model.
```
kjdelisle: - Line is too long
- `defaultly` isn't a word
Suggestion: `...which will use the legacy juggler to create a repository instance complete with CRUD methods for accessing a specific model's data.`

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,379 @@
...
+  A use case can be found in section [Repository decorator](#repository-decorator)
+
+  Or you also have the flexibility to define a model without being resolved with Legacy Juggler.
```
kjdelisle: `You can also define your own repositories without using legacy juggler, and use an ORM/ODM of your choice.`
(You can probably drop the 2nd line as well; feel free to copy-paste this comment. :P )

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,379 @@
...
+  // Missing example here
+  // Will be provided in Model.htm
+  // refer to [example code](https://github.com/strongloop/loopback-next-example/blob/master/services/account-without-juggler/repositories/account/models/Account.ts)
+  ```
+
+#### Property Decorator
+
+  Syntax: `@property(definition: PropertyDefinition)`
+
+  It is a Class property decorator, defines the metadata for a Model property.
```
kjdelisle: Might want to rephrase this to better establish context.
"The property decorator defines metadata for a property on a Model definition." or something similar.

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,379 @@
...
+  The format of property definitions can be found in [Property definitions](https://loopback.io/doc/en/lb2/Model-definition-JSON-file.html#properties) 
+
+  For usage examples, see [Define Models](Repositories.html#define-models)
+
+### Relation Decorators
+
+  Relation represents how you navigate from one model to other models.
```
kjdelisle: The relation decorator defines the nature of a relationship between two models.

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,379 @@
...
+
+  This decorator either injects an existing repository or creates a repository 
+  from a model and a datasource.
+
+  The injection example can be found in [Repository#controller-configuration](Repositories.html#controller-configuration)
+
+  To create a repository in a controller, you can define your model and datasource 
+  first, then import them in your controller file:
+
+  *Know more about how to create model and datasource, please see example in [Thinking in LoopBack](Thinking-in-LoopBack.htm#define-product-model-repository-and-data-source)*
```
kjdelisle: `To learn more about creating models and datasources, please see the example in...`

kjdelisle (changes requested):
So close! Just a few awkward phrases and typos.

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,379 @@
...
+    }
+  }
+  ```
+
+## Repository Decorators
+
+  As a Domain-driven design concept, 
+  repository means a mediate between your domain object and data mapping layers
+  using a collection-like interface for accessing domain objects. 
+  In framework LoopBack, a domain object is usually a TypeScript/JavaScript Class instance,
```
kjdelisle: `In LoopBack, a domain object is...`

kjdelisle (changes requested):
Missed one, adding it to the list.

pages/en/lb4/Decorators.md
jannyHou: wait...this is in section#model-decorator, so it's talking about how to define `Model` not `Repository` :)

pages/en/lb4/Decorators.md
jannyHou: actually I prefer to explain the relation as a "navigation" concept, otherwise people will be confused what does "relationship" mean, if they are not familiar with ORM relation.

I think it's a good idea to keep them both.

pages/en/lb4/Decorators.md
jannyHou: The reason why I explain "with-juggler" and "without-juggler" scenarios in model decorator is:

By using decorator `@repository`, you already choose legacy juggler.

jannyHou:
@kjdelisle @raymondfeng @b-admike @virkt25 Feedbacks applied from commit `Improve syntax` to the last one, PTAL, thanks!

pages/en/lb4/Decorators.md
kjdelisle: I disagree. "Navigate" can mean all sorts of confusing things.
"What do they mean by navigate? I need to navigate between my Models? Does that mean one of them calls the other?"

It's not really the correct word for what's happening here. It's not a traversal of anything, it's the established relationship between two data types. Whether or not that means one owns the other, has many of them, must be created when the other is created (and so on) doesn't have to be defined here.

One might "navigate" an existing relationship, but this is _how you make that relationship exist_.

Part of the problem is that you can't really be more specific until we have code examples that represent what's going on (whenever we start getting close to Juggler GA).

pages/en/lb4/Decorators.md
kjdelisle: @raymondfeng We really need an answer here.

pages/en/lb4/Decorators.md
jannyHou: I think in LB3, relation means you can access/modify/delete other models' data by a model's apis, which fits `navigation` more than `relationship`, but if that's not what we are going to build in LB-next then I am good with the more general description "relationship"

pages/en/lb4/Decorators.md
raymondfeng: @kjdelisle See https://github.com/strongloop/loopback-next/blob/master/packages/rest/test/unit/router/metadata/param.test.ts#L40-L119

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,396 @@
...
+### Commonly-used Operation Decorators
+  
+  Syntax: [`@get(path: string, spec?: OperationObject)`](http://apidocs.loopback.io/@loopback%2fcore/#798)
+  
+  Same Syntax for decorators [`@post`](http://apidocs.loopback.io/@loopback%2fcore/#802)
+  , [`@put`](http://apidocs.loopback.io/@loopback%2fcore/#806)
+  , [`@patch`](http://apidocs.loopback.io/@loopback%2fcore/#810)
+  , [`@del`](http://apidocs.loopback.io/@loopback%2fcore/#814)
+
+  You can call these suger operation decorators as a shortcut of `@operation`, for example:
```
kjdelisle: `suger` -> `sugar`

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,396 @@
...
+      ) {}
+    }
+  ```
+
+  By the first example you apply multiple decorators to a single declaration, 
+  their sequence is important, a `@param` decorator must be applied after 
+  an operation decorator. Know more about TypeScript decorator composition, 
+  refer to [TypeScript Decorator Documentation](https://www.typescriptlang.org/docs/handbook/decorators.html)
+
+  Please note method level `@oaram` and parameter level `@param` are mutually exclusive,
```
kjdelisle: The first `@param` here is misspelled as `@oaram`.

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,396 @@
...
+    class MyController {
+      @get('/')
+      list(
+        @param(offsetSpec) offset?: number,
+        @param(pageSizeSpec) pageSize?: number,
+      ) {}
+    }
+  ```
+
+  By the first example you apply multiple decorators to a single declaration,
```
kjdelisle: `In the first example, we apply multiple decorators to a single declaration. The order in which the decorators are called is important because...`

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,396 @@
...
+  - type: one of the following values:  `string`, `number`, `boolean`, `integer`
+  - name: a `string`, name of the parameter
+
+  So an example would be `@param.query.number('offset')`. 
+  You can find the specific usage in [Writing Controller methods](controller.htm#writing-controller-methods)
+  
+## Dependency Injection
+
+`@inject` is a decorator to annotate method arguments for automatic injection by 
+LoopBack IoC container.
```
kjdelisle: `LoopBack's`

jannyHou:
feedback from Raymond:
https://github.com/strongloop/loopback.io/pull/463/commits/48dbbfe59ae0024f92932c8287c926c9239ab4dd
https://github.com/strongloop/loopback.io/pull/463/commits/6f17332713418bd8936385eab543bce2b2bc7411

from Kevin:
https://github.com/strongloop/loopback.io/pull/463/commits/3aa28392d556119d0642e3e18e04ff44e083c190
https://github.com/strongloop/loopback.io/pull/463/commits/d2f296bebc8f608f5b74878224aed0880451a9e7
https://github.com/strongloop/loopback.io/pull/463/commits/7389fcf6a69f99ba81b82211cefc94f0ef74db47
https://github.com/strongloop/loopback.io/pull/463/commits/62bc19be787eba842e5566e5137462e77f3499c9


from Taranveer:
https://github.com/strongloop/loopback.io/pull/463/commits/345cf2103621b93eff75b9360669e4d55f06b608

from Biniam:
https://github.com/strongloop/loopback.io/pull/463/commits/3aa28392d556119d0642e3e18e04ff44e083c190

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,396 @@
...
+## Dependency Injection
+
+`@inject` is a decorator to annotate method arguments for automatic injection by 
+LoopBack IoC container. 
+
+The injected values are applied to a constructed instance, so it can only be used on 
+non-static properties or constructor parameters of a Class.
+
+The `@inject` decorator allows you inject dependencies bound to the application context, 
+to the per-request contextor, to any implementation of the [Context](#context) object.
```
kjdelisle: `The @inject decorator allows you to inject dependencies bound to any Context, such as an Application instance or a request context instance.`

I can't add the backticks in the example for `@inject` so keep that in mind. :P

raymondfeng (approved):

kjdelisle (changes requested):

b-admike (approved):
:ship::it:

pages/en/lb4/Decorators.md
```diff
@@ -0,0 +1,397 @@
...
+        @param(offsetSpec) offset?: number,
+        @param(pageSizeSpec) pageSize?: number,
+      ) {}
+    }
+  ```
+
+  In the first example, we apply multiple decorators to a single declaration. 
+  The order in which the decorators are called is important because a `@param` 
+  decorator must be applied after an operation decorator. 
+  Know more about TypeScript decorator composition,
```
kjdelisle: `To learn more about...`

kjdelisle (changes requested):
Last change, then LGTM

kjdelisle (approved):

PR #464, opened 2017-09-18
fix(Schemas): Decorators and legacy juggler

kjdelisle:
Document the current state of Schema decorators with respect to legacy juggler.

connected to https://github.com/strongloop/loopback-next/issues/554

### DIFF
--- a/pages/en/lb4/Schemas.md
+++ b/pages/en/lb4/Schemas.md
@@ -9,9 +9,115 @@ summary:
 ---
 {% include content/tbd.html %}
 
-> Stub page per [#554](https://github.com/strongloop/loopback-next/issues/554).
+## Overview
+A `schema` represents the definition of a model in LoopBack, with respect to
+the [datasource juggler](https://github.com/strongloop/loopback-datasource-juggler).
+Currently, we provide the `@loopback/repository` module, which provides special
+decorators for adding metadata to your TypeScript/JavaScript
+classes in order to use them with the legacy implementation of the Juggler.
 
-- What is a Schema?
-- How to:
-  - Define a schema for a route response
-  - Define a model with a schema
+## Using Legacy Juggler
+Defining a schema for use with the legacy juggler involves decorating your
+classes with the `@model` and `@property` decorators.
+
+```ts
+import { model, property } from '@loopback/repository';
+
+@model()
+export class Product extends Entity {
+  @property({
+    type: 'number',
+    id: true,
+    description: 'The unique identifier for a product',
+  })
+  id: number;
+
+  @property({type: 'string'})
+  name: string;
+
+  @property({type: 'string'})
+  slug: string;
+
+  constructor(data?: Partial<Product>) {
+    super(data);
+  }
+}
+```
+
+### Model Decorator
+The model decorator can be used without any additional parameters, or can take
+in a ModelDefinitionSyntax object which follows the general format provided in
+LoopBack 3:
+```ts
+{
+@model({
+  name: "Category",
+  properties: {
+    // define properties here.
+  },
+  settings: {
+    // etc...
+  }
+})
+class Category extends Entity {
+  // etc...
+}
+}
+```
+
+### Property Decorators
+The property decorator takes in the same arguments used in LoopBack 3 for
+individual property entries:
+```ts
+@model()
+class Product extends Entity {
+  @property({
+    name: "name",
+    description: "The product's common name.",
+    type: "string",
+  })
+  public name: string;
+}
+```
+
+The complete list of valid attributes for property definitions can be found in
+LoopBack 3's [Model definition section](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#properties)
+
+### Defining Metadata By Hand?
+You might be wondering why we have to redefine the type information when
+it's already a part of the class definition.
+
+The property type data isn't available at runtime, only at compile time.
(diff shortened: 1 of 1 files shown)


pages/en/lb4/Schemas.md
```diff
@@ -9,9 +9,125 @@ summary:
...
+```
+
+## Other ORMs
+You might decide to use an alternative ORM/ODM in your LoopBack application.
+Loopback v4 no longer expects you to provide your data in its own custom Model
+format for routing purposes, which means you are free to alter your classes
+to suit these ORMs/ODMs.
+
+However, this also means that the provided schema decorators will serve no
+purpose for these ORMs/ODMs, and may even conflict with some of them.
```
raymondfeng: The decorators will attach metadata with the target class under `loopback` namespaces. I don't see much possibility of conflicts.

kjdelisle: What I meant by this was that other ORMs may have their own `@model` and `@property` decorators.

pages/en/lb4/Schemas.md
```diff
@@ -9,9 +9,127 @@ summary:
 ---
 {% include content/tbd.html %}
 
-> Stub page per [#554](https://github.com/strongloop/loopback-next/issues/554).
+<!-- TOC -->
```
crandmck: The doc site automatically creates a TOC for each page (unless you specify not to).  So, this TOC is redundant.

pages/en/lb4/Schemas.md
```diff
@@ -9,9 +9,127 @@ summary:
...
+    name: "title",
+    description: "The title of the Foo",
+    type: "string",
+  })
+  public title: string;
+}
+```
+
+The complete list of attributes in property definition can be found in here:
+https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#properties
```
crandmck: Why are we linking to LB3 doc?  If this information is applicable to LB4, then we should provide it in the `lb4` folder, with any changes specific to LB4.

crandmck (changes requested):

pages/en/lb4/Schemas.md
kjdelisle: This is for the legacy Juggler; as far as I know, we're not documenting the old legacy behaviour in LB 4
@raymondfeng @bajtos your thoughts?

kjdelisle: I should add that we're not yet set on what we intend for the new "juggler"; if we're supporting the same DSL as previously, we'll need to migrate documentation to that effect into the lb4 docs (and then this section would need re-writing to match that).

Regardless of what we go with, I'll make it a bit clearer that this is a link to the previous version's documentation.

dhmlau (approved):

crandmck (approved):

kjdelisle:
@raymondfeng @bajtos Just want to make sure that this is good from the technical standpoint as well; I remember the original task making mention of OpenAPI stuff, but that doesn't seem like the purpose of this section. Is the Schema meant to be a unification of the OpenAPI DSL and the Juggler/Legacy DSL(s)?

kjdelisle:
I guess whether or not it is, any changes necessary in that respect could probably wait until Core GA.

virkt25 (approved):

jannyHou (approved):

pages/en/lb4/Schemas.md
```diff
@@ -9,9 +9,115 @@ summary:
...
+@model({
+  name: "Foo",
+  properties: {
+    // define properties here.
+  },
+  settings: {
+    // etc...
+  }
+})
+class Foo {
```
b-admike: Since you're showing both ways of using the `@model` decorator, wouldn't it be 
```ts
model()
+export class Foo extends Entity {
  // etc...
}
```
here?

bajtos: I believe it's important for `Foo` to extend `Entity`, otherwise the model cannot be used with the juggler bridge.

Can we use a "real" model name like "Category" instead of "Foo" please?

As for `model()` usage, that's already documented in the previous section.

pages/en/lb4/Schemas.md
```diff
@@ -9,9 +9,115 @@ summary:
...
+would require a vastly more involved form of compile-time parsing to build all
+of the metadata required to reconstruct your complex object relationships at
+runtime.
+
+Additionally, even the primitive type handling wouldn't be able to
+determine connector-specific properties you might want to apply to your
+individual types.
+
+```ts
+export class Foo extends Entity {
```
bajtos: Let's call this `Category`.

pages/en/lb4/Schemas.md
```diff
@@ -9,9 +9,115 @@ summary:
...
+  @property name: string; // This would be easy.
+  @property bars: Bar[]; // This would be a lot more difficult!
```
bajtos: `@property() products: Product[]`

Notice that `@property` decorator is a function that must be invoked.

pages/en/lb4/Schemas.md
bajtos: AFAIK, our current plan is to let `@loopback/repository@4.x` to depend on `loopback-datasource-juggler@3.x`, i.e. there is no v4 of juggler planned. It makes more sense to me to keep the content in a single place in `doc/en/lb3`.

What are the downsides of such solution?

bajtos (dismissed):
PTAL ☝️

PR #465, opened 2017-09-20
improve lb4 repository docs

haoxins:


crandmck (commented):
This looks reasonable to me, but I'd like someone else to take a quick look...

@kjdelisle perhaps?

kjdelisle (commented):
I like the changes to the imported repository names (makes the examples easier to parse), but I noticed that we're changing `Product` to `Account` in some other examples. Given that we're changing the Repositories.md to use `ProductRepository`, wouldn't it be more consistent to have all of those items remain as `Product` instances? Alternatively, you could change `ProductRepository` to be `AccountRepository` to get the same effect.

Looks good, though. Approachable examples are key! :+1:

haoxins:
> Alternatively, you could change ProductRepository to be AccountRepository to get the same effect.

DONE :)

kjdelisle (approved):

kjdelisle:
LGTM

PR #467, opened 2017-09-20
Crafting loopback

raymondfeng:


crandmck (approved):
LGTM.  

Note that updates to the READMEs from loopback-component-passport and loopback-component-storage got into this PR via the slnode commit, which occurs periodically.   So there's no need to be concerned about that.

PR #468, opened 2017-09-21
Add security advisory sep 21 2017

loay:


pages/en/lb3/readmes/loopback-connector-kv-redis.md
```diff
@@ -1,3 +1,17 @@
-# loopback-connector-kv-redis
 
-The official Redis KeyValue connector for LoopBack.
+<?xml version="1.0" encoding="utf-8"?>
```
bajtos: This change looks suspicious to me, I think this file should be reverted?

loay: You are right .. I have no idea where did this file come from !!!!

loay: I got it .. another commit was made that I didn't do!! I will revert

crandmck: JFYI, these often get merged into doc commits from the Jenkins pipeline that runs the [`update-readmes.sh`](https://github.com/strongloop/loopback.io/blob/gh-pages/update-readmes.sh) script.  This script grabs READMEs from various LB repos to incorporate them into the doc repo.  See http://loopback.io/doc/en/contrib/Including-READMEs.html for more info.

Since Jenkins runs the script every time there is a commit to the loopback.io repo, they often get merged with random commits.  While disconcerting, this is normal and nothing to worry about.

crandmck (changes requested):
Please add this file to the sidebar nav, under "Security advisories" - https://github.com/strongloop/loopback.io/blob/gh-pages/_data/sidebars/lb3_sidebar.yml#L126

Should be pretty obvious how to add it, but LMK if you have questions.

PR #475, opened 2017-10-05
Update LoopBack 4 docs with rest refactor changes

kjdelisle:
## Overview

Refactor the docs to cover off the changes made to the framework architecture
- Areas where we previously bound REST-specific functions should now have updated examples that leverage the new method
- Testing-Your-Application required a large rewrite because
  - It wasn't compatible with the refactor
  - It didn't work before, either
  It's now a complete step-by-step set of working **JavaScript-based** examples that I've validated with my own repository: https://github.com/kjdelisle/hello-loopback/tree/doc-validation

### Related Issues
https://github.com/strongloop/loopback-next/issues/613

### DIFF
--- a/pages/en/lb4/Context.md
+++ b/pages/en/lb4/Context.md
@@ -61,8 +61,8 @@ Let's see this in action:
 ```js
 class MySequence extends DefaultSequence {
   handle(request, response) { // we provide these value for convenience (taken from the Context)
-    const req = this.ctx.getSync('rest.http.request');
-    // but it is still available in the sequence/request context
+    // but they are still available in the sequence/request context
+    const req = await this.ctx.get('rest.http.request');
     this.send(`hello ${req.params.name}`);
   }
 }
--- a/pages/en/lb4/Context.md
+++ b/pages/en/lb4/Context.md
@@ -12,57 +12,73 @@ summary:
 
 - An abstraction of all state and dependencies in your application.
 - Context is what LB uses to "manage" everything.
-- A global registry for anything/everything in your app (all configs, state, dependencies, classes, etc).
+- A global registry for anything/everything in your app (all configs, state,
+dependencies, classes, etc).
 - An IoC container used to inject dependencies into your code.
 
 Why is it important to you?
 
-- You can use the context as a way to give loopback more "info" so that other dependencies in your app may retrieve it (ie. a centralized place/global builtin/in-memory storage mechanism).
-- LoopBack can help "manage" your resources automatically (through [Dependency Injection](Dependency-injection.html) and decorators).
-- You have full access to updated/real-time application+request state at all times.
+- You can use the context as a way to give loopback more "info" so that other
+dependencies in your app may retrieve it (ie. a centralized place/global
+builtin/in-memory storage mechanism).
+- LoopBack can help "manage" your resources automatically (through
+[Dependency Injection](Dependency-injection.html) and decorators).
+- You have full access to updated/real-time application+request state at all
+times.
 
 LoopBack supports two types of context: Application-level and Request-level
 
 ## Application-level context (global)
 
-- stores all the initial and modified app state throughout the entire life of the app (while the process is alive)
-- Generally configured when the application is created (though other things may modify things in the context while alive/running)
+- stores all the initial and modified app state throughout the entire life of
+the app (while the process is alive)
+- Generally configured when the application is created (though other things may
+modify things in the context while alive/running)
 
 Here is a simple example:
 
 ```js
 const Application = require('@loopback/core').Application;
 const app = new Application(); // `app` is a "Context"
-class MySequence { ... }
-app.sequence(MySequence);
+class MyController { ... }
+app.controller(MyController);
 ```
 
-In this case, you are using our sequence sugar/helper to register a new default sequence. The important point to note is `MySequence` is actually registered into the Application Context (`app` is a Context).
+In this case, you are using the `.controller` helper method to register a new
+controller. The important point to note is `MyController` is actually registered
+into the Application Context (`app` is a Context).
 
 ## Request-level context (request)
 
-- dynamically created for each incoming server request
-- extends the application level context (to give you access to application level dependencies during the request/response lifecycle)
-- is garbage collected once the response is sent (memory management)
+Using [`@loopback/rest`](https://github.com/strongloop/loopback-next/blob/master/packages/rest) as an
+example, we can create custom sequences that:
+- are dynamically created for each incoming server request
+- extend the application level context (to give you access to application level dependencies during the request/response lifecycle)
+- are garbage collected once the response is sent (memory management)
 
 Let's see this in action:
 
 ```js
 class MySequence extends DefaultSequence {
   handle(request, response) { // we provide these value for convenience (taken from the Context)
-    const req = this.ctx.getSync('http.request'); // but it is still available in the sequence/request context
+    const req = this.ctx.getSync('rest.http.request');
+    // but it is still available in the sequence/request context
     this.send(`hello ${req.params.name}`);
   }
 }
 ```
 
 - `this.ctx` is available to your sequence
-- allows you to craft your response using resources from the app in addition to the resources available to the request in real-time (right when you need it)
-- `getSync` is one way to get stuff out of the context, there are many others, see below
+- allows you to craft your response using resources from the app in addition to
+the resources available to the request in real-time (right when you need it)
+- `getSync` is one way to get stuff out of the context, there are many others,
+see below
 
 ## Storing and retrieving items from a Context
 
-Items in the Context are indexed via a key and bound to a `ContextValue`. A `ContextKey` is simply a string value and is used to look up whatever you store along with the key. For example:
+Items in the Context are indexed via a key and bound to a `ContextValue`.
+A `ContextKey` is simply a string value and is used to look up whatever you
+store along with the key. For example:
 
 ```js
 // app level
@@ -71,18 +87,26 @@ app.bind('hello').to('world'); // ContextKey='hello', ContextValue='world'
 console.log(app.getSync('hello')); // => 'world'
 ```
 
-In this case, we bind the 'world' string ContextValue to the 'hello' ContextKey. When we fetch the ContextValue via `getSync`, we give it the ContextKey and it returns the ContextValue that was initially bound (we can do other fancy things too -- ie. instantiate your classes, etc)
+In this case, we bind the 'world' string ContextValue to the 'hello' ContextKey.
+When we fetch the ContextValue via `getSync`, we give it the ContextKey and it
+returns the ContextValue that was initially bound (we can do other fancy things
+too -- ie. instantiate your classes, etc)
 
-The process of registering a ContextValue into the Context is known as _binding_. Sequence-level bindings work the same way (shown 2 examples before).
+The process of registering a ContextValue into the Context is known as
+_binding_. Sequence-level bindings work the same way (shown 2 examples before).
 
 ## Dependency injection
 
 - Many configs are adding to the Context during app instantiation/boot time by you/developer.
-- When things are registered, the Context provides a way to use your dependencies during runtime.
+- When things are registered, the Context provides a way to use your
+dependencies during runtime.
 
-How you access these things is via low level helpers like `app.getSync` or the `sequence` class that is provided to you as shown in the example in the previous section.
+How you access these things is via low level helpers like `app.getSync` or the
+`sequence` class that is provided to you as shown in the example in the previous
+section.
 
-However, when using classes, LoopBack provides a better way to get at stuff in the contxet via the `@inject` decorator:
+However, when using classes, LoopBack provides a better way to get at stuff in
+the context via the `@inject` decorator:
 
 ```js
 const Application = require('@loopback/core');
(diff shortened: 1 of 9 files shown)


pages/en/lb4/Testing-Your-application.md
```diff
@@ -166,43 +210,142 @@ Run the tests again:
...
+  }
+}
+
+module.exports = HelloWorldApp;
+```
+
+Here is the swagger.json that will hook up your `GreetController` to the
+`/greet` path, and associate the `greet` function with a `GET` request.
+
+{% include code-caption.html content="src/apidef/swagger.json".js" %}
```
crandmck: Liquid syntax error (spurious quote).  S/be:
```
{% include code-caption.html content="src/apidef/swagger.json" %}
```

crandmck (changes requested):

bajtos:
> Testing-Your-Application required a large rewrite 

I am going to rewrite that page as part of https://github.com/strongloop/loopback-next/issues/553 anyways. Thanks for updating the code examples though!

pages/en/lb4/Context.md
```diff
@@ -12,57 +12,73 @@ summary:
...
+- extend the application level context (to give you access to application level dependencies during the request/response lifecycle)
+- are garbage collected once the response is sent (memory management)
 
 Let's see this in action:
 
 ```js
 class MySequence extends DefaultSequence {
   handle(request, response) { // we provide these value for convenience (taken from the Context)
-    const req = this.ctx.getSync('http.request'); // but it is still available in the sequence/request context
+    const req = this.ctx.getSync('rest.http.request');
```
bajtos: I'd prefer to not teach new developers about `getSync`, because it does not work for values requiring async computation. I am proposing to update this example as follows:

```js
class MySequence extends DefaultSequence {
  // we provide request & response for convenience (taken from the Context)
  async handle(request, response) {
    // but they are still available in the sequence/request context
    const req = await this.ctx.get('rest.http.request');
    this.send(`hello ${req.params.name}`);
  }
}
```

pages/en/lb4/Dependency-Injection.md
```diff
@@ -43,10 +43,16 @@ In LoopBack, we use [Context](Context.html) to keep track of all injectable depe
...
+```ts
+// TypeScript example
+
 import {BasicStrategy} from 'passport-http';
+import {Application} from '@loopback/core';
+import {RestServer} from '@loopback/rest';
+// basic scaffolding stuff happens in between...
 
-app.bind('authentication.strategy').to(new BasicStrategy(loginUser));
+const server = app.getServer(RestServer); // The REST server has its own context!
```
bajtos: `getServer` returns a promise, doesn't it?

```ts
const server = await app.getServer(RestServer);
```

pages/en/lb4/Dependency-Injection.md
```diff
@@ -43,10 +43,16 @@ In LoopBack, we use [Context](Context.html) to keep track of all injectable depe
...
+server.bind('authentication.strategy').to(new BasicStrategy(loginUser));
```
bajtos: I feel the necessity to configure things like authentication strategy and sequence handler at server-level (and not at app-level) is rather suboptimal UX. What used to be a single line of sync code before, becomes two lines of async/await code that cannot even happen in application's constructor (because it's async).

Can we come up with a better approach that's easier to use? I am fine to leave that out of scope of this patch if you prefer.

pages/en/lb4/Getting-started.md
```diff
@@ -28,21 +28,27 @@ The starting point of this flavor of JavaScript is [ECMAScript 6](http://www.ecm
...
-const app = new Application();
-app.handler((sequence, request, response) => {
-  sequence.send(response, 'hello world');
+const app = new Application({
+  components: [RestComponent],
 });
 
 (async function start() {
+  // Grab the REST server instance
+  const server = app.getServer(RestServer);
```
bajtos: `getServer` returns a promise, doesn't it?

I also find it weird that we are configuring the sequence handler every time the server is started, and that there is no sequence handler configured after application's constructor was called. (I think this is sort of an addition to my previous comment.)

pages/en/lb4/Reserved-binding-keys.md
```diff
@@ -88,12 +88,33 @@ import { CoreBindings } from '@loopback/authentication'
...
+
+|`rest.handler`|`HANDLER`|`HttpHandler`|The HTTP Request Handler|
+|`rest.port`|`PORT`|`number`|HTTP Port the application will run on|
+|`rest.http.request`|`Http.REQUEST`|`ServerRequest`|The raw `http` request object|
+|`rest.http.request.context`|`Http.CONTEXT`|`Context`|Request level context|
+|`rest.http.response`|`Http.RESPONSE`|`ServerResponse`|The raw `http` response object|
+|`routes.${route.verb}.${route.path}`||`RouteEntry`|Route entry specified in api-spec|
+|`rest.sequence`|`SEQUENCE`|`SequenceHandler`|Class that implements the sequence for your application|
+
 **Sequence Actions binding keys**
```
bajtos: I think this title should include "REST" keyword, now that the sequence actions listed below are specific to `RestBindings`.

pages/en/lb4/Routes.md
```diff
@@ -48,17 +52,23 @@ const spec = {
...
-const route = new Route('get', '/', spec, greet);
-app.route(route);
 
-app.start();
+(async function start() {
+  const server = await app.getServer(RestServer);
+  const route = new Route('get', '/', spec, greet);
+  server.route(route);
+  await app.start();
+})();
```
bajtos, line 61: This will swallow any errors thrown from `start`, causing "unhandled promise rejection" warnings. You should add `.catch()` handler at minimum.

It makes me wonder, can we move methods like `route`, `handler` and `sequence` to a different class, one that does not have any dependencies so that it will always resolve synchronously, and then inject this class as a RestServer dependency?

An mock-up:

```ts
// inside RestComponent (or a mixin?)
app.rest = new RestServerConfigurator();
app.bind(RestBindings.CONFIGURATOR).toDynamicValue(() => app.rest);

// in RestServer constructor
class RestServer {
  constructor(
    @inject(RestBindings.CONFIGURATOR) configurator: RestServerConfigurator,
    /*...*/) {
      // define routes, handler, sequence etc. from the configurator
    }
 }

// usage in applications
app.rest.route('get', '/', spec, greet);
```

This can be extended to support multiple rest servers by replacing `app.rest` with `app.${serverName}` , where the server name is the key used in `servers.${serverName}` binding and in `appConfig.servers` object.

Thoughts?

pages/en/lb4/Routes.md
```diff
@@ -77,7 +87,7 @@ const spec = {
   }
 };
 
-app.api(spec);
+server.api(spec);
```
bajtos, line 90: This feels very wrong to me. The API is tightly coupled with Controllers and controllers are defined at application level. As a user, I'd expect `api()` method to be available in the same place as `controller()`.

Also, if the API is server-specific, and we have multiple REST servers configured, then each server can be providing a different REST API. Is it a good idea?

pages/en/lb4/Sequence.md
```diff
@@ -100,17 +110,23 @@ class MySequence extends DefaultSequence {
...
+In order for LoopBack to use your custom sequence, you must register it on any
+applicable `Server` instances before starting your `Application`:
 
 ```js
 import {Application} from '@loopback/core';
+import {RestComponent, RestServer} from '@loopback/rest';
 
 const app = new Application({
-  sequence: MySequence
+  components: [RestComponent],
```
bajtos, line 121: This is a step backwards in terms of UX, at least IMO. Isn't it possible to configure the sequence via component-specific configuration? Something along the following lines:

```js
const app = new Application({
  components: [RestComponent],
  rest: {
    sequence: MySequence,
  }
});
```

bajtos (changes requested):
I am concerned that the recent changes in server-related API worsened user experience and made the application setup code too complex.

I am proposing to take another look at the current implementation and find a way how to make the setup easier to use, to ensure we keeping the high UX level which we started with.

pages/en/lb4/Context.md
kjdelisle: Ah, sorry, I must've glossed over that bit, was just correcting the injection key.

pages/en/lb4/Dependency-Injection.md
kjdelisle: We could easily have the examples use application-level context to do their work, since it's all available at the child context of the server, but then we're potentially encouraging a pattern that might prevent use of separate strategies in multiple server instances.

If we're going to alter the examples to simplify the bindings they work with, we'll need to make a section somewhere that details more advanced usage scenarios so that users properly separate their contextual bindings.

pages/en/lb4/Routes.md
kjdelisle: >This feels very wrong to me. The API is tightly coupled with Controllers and controllers are defined at application level. As a user, I'd expect api() method to be available in the same place as controller().

My understanding was that we might have multiple controllers servicing more than one server, either of the same protocol or otherwise. Combine this with the idea that the OpenAPI spec handling is REST-specific and we end up with server-based API binding instead.

>Also, if the API is server-specific, and we have multiple REST servers configured, then each server can be providing a different REST API. Is it a good idea?

I think it's a great idea! Your microservice could have a public-facing API that defines its consumable endpoints, and a private/internal API for infrastructure, exporting logs, administration, or whatever you can think of. Without server-level API binding, these scenarios aren't possible anymore.

Playing "Devil's Advocate" for a moment, if we want to strictly enforce the idea that a LoopBack application is a one-server microservice, then it would make sense to migrate some of this configuration back up to application level, and enforce the idea that you pick **one, and only one** server component to power the application. This would definitely make sense from a microservice architecture perspective, where you're minimizing the responsibility of each running instance and using other technologies to handle communication between services, log aggregation, etc.

I think trying to simultaneously support both philosophies is easily doable, but could be a documentation and UX nightmare for new users.

pages/en/lb4/Testing-Your-application-advanced.md
```diff
@@ -225,7 +234,21 @@ describe('smoke test', () => {
...
+  scaffolding to avoid compilation errors).
+  
+  The *reason* for this is that you want to make sure your test doesn't pass
+  without any assistance from the code under test. If it **DOES**, then you've
+  successfully demonstrated that your test itself is what's incorrect and you
+  can adjust it before getting too far down the road with the implementation.
+
+  I know this isn't a part of my PR, which is why I'm only leaving this comment
+  instead of fixing it.
+ -->
```
jannyHou, line 251: when I run `npm start` on local, there is an error complains
```
Liquid Exception: Invalid syntax for include tag: content="src/apidef/swagger.json".js" Valid syntax: {% include file.ext param='value' param2='value' %} in pages/en/lb4/Testing-Your-application.md
jekyll 3.3.1 | Error:  Invalid syntax for include tag:

  content="src/apidef/swagger.json".js"

Valid syntax:

  {% include file.ext param='value' param2='value' %}
```
branch `gh-pages` doesn't fail on it. 
This line is not directly modified in PR, maybe somewhere else affects it?

jannyHou: oops I commented in the wrong file that's why I didn't find it, should be
https://github.com/strongloop/loopback.io/pull/475/files#diff-ad241bc342976380fb64dab3b286802bR248
their names are so similar lol

pages/en/lb4/Context.md
```diff
@@ -12,57 +12,73 @@ summary:
...
+modify things in the context while alive/running)
 
 Here is a simple example:
 
 ```js
 const Application = require('@loopback/core').Application;
 const app = new Application(); // `app` is a "Context"
-class MySequence { ... }
-app.sequence(MySequence);
+class MyController { ... }
```
b-admike, line 43: Can we revert this since either can be used as an example, but in the `Request-level context` section we expand on `MySequence`?

pages/en/lb4/Sequence.md
kjdelisle: I think the current design supports your suggestion, so I'll change it.

pages/en/lb4/Routes.md
bajtos: Let's leave this out of scope of this documentation update.

bajtos (approved):
LGTM. Let's leave usage improvements that I suggested in my comments out of scope of this pull request.

pages/en/lb4/Routes.md
bajtos: > Your microservice could have a public-facing API that defines its consumable endpoints, and a private/internal API for infrastructure, exporting logs, administration, or whatever you can think of. 

This is a pretty cool idea which makes a lot of sense to me 👍 I agree we should support this use case.

My concern is about ease of use for people starting with the framework, notice how many of our examples increased in size.

Anyways, this is out of scope of this pull request, because we have already updated may other places (package READMEs, example repos) to use the convention described in this pull request. To use a better convention, we will need to update both the docs and those other places.

crandmck (approved):

bajtos:
I am going to land this pull request even if it's not perfect yet, so that I can start working on https://github.com/strongloop/loopback-next/issues/553

@kjdelisle please open a follow-up pull request to address remaining issues/comments.

I apologize for the inconvenience this may have caused.

PR #484, opened 2017-10-12
fixed example of hiding of all remote methods

buokae:


slnode:

buokae:
It was a bit of a pain to find out what all the names of all the methods are and when to prefix the name with "prototype." and when not, which is completely illogical to me. Also, some names actually disable more than one method, so in the js comment in the example I've added which endpoint(s) (is/are) disabled by a call to disableRemoteMethodByName.

crandmck (approved):

PR #489, opened 2017-10-16
Improve `Extending LoopBack 4`

raymondfeng:


virkt25 (approved):
Looks great. 

Just a side comment (not related to this PR directly) is that the first image seems complex (a bit overwhelming) for someone who might be new to the project and trying to learn. Not sure how to simplify it though.

PR #491, opened 2017-10-17
mixin and components

jannyHou:
connect to https://github.com/strongloop/loopback-next/issues/580

- Add mixin doc in `Mixin.md`
- Also add a section for javascript/typescript related concepts as mixin's parent section
- Add how a component interact with mixin in `Creating-component.md`


crandmck:
@jannyHou I'll wait til this is a bit further along to review it OK?  Pls ping me when it's ready for review.

jannyHou:
@crandmck yeah I am still adding stuff on local, thanks!

_data/sidebars/lb4_sidebar.yml
```diff
@@ -98,6 +98,15 @@ children:
   url: Crafting-LoopBack-4.html
   output: 'web, pdf'
 
+- title: 'Language Related Concepts'
```
bajtos: Is it worth starting a new top-level section instead of nesting this new "Mixin" page under "Key concepts"? I don't know what's the right answer, just something to consider.

pages/en/lb4/Creating-components.md
```diff
@@ -38,8 +38,8 @@ Providers are the mechanism allowing components to export values that can be use
 ```js
 import {Provider} from '@loopback/context';
 
-export class MyValueProvider {
-  value() {
+export class MyValueProvider implements Provider<string>{
+  value(): string {
```
bajtos: I think this should not be strictly necessary, I believe TypeScript is able to infer the return type from `Provider` definition and/or the actual value returned by the code of this method. Unless I am mistaken?

pages/en/lb4/Creating-components.md
```diff
@@ -90,12 +90,13 @@ A note on binding names: In order to avoid name conflicts, always add a unique p
...
 
 ```js
+import {Provider} from '@loopback/context';
 const request = require('request-promise-native');
 const weatherUrl =
   'http://samples.openweathermap.org/data/2.5/weather?appid=b1b15e88fa797225412429c1c50c122a1'
 
-export class CurrentTemperatureProvider {
-  async value() {
+export class CurrentTemperatureProvider implement Provider<number>{
```
bajtos: `implements`?

pages/en/lb4/Creating-components.md
```diff
@@ -109,14 +110,15 @@ In this case, LoopBack will wait until the promise returned by `value()` is reso
 In some cases, the provider may depend on other parts of LoopBack, for example the current `request` object. Such dependencies should be listed in provider's constructor and annotated with `@inject` keyword, so that LoopBack runtime can resolve them automatically for you.
 
 ```js
+import {Provider} from '@loopback/context';
 const uuid = require('uuid/v4');
 
-class CorrelationIdProvider {
+class CorrelationIdProvider implement Provider<any>{
```
bajtos: `implements`?

pages/en/lb4/Creating-components.md
bajtos: ❗️ actually, the code examples in this file are all using JavaScript, not TypeScript - see `js` after three backticks. I think you should revert type-related changes in this file.

_data/sidebars/lb4_sidebar.yml
jannyHou: @bajtos I feel at least now `Mixin` is not a LB core concept, we don't have mixin level context, and don't reply on it to scaffold the app, it is more like a JS/TS strategy to extend class, which make other concepts extendable, like application, controller, as long as it's a class.

there is some context of this discussion, I pinged Raymond about moving `mixin.ts` from `@loopback/repository` to `@loopback/core`, and he suggests we can have a `@loopback/lang` or `@loopback/common` module for JS/TS related stuff.

I created an issue for this idea: https://github.com/strongloop/loopback-next/issues/660

jannyHou:
Regarding new category `Lang`, there are some context in story https://github.com/strongloop/loopback-next/issues/660

pages/en/lb4/Lang-concepts.md
```diff
@@ -0,0 +1,13 @@
+---
+lang: en
+title: 'Language Related Concepts'
```
crandmck: Language-related concepts

_data/sidebars/lb4_sidebar.yml
crandmck: S/be 
Language-related concepts

pages/en/lb4/Creating-components.md
```diff
@@ -228,6 +230,33 @@ export class AuthenticationProvider {
 }
 ```
 
+## Extends Application
```
crandmck: This section has many grammatical and stylistic errors.  Rather than go through them laboriously here, I'd prefer to either edit the text in thebranch directly or wait til the PR lands and then edit the text (none of the changes would require technical review) in master (ie. gh-pages).  @jannyHou LMK which you'd prefer.

crandmck (changes requested):

pages/en/lb4/Creating-components.md
```diff
@@ -228,6 +230,33 @@ export class AuthenticationProvider {
...
+
+When bind a component to an app, you may want to mount the component's propertis
```
dhmlau: - `When bind a component to an app` --> When *binding* a component to an app
- typo: propertis  --> properties

pages/en/lb4/Creating-components.md
```diff
@@ -228,6 +230,33 @@ export class AuthenticationProvider {
 }
 ```
 
+## Extends Application
+
+When bind a component to an app, you may want to mount the component's propertis 
+and methods to application level context. 
+A stragety to do that is using mixin, which extends a class with new properties and methods.
```
dhmlau: - typo: stragety --> strategy
- Not sure if i understand what it means..  Is it sth like:
a strategy is to use mixin to extends a class with new properties and methods ?

pages/en/lb4/Creating-components.md
```diff
@@ -228,6 +230,33 @@ export class AuthenticationProvider {
...
+
+An example scenario would be an app has multiple components with repositories bound to each of them, 
+you can use a `RepositoryMixin` to mount all of the repositories to the app.
+
+If you are not familiar with concept `mixin`, check [Mixin](Mixin.htm) to know more,
```
dhmlau: suggested change.. simply say sth like:
For more details, see [Mixin](Mixin.htm).

pages/en/lb4/Mixin.md
```diff
@@ -0,0 +1,119 @@
...
+import {BaseClass} from '<path_exports_BaseClass>';
+import {FooMixin, BarMixin} from '<path_exports_Mixins>';
+
+let newClass = MixinBuilder.mix(baseClass).with(FooMixin, BarMixin);
+```
+
+## Define Mixin
+
+By defining a mixin, you create a mixin function that takes in a base class, 
+and returns a new class extends the base class with new properties and methods mixed to it.
```
dhmlau: perhaps: "returns a new class extends the base class.." --> "returns a new class extending the base class..."

pages/en/lb4/Creating-components.md
jannyHou: hi @crandmck I think now this PR is still waiting for technical review, I will let you know when it's done but before merging.

pages/en/lb4/Creating-components.md
kjdelisle: typo here in the code snippet, should be "implements"

pages/en/lb4/Creating-components.md
kjdelisle: implements

pages/en/lb4/Creating-components.md
```diff
@@ -90,12 +90,13 @@ A note on binding names: In order to avoid name conflicts, always add a unique p
...
 ```js
+import {Provider} from '@loopback/context';
 const request = require('request-promise-native');
 const weatherUrl =
   'http://samples.openweathermap.org/data/2.5/weather?appid=b1b15e88fa797225412429c1c50c122a1'
 
-export class CurrentTemperatureProvider {
-  async value() {
+export class CurrentTemperatureProvider implement Provider<number>{
+  async value():number {
```
kjdelisle: missing a space: `async value(): number`

pages/en/lb4/Creating-components.md
```diff
@@ -109,14 +110,15 @@ In this case, LoopBack will wait until the promise returned by `value()` is reso
...
 const uuid = require('uuid/v4');
 
-class CorrelationIdProvider {
+class CorrelationIdProvider implement Provider<any>{
   constructor(@inject('http.request') request) {
     this.request = request;
   }
 
-  value() {
+  value():any {
```
kjdelisle: missing a space: `value(): any`

kjdelisle (changes requested):
I know you said focus on the technical aspects, but the typos I left comments about are code style errors or outright syntax errors.

pages/en/lb4/Creating-components.md
jannyHou: @dhmlau yep.

pages/en/lb4/Creating-components.md
jannyHou: I would like to have the current version since that document is not a must read for people who already knows JS `mixin`. A simplified version implies `Mixin.md`contains something additional/different to the JS concept, but it doesn't :)

pages/en/lb4/Creating-components.md
jannyHou: horrible typo :(

pages/en/lb4/Using-components.md
```diff
@@ -35,4 +35,4 @@ In general, components can contribute the following items:
 In the future (before the GA release), components will be able to contribute additional items:
 
  - Models
- - [Repositories](Repositories.html)
+ - [Repositories](Repositories.html)
```
bajtos, line 38: Unrelated whitespace change, could you please revert?

pages/en/lb4/Mixin.md
```diff
@@ -0,0 +1,119 @@
...
+LoopBack provides a mixin builder function `MixinBuilder` to easily extend a class 
+with one or multiple mixins,
+you can call it to define the new extended class with fluent syntaxed API like:
+
+```ts
+import {MixiBuilder} from '@loopback/repository';
+import {BaseClass} from '<path_exports_BaseClass>';
+import {FooMixin, BarMixin} from '<path_exports_Mixins>';
+
+let newClass = MixinBuilder.mix(baseClass).with(FooMixin, BarMixin);
```
bajtos: This example has an important problem: it's assigning a class constructor to a variable `newClass`. The usual convention is to start class names with a capital letter, e.g. `NewClass`. However, because we are using `let` + assignment instead of `class`, the name `NewClass` will trigger a tslint error about violating naming conventions for variables.

IMO, we should not promote MixinBuilder and tell people to compose mixins directly via regular JS/TS function invocation and class inheritance:

```ts
class NewClass extends FooMixin(BarMixin(baseClass)) {
}
```

Another important difference to consider:

In my proposed approach, the javascript runtime see a correct class name (`NewClass`) and `NewClass.toString()` operator returns `class NewClass...`:

```
> NewClass.name
"NewClass"

> NewClass.toString()
"class NewClass extends Mixin1(Mixin2(BaseClass)) {
        }"
```

Compare it with the result of `MixinBuilder.mix().with()`, which incorrectly sets the constructor name to `BaseClass` and where `newClass.toString()` returns a fully mixed class definition that makes it difficult to see what have been added by what mixin:

```
> newClass.name
"BaseClass"

> newClass.toString()
"class extends superClass {
        constructor() {
            super(...arguments);
            this.mixinProp2 = 'mixinProp2';
        }
        static staticMixinMethod2() {
            return 'mixin2.static';
        }
        mixinMethod2() {
            return 'mixin2';
        }
    }"
```

I am using the following unit-test code to verify my assumptions and produce sample output: https://github.com/strongloop/loopback-next/blob/568c6e2e7242949e1e5ece040bc3a54d83ff1411/packages/repository/test/unit/mixin/mixin.ts#L53

cc @raymondfeng as the author of MixinBuilder

pages/en/lb4/Mixin.md
```diff
@@ -0,0 +1,119 @@
...
+keywords: LoopBack 4.0, LoopBack 4
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Mixin.html
+summary:
+---
+
+It is a commonly used JavaScript/TypeScript strategy to extend a class with new properties and methods.
+
+*question: does our mixinBuilder allow interface extends classes?*
```
bajtos: I am not sure if I understand this question. Interfaces are compile-time things that are used by TypeScript compiler to verify type correctness. They are omitted from the transpiled javascript output and they are not available at runtime.

I think it's possible to declare that an interface extends a class, which means that the interface must have all prototype members (properties & methods) of the class being extended. However, no implementation details (no JS code) from the class is inherited this way, it's all about compile-time shape of objects (the type system) only.

pages/en/lb4/Mixin.md
```diff
@@ -0,0 +1,119 @@
...
+
+Define mixin `timeStampMixin`:
+
+{% include code-caption.html content="Mixins/timeStampMixin.ts" %}
+
+```ts
+import {Class} from "@loopback/repository";
+
+export function timeStampMixin<T extends Class<any>> (baseClass: T) {
+  return class newClass extends baseClass {
```
bajtos: I think the usual convention for creating mixins is to use anonymous class:

```ts
return class extends baseClass {
  // ...
};
```

I think V8 is smart enough to infer a better class name in such case, most likely use baseClass's name. E.g. `timeStampMixin(UserController)` returns a class called `UserController` and `timeStampMixin(ProductController)` returns a class called `ProductController`, which is much better than having both calls return classes with the same name `newClass`.

pages/en/lb4/Mixin.md
```diff
@@ -0,0 +1,119 @@
...
+
+And define mixin `loggerMixin`:
+
+{% include code-caption.html content="Mixins/loggerMixin.ts" %}
+
+```ts
+import {Class} from "@loopback/repository";
+
+function loggerMixin<T extends Class<any>> (baseClass: T) {
+  return class newClass extends baseClass {
```
bajtos: Ditto, just `return class extends baseClass`

pages/en/lb4/Mixin.md
```diff
@@ -0,0 +1,119 @@
...
+class SimpleController {
+    constructor() {
+
+    }
+    greet() {
+        console.log('hi!');
+    }
+}
+
+let AdvancedController = MixinBuilder.mix(SimpleController).with(timeStampMixin, loggerMixin);
```
bajtos: As I commented above, for cases where the list of mixins is known at compile time, we should not be using `MixinBuilder`.

pages/en/lb4/Mixin.md
```diff
@@ -0,0 +1,119 @@
...
+
+{% include code-caption.html content="Mixins/loggerMixin.ts" %}
+
+```ts
+import {Class} from "@loopback/repository";
+
+function loggerMixin<T extends Class<any>> (baseClass: T) {
+  return class newClass extends baseClass {
+    // add a new method `logger()`
+    logger(str: string) {
```
bajtos: It's a good practice to give methods names that start with a verb, i.e. `log(str: string)` in this case.

bajtos (changes requested):
Please see my comments below.

In addition to them, please include reference to the following materials explaining ES2015 and TypeScript mixins in more details:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#Mix-ins
- http://justinfagnani.com/2015/12/21/real-mixins-with-javascript-classes/
- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html

pages/en/lb4/Mixin.md
b-admike: >we should not promote MixinBuilder and tell people to compose mixins directly via regular JS/TS function invocation and class inheritance:

I wonder if one of the objectives of our MixinBuilder is to compose mixins for this use case. If that's the case, IMO shouldn't the constructor name and class def be preserved by MixinBuilder? (unless there is a reason I'm not aware of)

pages/en/lb4/Creating-components.md
```diff
@@ -228,6 +230,33 @@ export class AuthenticationProvider {
...
+
+An example scenario would be an app has multiple components with repositories bound to each of them, 
+you can use a `RepositoryMixin` to mount all of the repositories to the app.
+
+If you are not familiar with concept `mixin`, check [Mixin](Mixin.htm) to know more, 
+it also introduces how to apply mixins by the shortcut function in LoopBack.
+
+From Janny:
+
+I am going to add a Lite version of `RepositoryMixin`'s code as the end of this section
```
b-admike: Here I think you have some options. You could:
 - Give an overview of `RepositoryMixin` and point to its implementation
 - Give a brief overview of `RepositoryMixin` with code snippets like you said, and point to the source code at the end

I think the first option might suffice, given that Mixins are described in detail in their own page. IMO, the main points this section should cover would be:
- How component developers can leverage mixins
- Guideline on how they should use mixins 
- other considerations that they should take into account

pages/en/lb4/Mixin.md
```diff
@@ -0,0 +1,119 @@
...
+
+```ts
+import {MixiBuilder} from '@loopback/repository';
+import {BaseClass} from '<path_exports_BaseClass>';
+import {FooMixin, BarMixin} from '<path_exports_Mixins>';
+
+let newClass = MixinBuilder.mix(baseClass).with(FooMixin, BarMixin);
+```
+
+## Define Mixin
```
b-admike: I wonder if we should walk readers through this section first before `Mixin Builder` section. IMO I see it as introducing the concept first, then showing them how we make their lives easier with our `MixinBuilder` :)

jannyHou: I also think of the sequence when wrote it. Eventually decided to introduce `MixinBuilder` first:

`MixinBuilder.mix(baseClass).with(FooMixin, BarMixin)` is a very straight forward syntax to give a new  mixin learner a big picture of what is the relation between class and mixin, then they go into section#define-mixin to dig more into implementation and usage. And to tell people already know mixin how `MixinBuilder` provides a prettier syntax to apply mixins.

pages/en/lb4/Creating-components.md
jannyHou: @bajtos good catch...I thought the doc is out-of-date so update it according to our latest tsdoc in `provider.ts`. But didn't think it in the JS compatible direction.

Let me verify it.

pages/en/lb4/Mixin.md
jannyHou: I should have removed this, it is a question I am curious to know when reading typescript's class docuemnt:
search "Interfaces Extending Classes" on page https://www.typescriptlang.org/docs/handbook/interfaces.html

`MixinBuilder`'s input is typed to `Class`(we defined in [common-types.ts](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/common-types.ts))

```ts
constructor(public baseClass: Class<any>) {}
```

So I am wondering would `Interface` work as well.

pages/en/lb4/Mixin.md
jannyHou: @raymondfeng Could you take a look of comment https://github.com/strongloop/loopback.io/pull/491/files#r145687471? Thanks!

My understanding of `MixinBuilder` is it provides a prettier syntax to compose mixins to a class, but @bajtos has a good point in comment above.

raymondfeng: I borrowed the idea from http://justinfagnani.com/2015/12/21/real-mixins-with-javascript-classes/

jannyHou: FYI 
> Compare it with the result of MixinBuilder.mix().with(), which incorrectly sets the constructor name to BaseClass and where newClass.toString() returns a fully mixed class definition that makes it difficult to see what have been added by what mixin:

This could be solved by defining `newClass` as 
`newClass = class extends MixinBuilder.mix(BaseClass).with(Mixin1, Mixin2) {};`

Then print out:
```ts
console.log(newClass.name);
// newClass
console.log(newClass.toString());
// class extends _1.MixinBuilder.mix(BaseClass).with(Mixin1, Mixin2) {}
```

---

The problem of `MixinBuilder` from my test is: function `with()` doesn't set its returned type as a class extends the `BaseClass`:
[click to see example test](https://github.com/strongloop/loopback-next/blob/master/packages/repository/test/unit/repository-mixin/repository-mixin.test.ts#L60)
```ts
let AppWithRepoMixin = class extends MixinBuilder.mix(Application).with(RepositoryMixin){};

const myApp = new AppWithRepoMixin();
const boundComponentsBefore = myApp.find('components.*').map(b => b.key);
``` 
fails when compiling, because compiler doesn't know `myApp` is an instance of `Application`, therefore doesn't know it has a prototype function `find()` returning an `Array`. The code calls `map` with `any` instead of `Array` and fails.

This looks like a stop for people wanting to use it...

I am going to defer writing doc for `MixinBuilder` and create an issue for it. I still love the fluent api and would like to dig more :star:

pages/en/lb4/Using-components.md
jannyHou: weird...I could not see where is the whitespace(difference), I copy paste the original file but diff is still here...

jannyHou:
Apply feedback in https://github.com/strongloop/loopback.io/pull/491/commits/8b43c0f7121676320295eee5863554d175d03ba2:

- rename sidebar title to `Lauguage-related concepts`
- removed document of `MixinBuilder`, details see discussion https://github.com/strongloop/loopback.io/pull/491#discussion_r145687471 and issue https://github.com/strongloop/loopback-next/issues/673 
- complete the creating component doc

crandmck (approved):

b-admike (approved):

jannyHou:
@bajtos @kjdelisle feedback applied, PTAL again, thanks!

pages/en/lb4/Creating-components.md
```diff
@@ -228,6 +228,66 @@ export class AuthenticationProvider {
 }
 ```
 
+## Extends Application with Mixin
+
+When binding a component to an app, you may want to extend the app with component's
```
kjdelisle: `components`

pages/en/lb4/Mixin.md
```diff
@@ -0,0 +1,123 @@
...
+
+A good approach to apply mixins is defining them as sub-class factories.
+Then declare the new mixed class as:
+
+```js
+class MixedClass extends MixinFoo(MixinBar(BaseClass)) {};
+```
+
+Check article [real mixins with javascript classes](http://justinfagnani.com/2015/12/21/real-mixins-with-javascript-classes/) 
+to know more about it.
```
kjdelisle: `to learn more`.

pages/en/lb4/Mixin.md
```diff
@@ -0,0 +1,123 @@
...
+    }
+    greet() {
+        console.log('hi!');
+    }
+}
+```
+
+Now let's add mixins to it:
+
+- A time stamp mixin that adds a property `createdAt` to record when a controller instance get created.
```
kjdelisle: `to a record when a controller instance is created`

pages/en/lb4/Mixin.md
```diff
@@ -0,0 +1,123 @@
...
+
+- A logger mixin that adds a method `log` to log the `string` you provide.
```
kjdelisle: `A logger mixin to provide logging tools`

pages/en/lb4/Using-components.md
kjdelisle: Sounds like a character formatting problem (was this file made on Windows or something? :P)

kjdelisle (changes requested):
Minor nits.

pages/en/lb4/Creating-components.md
jannyHou: @kjdelisle I think it should be `the component's`

whole sentence is:
`When binding a component to an app, you may want to extend the app with the *component's* properties and methods. `

kjdelisle: I don't see `with the component's` in the existing snippet. ;)
I'm alright with that change as well.

jannyHou:
@kjdelisle thank you :) https://github.com/strongloop/loopback.io/pull/491/commits/8c8890dcf1dd00d14cb4ac26cef74128cc67550a

Really couldn't figure out how `- [Repositories](Repositories.html)` differs from the original file... so I leave it like this :(

pages/en/lb4/Creating-components.md
```diff
@@ -228,6 +228,66 @@ export class AuthenticationProvider {
 }
 ```
 
+## Extends Application with Mixin
+
+When binding a component to an app, you may want to extend the app with the component's 
+properties and methods. 
+This can be achieved by using mixin. 
+
+If you are not familiar with concept `mixin`, check [Mixin](Mixin.htm) to know more.
```
kjdelisle: `If you are not familiar with the mixin concept, see [Mixin](Mixin.htm) to learn more.`

pages/en/lb4/Creating-components.md
```diff
@@ -228,6 +228,66 @@ export class AuthenticationProvider {
 }
 ```
 
+## Extends Application with Mixin
+
+When binding a component to an app, you may want to extend the app with the component's 
+properties and methods. 
+This can be achieved by using mixin.
```
kjdelisle: `by using mixins`

kjdelisle (changes requested):
Two more tiny fixes

pages/en/lb4/Mixin.md
bajtos: > I am going to defer writing doc for MixinBuilder and create an issue for it. I still love the fluent api and would like to dig more ⭐️

@jannyHou could you please paste a link to that new issue here for posterity?

bajtos (approved):

pages/en/lb4/Mixin.md
jannyHou: @bajtos sure, sorry forgot to link it here: https://github.com/strongloop/loopback-next/issues/673

jannyHou:
@slnode test please

kjdelisle (approved):

jannyHou:
@slnode test please

jannyHou:
@slnode test please

Issue #315, opened 2017-03-08
Make separate article for access control using multiple user models


crandmck:
http://loopback.io/doc/en/lb3/Authentication-authorization-and-permissions.html#preparing-access-control-models

Pull into a separate article.

n.b. https://github.com/strongloop/loopback/pull/3231#discussion_r104117107

NOTE from bajtos:
> I setup http://ibm.biz/setup-loopback-auth shortcut to point to http://loopback.io/doc/en/lb3/Authentication-authorization-and-permissions.html#preparing-access-control-models. If/when you move the setup instructions to a different page, then we need to change the target url of the shortcut http://ibm.biz/setup-loopback-auth points to.

dhmlau:
Closing due to inactivity.

PR #492, opened 2017-10-17
Update Authentication-authorization-and-permissions.md

ghost:
so small title, 4 #


slnode:

crandmck:
I agree the title is small, but it's a 5th-level (`#####`) heading because it comes under a 4th-level header (`#### Access control with a single user model`).  It's not a peer, but a child of that heading.

@dotu Thanks for the PR, but a better way to address this is #315 .  I'll close this, but will try to prioritize that work.  Thanks for bringing it (back) to my attention.

PR #493, opened 2017-10-18
Test and clean up Installation article

crandmck:
I ran through the [Installation](http://loopback.io/doc/en/lb4/Installation.html) article, and there were some things missing, some things duplicated and one contradiction.  This PR addresses these issues.

The contradiction was that in one place it said that `tsconfig.json` must have `'target': 'es6'` then it said `"target": "es2017"` .  I'm assuming the latter is correct since we say elsewhere that we're using ES2017 and that is one of the options given in the `tsconfig.json` created by `tsc --init`.

Not addressed in this PR:  Duplication between this article and http://loopback.io/doc/en/lb4/Getting-started.html 
IMHO we should have just one article that covers installation and "Hello World" example.


dhmlau:
@crandmck , not sure if that's the right place to comment..  i was actually thinking whether there's a need for Installation page at all.  Here is my thoughts: 
- we may not need the Hello World example because there's already a hello world in the Getting Started page that should serve the same purpose
- If we don't suggest people to install typescript and ts-node globally (as @bajtos pointed out earlier), then there is only Node to install.  If that's the case, there's no need to have a separate page.  

@crandmck @bajtos , what do you think?

crandmck:
In general, I think an "Installation" page is worthwhile, because people expect it, and it provides a place to add Prerequisites and other useful info.  All the other LoopBack versions have it (as well as Express, etc.).  Even if it's just `npm install xxx` I think it's worthwhile to have a separate page.

> If we don't suggest people to install typescript and ts-node globally ...

I guess I missed this discussion.  I understand that installing packages globally can be an issue, but both [TypeScript docs](https://www.typescriptlang.org/index.html#download-links) and ts-node docs say to install globally, so why would we say differently?  Also, if someone doesn't do it, then `tsc` command won't work everywhere, right? NOT being able to assume that they are installed globally will complicate the docs (we will need a disclaimer/explanation everywhere docs say to run `tsc` or `ts-node`). 

If someone really wants to develop a LB4 app or component, wouldn't they want to install these tools globally?  Seems odd NOT to.  Certainly, when we have the LB4 CLI, we'll tell them to install it globally, so why wouldn't the same apply to these tools?

> we may not need the Hello World example because there's already a hello world in the Getting Started page that should serve the same purpose

Yes, I agree it should be in only one place, and it belongs in "Getting Started".  We should also remove "Prerequisites" and "Installation" sections in GS that duplicate what's in "Installation".
Right now there is a LOT of duplication btw these 2 articles.

Currently, the example in http://loopback.io/doc/en/lb4/Installation.html#create-a-hello-world-project is different (a bit simpler) than that in http://loopback.io/doc/en/lb4/Getting-started.html.  We should discuss which is preferable.  In general, it should the the simplest possible app.
Also the process should use `npm init` and `tsc init` to create `package.json` and `ts-config.json` (then edit as needed) as I did in this PR, since that's the standard tooling workflow.  IMHO it's more instructive to do it that way than to just copy/paste the whole file.  We could also offer both options.

crandmck:
@dhmlau said:

> There could be other implications in dropping support of Node 6, so we’re planning to have a discussion on that.  In the meanwhile, I’d think we’re not dropping support for Node 6.  Can we say we support node 6 or 8, but highly recommend 8 because xxxxxx ?

@bajtos asaid:
> `target` depends on what Node.js version your are using:
> - On Node.js 6.x, you need `target: 'es6'`. 
> - On Node.js 8.x, you should use `target: 'es2017'`. 
> - If you are writing a module that can be consumed by both groups of people (Node.js 6.x and 8.x), then IMO it's best to compile for both targets and decide at runtime which version to load - that's what we are doing in loopback-next packages.
>
>  As for installation instructions, I think we should nudge people towards Node.js 8.x setup, and perhaps tell them what to do differently if they are still on Node.js 6.x.

Thanks, I'll edit this PR accordingly.

@dhmlau This is an example of the type of detail that would go into the "Installation" page, and why I think it's worthwhile to keep one!

dhmlau:
@crandmck  sorry about the late reply. 
so going through the comments, I think we have:
- keeping the installation page
- Node 8 is highly recommended, but can run on Node 6 
- Different `target` in tsconfig.json for different version of Node
- remove the example in Installation page

Question: do we want `ts-node`?  @bajtos , you recommend using it...

crandmck:
Thanks @dhmlau I've edited the files based on the above (I'm not _that_ fast--I had already started working on it ;-) .  Currently this PR moves all the installation prerequisites into the "Installation" article and I have combined the two "Hello World" projects into "Getting Started", with the a basic and a slightly more complex TypeScript procedure.

I ran through the procedures several times myself to make sure they all work.

> Question: do we want ts-node? @bajtos , you recommend using it...

Currently, I have the user install `ts-node` globally.  That seems to work well, but I don't feel strongly one way or the other.

Also, instead of creating`package.json` manually, I have the user  create it with `npm init` and modify it via`npm install -s ... `, which is simpler and less error prone IMO.

I also have them use `tsc --init` to create `tsconfig.json`.  While this has the same benefits as `npm init`, it does add a ton of commented-out lines, which makes it a bit hard to read, so I could be convinced to go the other way.

One thing that's missing is some explanation of what's actually going on.  We can land this PR without that, but eventually I think it would benefit from some commentary so that the user actually learns about LB 4 rather than just going thru the motions, as it were....  Just a few sentences or choice paragraphs here and there with links our for more info.

crandmck:
One other small thing I forgot to mention.  I modified the Hello World program thus:
```
app.bind('message').to('Hello world!');
app.get('message').then(value => {
  console.log(value);
});
```
So it prints "Hello world!" to the console rather than just "world".  This seems more logical.  

I did something similar to the "advanced" TypeScript Hello World.

Not a huge deal, but thought I'd mention.

crandmck:
@dhmlau @bajtos PTAL - This is ready to land IMO.  

If it's easier, I can merge it, and you can review _in situ_.

pages/en/lb4/Getting-started.md
```diff
@@ -5,35 +5,202 @@ keywords: LoopBack 4.0, LoopBack 4
...
- - ES2015-modules
+With LoopBack 4 you can code in JavaScript or TypeScript.  Regardless of which you use, you'll need to do the following:
 
-**All examples and snippets in the documentation use this flavor of JavaScript.**
+ 1. Create a new directory called `lb4-hello-world` and make that your current directory:
+ ```
+ $ mkdir lb4-hello-world
+ $ cd lb4-hello-world
+ ```
+ 1. Create a `package.json` by following the instructions below.
```
dhmlau: there are 2 `1.`  line 16 and line 21.

dhmlau (approved):
LGTM with the minor change in line 21.

pages/en/lb4/Getting-started.md
crandmck: In markdown, an ordered list items can be denoted by any number as the first character on a line, followed by a dot, e.g. 
```
1. One
1. Two
1. Three
```
And it is rendered as 

1. One
1. Two
1. Three

So, what's there is correct.

pages/en/lb4/Getting-started.md
```diff
@@ -5,35 +5,202 @@ keywords: LoopBack 4.0, LoopBack 4
...
+{
+  "compilerOptions": {
+   ...
+    "target": "es6",
+    ...
+  }
+}
+```
+
+**With Node 8.x**:
```
bajtos: Nitpick: can we reverse the order and show instructions for Node.js 8.x first, 6.x second?

pages/en/lb4/Getting-started.md
```diff
@@ -5,35 +5,202 @@ keywords: LoopBack 4.0, LoopBack 4
...
+  console.log(value);
+});
+```
+
+### Run the project
+
+Then run `index.ts` by entering this command:
+
+```
+ts-node index.ts
```
bajtos, line 195: Initially we were using `ts-node` because it promised simpler project setup. Eventually we got frustrated with its shortcomings and ditched it out.

I think that for a quick & short Getting Started Guide it's ok to use ts-node, but we should not be promoting it anywhere else.

As an alternative to a global install of `ts-node`, I am suggesting to define a "start" script that invokes `ts-node index.ts` and move `ts-node` to project dependencies.

```
{
  // ...
  scripts: {
    // ...
    "start": "ts-node index.ts"
  },
  "dependencies": {
    // ...
    "ts-node": "^3.3.0"
  }
}
```

pages/en/lb4/Getting-started.md
```diff
@@ -5,35 +5,202 @@ keywords: LoopBack 4.0, LoopBack 4
...
+```
+
+You should see "world" written to the console.
```
bajtos: ... see "Hello world" ...

pages/en/lb4/Getting-started.md
```diff
@@ -5,35 +5,202 @@ keywords: LoopBack 4.0, LoopBack 4
...
-Make sure you have Node.js 8.0.0 or higher installed.
+Install the LoopBack REST package with npm by entering this command:
 
-## A simple example
+```js
+npm i -s @loopback/rest
+```
 
-This example creates an `Application` and a `RestServer` that responds to all HTTP requests with the text "Hello World".
+This adds the `@types/node` package to the `dependencies` property of `package.json`.
```
bajtos: This adds `@loopback/rest`, not `@types/node`.

pages/en/lb4/Getting-started.md
```diff
@@ -5,35 +5,202 @@ keywords: LoopBack 4.0, LoopBack 4
...
 
-### Building a LoopBack 4 application
-First create a new folder.  Then, create three files within this folder: `index.ts`, `tsconfig.json` and `package.json`.  
+
+### Install Node type definitions package
+
+Install the `@types/node` package as a development dependency by entering this command:
+
+```js
+npm i --save-dev @types/node
```
bajtos: If we are using `ts-node` to run the app, then `@types/node` is a regular runtime (production) dependency.

It would be a dev-dependency if we were using `tsc` to transpile from TS to JS at dev-time and then run the transpiled javascript in production.

pages/en/lb4/Getting-started.md
```diff
@@ -108,15 +253,11 @@ To run the application, enter:
 npm start
```
bajtos: This is using `npm start` script mentioned by my earlier comment. I think it's not defined if users follow the instructions in the revision of this document proposed by this pull request.

pages/en/lb4/Installation.md
```diff
@@ -12,111 +12,26 @@ summary:
 
 Install the following:
 
-- [Node.js](https://nodejs.org/en/download/) version 7 or higher.
+- Node.js, either:
+  - [Version 6.x](https://nodejs.org/en/download/): The long-term support (LTS) version.
+  - **Recommended**: [Version 8.x](https://nodejs.org/en/download/current/): The latest/current version.
```
bajtos: Can we reverse the order, start with the recommended version 8.x, and then mention that 6.x can be used too?

pages/en/lb4/Installation.md
```diff
@@ -12,111 +12,26 @@ summary:
...
 - [TypeScript](https://www.typescriptlang.org/index.html#download-links) version 2 or higher:
 
-   ```
-   npm i -g typescript
-   ```
-- [TypeScript Node](https://github.com/TypeStrong/ts-node#installation) version 3 or higher:
+  ```
+  $ npm i -g typescript
```
bajtos: As was pointed out earlier, global dependencies are an anti-pattern in Node.js world. Projects should list their build tooling dependencies in their package.json and then provide scripts like `npm run build` to invoke them. This is possible because when a package script is invoked by `npm`, `npm` automatically adds dev-dependencies to `PATH`.

I am ok to tell people to use globally installed `tsc` and `ts-node` to get them started quickly, but I think we should also tell them this is a shortcut not suitable for longer term, and show them how to setup their TypeScript project properly. The project https://github.com/strongloop/loopback4-extension-starter has a pretty good setup, I am not sure if other example projects have picked it up yet.

bajtos (commented):
+1 for restructuring the content into two pages "Installation" and "Getting Started" and removing duplications 👍 

I have few comments and concerns about the actual text, please see below.

pages/en/lb4/Getting-started.md
crandmck: Done.

pages/en/lb4/Installation.md
crandmck: OK, I added a note about this just being a shortcut and not best practice.  I'd be OK with changing this procedure to "do it properly" if you think that's preferable.

crandmck:
@bajtos, I incorporated your input.  PTAL.

pages/en/lb4/Getting-started.md
crandmck: I didn't do this, but if you feel strongly that it's preferable, I'll make that change.

bajtos (approved):
LGTM 👍 

Regarding `npm start` vs running `ts-node` directly, I think the current content is good enough.

PR #497, opened 2017-10-24
docs: add new security advisory

virkt25:


pages/en/lb3/Security-advisory-10-24-2017.md
```diff
@@ -0,0 +1,47 @@
...
+lang: en
+layout: page
+toc: false
+keywords: LoopBack
+tags: security
+sidebar: lb3_sidebar
+permalink: /doc/en/lb3/Security-advisory-10-24-2017.html
+---
+*   **Security risk**: Medium
+*   **Vulnerability**: Multi User Password Reset Exploit
```
crandmck: Multi-user password reset exploit

pages/en/lb3/Security-advisory-10-24-2017.md
```diff
@@ -0,0 +1,47 @@
...
+
+See [issue](https://github.com/strongloop/loopback/issues/3577) for more details.
+
+### Reported by
+
+GitHub user [sebastianfelipe](https://github.com/sebastianfelipe) via [Issue #3577](https://github.com/strongloop/loopback/issues/3577).
+
+### Versions affected
+
+loopback < 3.16.0 >= 3.3.0
```
crandmck: This is confusing.  Does it mean versions below 3.16.0 and above 3.3.0?  Based on the subsequent text, I think it's just versions below 3.16.0.  I don't understand what the `>= 3.3.0` means.

crandmck (changes requested):

crandmck:
Also, you need to add the new page to `_data/sidebars/lb3_sidebar.yml`, i.e.

```
- title: 'Security advisories'
  url: Security-advisories.html
  output: 'web, pdf'
  children:

  - title: '10-24-2017'
    url: Security-advisory-10-24-2017.html
    output: 'web, pdf'
```

pages/en/lb3/Security-advisory-10-24-2017.md
virkt25: It does mean versions below 3.16.0 and above or equal to 3.3.0. I can see why it's confusing (I was using this as snyk uses a similar syntax). Which out of the two options below would you suggest is better?

- 3.3.0 <= loopback < 3.16.0
- loopback (versions 3.3.0 or greater till version 3.16.0)

dhmlau: IMHO, the 2nd option is more appropriate, because we're not defining a variable range. :) 
In terms of wording, I'm wondering whether `.. 3.3.0 or higher up till 3.16.0` is better because we usually use `higher` instead of `greater` to describe versioning.  @crandmck probably knows best.

dhmlau (approved):

crandmck:
Thanks @virkt25 ... almost -- please put the new article in chronological order under "Security advisories" in the sidebar, i.e. it should be first, not last child.

crandmck:
LGTM, ty!

PR #499, opened 2017-10-25
Add READMEs for examples, add to sidebar, add sh script and Jenkinsfile

crandmck:
This brings LB4 in line with how v3 and v2 handle READMEs of examples, etc., that are in separate repos.  See [Including READMES](http://loopback.io/doc/en/contrib/Including-READMEs.html) for details.

This PR adds the READMEs for the 3 examples: 
- loopback-next-example
- loopback-next-hello-world
- loopback4-extension-starter

And it adds a new shell script, `update-readmes-lb4.sh` that pulls them from npm or GitHub if the module is not published to npm.  Also adds the script to the Jenkinsfile that runs whenever the repo is updated, so the READMEs are kept in sync with the other repos.

NOTE: Currently, the READMEs themselves need some editing, but since they live in other repos, that should be done in PRs in those repos.  IMHO, pre-beta it's OK if they are not polished, complete, etc., but we should get them fixed for beta release.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,6 +14,7 @@ node('linux && git') {
   stage('download updates') {
     sh './update-readmes.sh'
     sh './update-community-readmes.sh'
+    sh './update-readmes-lb4.sh'
   }
   stage('check updates') {
      sh 'git status'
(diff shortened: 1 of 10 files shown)


kjdelisle:
I think we were planning to rename some of those repos (as well as the aforementioned cleanup).
Not sure if we're ready for this just yet.

crandmck:
Renaming is not a big deal, easy enough to change the names in the script, etc.

> Not sure if we're ready for this just yet.

If you think we should wait, that's fine... We can keep the PR open and change it when things come together.

PR #500, opened 2017-10-26
Update rm-options.md

STRML:
Add `documented` options for both method & accepts definition.

Related: https://github.com/strongloop/loopback-swagger/pull/107

### DIFF
--- a/_includes/content/rm-options.md
+++ b/_includes/content/rm-options.md
@@ -68,6 +68,13 @@ accepts: [
       <td> </td>
     </tr>
     <tr>
+      <td>documented</td>
+      <td>
+        If set to <code>false</code>, this method will not be present in generated documentation.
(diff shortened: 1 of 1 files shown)


crandmck:
Thanks for the PR @STRML .

Just to be really clear, I think it should say "...this method will not be present in generated **Swagger API specification** documentation." (IIUC)

STRML:
Well, will it always be Swagger? It is becoming OpenAPI soon and I assume the intention is to possibly hook up more generators at another time.

crandmck:
Oh right, it's properly called "OpenAPI" now, so you should say "OpenAPI (formerly Swagger)".  

When and if you hook up other generators, then we can add those, too.

crandmck:
Thanks!

PR #501, opened 2017-10-26
docs: add section on HttpErrors in Controllers

shimks:
## Description
- add blurb on HttpErrors under Controllers
- connected to https://github.com/strongloop/loopback-next/issues/551

### DIFF
--- a/pages/en/lb4/Controllers.md
+++ b/pages/en/lb4/Controllers.md
@@ -156,3 +156,52 @@ class HelloController {
 - A `Repository` is LoopBack's database abstraction. See [Repositories](Repositories.html) for more.
 - `@get('/messages')` creates the `Route` for the Operation using `app.route()`.
 - `@param.query.number` adds a `number` param with a source of `query`.
+
+## Handling Errors in Controllers
+
+In order to specify errors for controller methods to throw, the class `HttpErrors` is used. `HttpErrors` can be found in the `@loopback/rest` package.
+
+The example below shows how `HttpErrors` can be used inside a controller method and how it can be tested.
+
+```js
+// the test
+import {HelloController} from 'path.to.controller'
+import {expect} from '@loopback/testlab'
+// ...
+
+// ...
+describe('Hello Controller', () => {
+  // ...
+  it('returns 412 Precondition Failed for non-positive limit', () => {
+    const controller = new HelloController();
+    try {
+      controller.list(0);
+      throw new Error('should have thrown an error');
+    } catch (err) {
+      expect(err).to.have.property('statusCode', 412);
+      expect(err.message).to.match(/precondition failed/i);
+    }
+  })
+  // ...
+})
+// ...
+```
+```js
+// the controller
+import 'HttpErrors' from '@loopback/rest'
+
+class HelloController {
+  constructor() {
+    this.messages = new Repository('messages');
+  }
+  @get('/messages')
+  @param.query.number('limit')
+  list(limit = 10)
+    if (limit < 1)
+      throw new HttpErrors.PreconditionFailed('limit is non-positive');
+    else if (limit > 100)
+      limit = 100; // your logic
+    return this.messages.find({limit});
+  }
+}
+```
--- a/pages/en/lb4/Controllers.md
+++ b/pages/en/lb4/Controllers.md
@@ -140,20 +140,24 @@ app.controller(MyController);
 Below is an example Controller that uses several built in helpers (decorators). These helpers give LoopBack hints about the Controller methods.
 
 ```js
+import 'HelloRepostory' from 'path.to.repository';
+import 'HelloMessage' from 'path.to.type';
+
 class HelloController {
   constructor() {
-    this.messages = new Repository('messages');
+    this.repository = new HelloRepository(); // our repository
   }
   @get('/messages')
   @param.query.number('limit')
-  list(limit = 10) {
+  async list(limit = 10): Promise<HelloMessage[]> { // returns a list of our objects
     if (limit > 100) limit = 100; // your logic
-    return this.messages.find({limit});
+    return await this.repository.find({limit}); // a CRUD method from our repository
   }
 }
 ```
 
-- A `Repository` is LoopBack's database abstraction. See [Repositories](Repositories.html) for more.
+- `HelloRepository` extends from `Repository`, which is LoopBack's database abstraction. See [Repositories](Reositories.html) for more.
+- `HelloMessage` is the arbitrary object that `list` returns a list of.
 - `@get('/messages')` creates the `Route` for the Operation using `app.route()`.
 - `@param.query.number` adds a `number` param with a source of `query`.
 
@@ -173,7 +177,7 @@ Listed below are some of the most common error codes. The full list of supported
 |503        |ServiceUnavailable   |
 |504        |GatewayTimeout       |
 
-The example below shows how `HttpErrors` can be used inside a controller method and how it can be tested.
+The example below shows the previous controller revamped with `HttpErrors` along with a test to verify that the error is thrown properly.
 
 ```js
 // the test
@@ -181,11 +185,11 @@ import {HelloController} from 'path.to.controller';
 import {HttpErrors, expect} from '@loopback/testlab';
 
 describe('Hello Controller', () => {
-  it('returns 422 Unprocessable Entity for non-positive limit', () => {
+  it('returns 422 Unprocessable Entity for non-positive limit', async () => {
     const controller = new HelloController();
     var errCaught: HttpErrors;
     try {
-      controller.list(0.4); // an HttpError should be thrown here
+      await controller.list(0.4); // an HttpError should be thrown here
     } catch (err) {
       errCaught = err;
     }
@@ -203,13 +207,13 @@ import 'HelloMessage' from 'path.to.type';
 
 class HelloController {
   constructor() {
-    this.repository = new HelloRepository(); // our repository
+    this.repository = new HelloRepository();
   }
   @get('/messages')
   @param.query.number('limit')
-  async list(limit = 10): Promise<HelloMessage[]>{ // returns a list of our objects
+  async list(limit = 10): Promise<HelloMessage[]>{
     // throw an error when the parameter is not a non-positive integer
-    if (limit < 1 || Math.floor(limit) !== limit)
+    if (!Number.isInteger(limit) || limit < 1)
       Promise.resolve(
         new HttpErrors.UnprocessableEntity('limit is non-positive'));
     else if (limit > 100)
--- a/pages/en/lb4/Controllers.md
+++ b/pages/en/lb4/Controllers.md
@@ -159,49 +159,62 @@ class HelloController {
 
 ## Handling Errors in Controllers
 
-In order to specify errors for controller methods to throw, the class `HttpErrors` is used. `HttpErrors` can be found in the `@loopback/rest` package.
+In order to specify errors for controller methods to throw, the class `HttpErrors` is used. `HttpErrors` is a class that has been re-exported from [http-errors](https://www.npmjs.com/package/http-errors), and can be found in the `@loopback/rest` package.
+
+Listed below are some of the most common error codes. The full list of supported codes is found [here](https://github.com/jshttp/http-errors#list-of-all-constructors).
+|Status Code|Error                |
+|-----------|---------------------|
+|400        |BadRequest           |
+|401        |Unauthorized         |
+|403        |Forbidden            |
+|404        |NotFound             |
+|500        |InternalServerError  |
+|502        |BadGateway           |
+|503        |ServiceUnavailable   |
+|504        |GatewayTimeout       |
 
 The example below shows how `HttpErrors` can be used inside a controller method and how it can be tested.
 
 ```js
 // the test
-import {HelloController} from 'path.to.controller'
-import {expect} from '@loopback/testlab'
-// ...
+import {HelloController} from 'path.to.controller';
+import {HttpErrors, expect} from '@loopback/testlab';
 
-// ...
 describe('Hello Controller', () => {
-  // ...
-  it('returns 412 Precondition Failed for non-positive limit', () => {
+  it('returns 422 Unprocessable Entity for non-positive limit', () => {
     const controller = new HelloController();
+    var errCaught: HttpErrors;
     try {
-      controller.list(0);
-      throw new Error('should have thrown an error');
+      controller.list(0.4); // an HttpError should be thrown here
     } catch (err) {
-      expect(err).to.have.property('statusCode', 412);
-      expect(err.message).to.match(/precondition failed/i);
+      errCaught = err;
     }
+    // the test fails here if the error was not thrown
+    expect(errCaught).to.have.property('statusCode', 422);
+    expect(errCaught.message).to.match(/unprocessable entity/i);
   })
-  // ...
 })
-// ...
 ```
 ```js
 // the controller
-import 'HttpErrors' from '@loopback/rest'
+import 'HttpErrors' from '@loopback/rest';
+import 'HelloRepostory' from 'path.to.repository';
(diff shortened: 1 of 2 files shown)


pages/en/lb4/Controllers.md
```diff
@@ -156,3 +156,52 @@ class HelloController {
 - A `Repository` is LoopBack's database abstraction. See [Repositories](Repositories.html) for more.
 - `@get('/messages')` creates the `Route` for the Operation using `app.route()`.
 - `@param.query.number` adds a `number` param with a source of `query`.
+
+## Handling Errors in Controllers
+
+In order to specify errors for controller methods to throw, the class `HttpErrors` is used. `HttpErrors` can be found in the `@loopback/rest` package.
```
kjdelisle: I'd mention that we're actually re-exporting the `http-errors` package: https://www.npmjs.com/package/http-errors

pages/en/lb4/Controllers.md
```diff
@@ -156,3 +156,52 @@ class HelloController {
...
+```js
+// the test
+import {HelloController} from 'path.to.controller'
+import {expect} from '@loopback/testlab'
+// ...
+
+// ...
+describe('Hello Controller', () => {
+  // ...
+  it('returns 412 Precondition Failed for non-positive limit', () => {
```
kjdelisle: ~Whoa! I don't think I've ever used 412 in my life!
...I like it.~

EDIT: Sorry, now that I looked it up, it's not meant to be used in this way.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/412):
> The HTTP 412 Precondition Failed client error response code indicates that access to the target resource has been denied. This happens with conditional requests on methods other than GET or HEAD when the condition defined by the If-Unmodified-Since or If-None-Match headers is not fulfilled. In that case, the request, usually an upload or a modification of a resource, cannot be made and this error response is sent back.

kjdelisle (approved):
One minor thing you should add, but I'll approve it now. Excellent example!

kjdelisle (changes requested):
I just had to Google it, didn't I?
I was curious about how code 412 worked and it's not actually intended to be used this way.

pages/en/lb4/Controllers.md
```diff
@@ -156,3 +156,52 @@ class HelloController {
...
+
+In order to specify errors for controller methods to throw, the class `HttpErrors` is used. `HttpErrors` can be found in the `@loopback/rest` package.
+
+The example below shows how `HttpErrors` can be used inside a controller method and how it can be tested.
+
+```js
+// the test
+import {HelloController} from 'path.to.controller'
+import {expect} from '@loopback/testlab'
+// ...
```
jannyHou: I think the `...` here is not a necessary, same to the other 4 `...`

pages/en/lb4/Controllers.md
kjdelisle: I'd change it to be a 422 Unprocessable Entity. The syntax is correct, but the _semantics_ of a request whose limit is a non-positive integer are incorrect, so it'd be the best match, IMO.

That being said, there's still debate about whether or not people should use 400 for that purpose, so either one is cool with me.

kjdelisle:
@shimks If you wanted to dive into the murky depths of "which code do I use", feel free to start here! https://stackoverflow.com/questions/16133923/400-vs-422-response-to-post-of-data

pages/en/lb4/Controllers.md
```diff
@@ -156,3 +156,52 @@ class HelloController {
...
+})
+// ...
+```
+```js
+// the controller
+import 'HttpErrors' from '@loopback/rest'
+
+class HelloController {
+  constructor() {
+    this.messages = new Repository('messages');
```
jannyHou: `this.repository` would be the correct property name.
Each controller has a property `repository`

pages/en/lb4/Controllers.md
b-admike: You can mention here if `HttpErrors` has an exhaustive list of http errors or a select ones which are common. Also, you can mention that it handles the status codes and messages for the errors you want, etc.

pages/en/lb4/Controllers.md
```diff
@@ -156,3 +156,52 @@ class HelloController {
...
+// ...
+
+// ...
+describe('Hello Controller', () => {
+  // ...
+  it('returns 412 Precondition Failed for non-positive limit', () => {
+    const controller = new HelloController();
+    try {
+      controller.list(0);
+      throw new Error('should have thrown an error');
```
b-admike: why are we throwing a new error here? shouldn't 
```
try { 
controller.list(0);
} catch (err) {
...
}
```
be enough?

pages/en/lb4/Controllers.md
```diff
@@ -156,3 +156,52 @@ class HelloController {
...
+// the controller
+import 'HttpErrors' from '@loopback/rest'
+
+class HelloController {
+  constructor() {
+    this.messages = new Repository('messages');
+  }
+  @get('/messages')
+  @param.query.number('limit')
+  list(limit = 10)
```
jannyHou: ~please specify the type of `limit` instead of assigning a value :) (or maybe I misunderstand the code here....)~

my bad, I do miss an important thing(define default value in ts), the code is correct and elegant here!

pages/en/lb4/Controllers.md
```diff
@@ -156,3 +156,52 @@ class HelloController {
...
+    if (limit < 1)
```
jannyHou: nitpick: I think in javascript `number` could be integer or float, so any number between 0~1 is still positive

> As in JavaScript, all numbers in TypeScript are floating point values.

https://www.typescriptlang.org/docs/handbook/basic-types.html

pages/en/lb4/Controllers.md
kjdelisle: Good catch! Sorry I missed that. Actually, it's even worse because the catch block will catch it instead.

pages/en/lb4/Controllers.md
kjdelisle: method should be async

jannyHou (commented):
nice doc! a few comments :)

pages/en/lb4/Controllers.md
```diff
@@ -156,3 +156,52 @@ class HelloController {
...
+      throw new HttpErrors.PreconditionFailed('limit is non-positive');
```
kjdelisle: function should return a rejected promise instead
`return Promise.resolve(new HttpErrors.UnprocessableEntity('limit is non-positive'));`

pages/en/lb4/Controllers.md
```diff
@@ -156,3 +156,52 @@ class HelloController {
...
+    else if (limit > 100)
+      limit = 100; // your logic
+    return this.messages.find({limit});
```
kjdelisle: `return await this.messages.find({limit});`

kjdelisle (changes requested):

pages/en/lb4/Controllers.md
kjdelisle: and it needs a bracket

pages/en/lb4/Controllers.md
```diff
@@ -156,3 +156,52 @@ class HelloController {
...
 - `@get('/messages')` creates the `Route` for the Operation using `app.route()`.
 - `@param.query.number` adds a `number` param with a source of `query`.
+
+## Handling Errors in Controllers
+
+In order to specify errors for controller methods to throw, the class `HttpErrors` is used. `HttpErrors` can be found in the `@loopback/rest` package.
+
+The example below shows how `HttpErrors` can be used inside a controller method and how it can be tested.
+
+```js
```
virkt25: I think most of the docs use `ts` not `js` ... switching to which would require you to add types for variables as per @jannyHou 's comment.

pages/en/lb4/Controllers.md
virkt25: +1 Would be helpful to have a list of all error's available (or atleast the common ones and a link to all the options in `http-errors` module).

pages/en/lb4/Controllers.md
shimks: what happens if controller.list(0) doesn't throw error as expected? Shouldn't expects() in the catch block not run and cause the test to pass when it shouldn't?

virkt25: Move `expect` statements to be outside of the try/catch block and you can assign `err` to another variable in the catch block so it's accessible outside. 

```
try {
  controller.list(0)
} catch (err) {
  goterr = err
}
expect(goterr).to.exist();
// ... 
```

pages/en/lb4/Controllers.md
jannyHou: @virkt25 yesterday I learned why we use `js` not `ts` lol... `.ts` doesn't recognize some keyword like `async`:

```ts
// a .ts file
async
```

```js
// a .js file
async
```

virkt25: :o interesting. I didn’t know that. Thanks for enlightening me @jannyHou

pages/en/lb4/Controllers.md
bajtos: I am personally fine with this approach. If we switch the order of the two `expect()` lines below and check the message first, then we will get a nice test failure message too - something along the lines "expected /precondition failed/" but got "should have thrown an error".

pages/en/lb4/Controllers.md
bajtos: Controllers should use dependency injection to receive a Repository instance. Let's simplify this example and make the controller "pure" with no dependencies.

bajtos (commented):
Could you please edit [Thinking in LoopBack] too and add a link referencing your new content? Here is the section in Thinking in LoopBack that is describing how to report errors from controller methods: http://loopback.io/doc/en/lb4/Thinking-in-LoopBack.html#handle-product-not-found-error

pages/en/lb4/Controllers.md
bajtos: ❗️ I am pretty certain this test will be confusing to many other readers too. Regardless of the  approach we decide to follow, there should be code comment(s) explaining why and how this test works, why it's important to be prepared for the case when the controller method does not throw any error, and how we handled that.

pages/en/lb4/Controllers.md
shimks: Could you clarify what you mean by that @bajtos ? Is it right for me to assume that I should continue to not use dependency injection here?

pages/en/lb4/Controllers.md
dhmlau: It would be better if we put some content instead of the `...`.  e.g. add your other imports here, or add your logics here.

pages/en/lb4/Controllers.md
```diff
@@ -156,3 +156,65 @@ class HelloController {
...
+// the test
+import {HelloController} from 'path.to.controller';
+import {HttpErrors, expect} from '@loopback/testlab';
+
+describe('Hello Controller', () => {
+  it('returns 422 Unprocessable Entity for non-positive limit', () => {
+    const controller = new HelloController();
+    var errCaught: HttpErrors;
+    try {
+      controller.list(0.4); // an HttpError should be thrown here
```
kjdelisle: `await controller.list(0.4);`

pages/en/lb4/Controllers.md
```diff
@@ -156,3 +156,65 @@ class HelloController {
...
+
+The example below shows how `HttpErrors` can be used inside a controller method and how it can be tested.
+
+```js
+// the test
+import {HelloController} from 'path.to.controller';
+import {HttpErrors, expect} from '@loopback/testlab';
+
+describe('Hello Controller', () => {
+  it('returns 422 Unprocessable Entity for non-positive limit', () => {
```
kjdelisle: `it('returns 422 Unprocessable Entity for non-positive limit', async () => {`

shimks:
I should mention, the code was based on http://loopback.io/doc/en/lb4/Controllers.html#writing-controller-methods. Since I've built on top of that section and made changes, should I also fix that section up a little?

kjdelisle (changes requested):
Couple more minor nits, but looking good!

pages/en/lb4/Controllers.md
```diff
@@ -156,3 +156,65 @@ class HelloController {
...
+
+class HelloController {
+  constructor() {
+    this.repository = new HelloRepository(); // our repository
+  }
+  @get('/messages')
+  @param.query.number('limit')
+  async list(limit = 10): Promise<HelloMessage[]>{ // returns a list of our objects
+    // throw an error when the parameter is not a non-positive integer
+    if (limit < 1 || Math.floor(limit) !== limit)
```
kjdelisle: `if (!Number.isInteger(limit) || limit  < 1)`

kjdelisle: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger

kjdelisle:
>I should mention, the code was based on http://loopback.io/doc/en/lb4/Controllers.html#writing-controller-methods. Since I've built on top of that section and made changes, should I also fix that section up a little?

Yeah, I'd put your cleaned-up code there in its place once you've got the green light for the rest of it.
(Omit the tests for that example, though)

crandmck (approved):

b-admike (approved):
LGTM once all the remaining comments are addressed. Great job @shimks 👍

jannyHou (approved):

kjdelisle (approved):

pages/en/lb4/Controllers.md
```diff
@@ -140,19 +140,85 @@ app.controller(MyController);
...
+
+```js
+// the test
+import {HelloController} from 'path.to.controller';
+import {HttpErrors, expect} from '@loopback/testlab';
+
+describe('Hello Controller', () => {
+  it('returns 422 Unprocessable Entity for non-positive limit', async () => {
+    const controller = new HelloController();
+    var errCaught: HttpErrors;
```
bajtos: please use `let` instead of `var`, see http://loopback.io/doc/en/contrib/style-guide.html#variable-declarations

pages/en/lb4/Controllers.md
```diff
@@ -140,19 +140,85 @@ app.controller(MyController);
...
+  constructor() {
+    this.repository = new HelloRepository();
+  }
+  @get('/messages')
+  @param.query.number('limit')
+  async list(limit = 10): Promise<HelloMessage[]>{
+    // throw an error when the parameter is not a non-positive integer
+    if (!Number.isInteger(limit) || limit < 1)
+      Promise.resolve(
+        new HttpErrors.UnprocessableEntity('limit is non-positive'));
```
bajtos: This part has two problems:
 - It calls `Promise.resolve()` instead of `Promise.reject()`, therefore the HTTP error won't be treated as an error, it won't be passed to `catch` handlers.
 - The Promise is not returned (missing `return` statement). As a result, this promise is ignored, and because it was created via `Promise.resolve`, no warning is printed.

Please simplify this part by using `throw`. The mechanism powering `async` function will convert the thrown error to a rejected promise automatically for you.

```ts
if (!Number.isInteger(limit) || limit < 1) {
  throw new HttpErrors.UnprocessableEntity('limit is non-positive'));
}
```

Also please copy your test and your implementation code into a small application and execute them to make sure both snippets work as expected!

bajtos (changes requested):

b-admike (commented):
Looks like you'll need to rebase your branch:
```
[workspace] $ /bin/sh /tmp/jenkins6528781950920441340.sh
**************************************************
**
**  Checking history of add-HttpErrors against gh-pages
**
**  PASS: add-HttpErrors merge-base and fork-point match, good job
**  FAIL: add-HttpErrors is behind gh-pages by 7 commits, rebase needed
**
**************************************************
Build step 'Execute shell' marked build as failure
```

bajtos (approved):
👍

PR #504, opened 2017-10-30
Fix urls on the getting started page

Juice10:
Fix the urls on the getting started page.
It shouldn't be #javascript-project but it should be #try-a-javascript-project otherwise the link doesn't go anywhere.
Super simple fix.


slnode:

raymondfeng (approved):

crandmck:
Thanks @Juice10

PR #505, opened 2017-10-30
Add ts-node as a loopback 4.0 prerequisite

Juice10:
I was going through the getting started and I bumped into ts-node being needed.
Getting started documentation mentions installing it locally but then the `ts-node` command doesn't get recognized by the shell.

### DIFF
--- a/pages/en/lb4/Installation.md
+++ b/pages/en/lb4/Installation.md
@@ -21,6 +21,13 @@ To get started with LoopBack 4, install the following:
 
    This installs the TypeScript compiler command, `tsc`.
 
+  ```
+  $ npm i -g ts-node
+  ```
+
+   This installs the TypeScript execution environment command for node, `ts-node`
+
+
 {% include tip.html content="
 You're installing `typescript` globally as a shortcut to get you started quickly,
 but it's not considered a best practice with Node.js.


slnode:

raymondfeng:
@Juice10 `ts-node` is not required for lb4. We set up `tsc` to transpile TypeScript into es2015/es2017 JHS code.

Juice10:
@raymondfeng then this needs to get changed in the getting started documentation as it is currently confusing there.

raymondfeng:
Good catch. We should fix Getting started to remove the usage of `ts-node`.

slnode:

crandmck:
We left it in there based on discussion in #493.  The procedure probably got broken due to diffs. btw calling `ts-node` directly and via `npm start`.     If we decide to remove it altogether that's fine, too, but I'll run thru it again and make sure it works for now.

crandmck:
@bajtos It appears that we need to pick one of the following scenarios for "Getting Started":
- Install both `typescript` and `ts-node` globally.
- Install `typescript` and `ts-node` as app dependencies and add `"ts-node index.ts"` for `scripts.start` in `package.json`.

Right now, we are half-and-half, which doesn't work; so I guess we need to pick one and stick with it.

crandmck:
I'm going to land this as a _temporary_ fix for the "Getting Started" procedure.  Currently, it also has you [install `ts-node` as project dependency](http://loopback.io/doc/en/lb4/Getting-started.html#install-project-dependencies), which is redundant; but at least it will work if someone follows the steps.  

But, as noted above, we need to go with one or the other.

bajtos:
@crandmck let's install both `ts-node` and `typescript` locally. Polluting global scripts is an anti-pattern - what if somebody has an existing front-end project that requires an older version of TypeScript? There is no way to globally install two versions of typescript (one for their legacy front-end and another for LoopBack Next).

PR #512, opened 2017-11-02
Add doc for Server[Under Technical Review...]

jannyHou:
Ready for a technical Review.

- Add Server concept
- Add doc of how to implement a Server
- The example contains an event listener, and after start the app, events are emitted periodically. Kind of mocking a server interacting with events from outside.

### DIFF
--- a/pages/en/lb4/Server.md
+++ b/pages/en/lb4/Server.md
@@ -10,10 +10,19 @@ summary:
 
 Server is a system that enters a working state of doing something by starting it, 
 and exits the working state by stopping it.
+Usually a Server is responsible for interacting with events outside, process incoming things
+and give corresponding responses.
 
 ## Implement a Server
 
-User can extend the context system to implement their own server and integrate it with your app.
+A Server is essencially a Context, see [Context](Context.htm) to learn more before
+you design a Server.
+
+User can extend the context system to implement their own Server and integrate it with their app.
+A Server module contains the following parts:
+
+- a Server Context class
+- a namespace for Context Binding
 
 ### Context Binding
 
@@ -22,8 +31,9 @@ Context, which inherits from the parent Application context. This way,
 any Server-specific bindings will remain local to the Server instance,
 and will avoid polluting its parent module scope.
 
-A server module can export its own Bindings as `namespace` and 
-receive app level config from `coreBindings`:
+A Server module can export its own Bindings as `namespace` and 
+receive app level config from `coreBindings`.
+Let's start with building a server that listen on some event:
 
 {% include code-caption.html content="myServer/src/keys.ts" %}
 ```js
@@ -31,33 +41,40 @@ import {CoreBindings} from '@loopback/core';
 
 export namespace MyServerBindings {
   export const CONFIG = `${CoreBindings.APPLICATION_CONFIG}#myserver`;
-  export const FOO = 'myserver.foo';
+  export const EMITTER_EVENT = 'myserver.emitter.event';
 }
 ```
 
-And the server class extends `Context` from `@loopback/context`.
+### Extend Context Class
+
+The Server class extends `Context` from `@loopback/context`.
 Its constructor takes in an `Application` instance as its parent `Context` and
-a configuration specific to a particular server instance.
+a configuration(optional) specific to a particular server instance.
 
 {% include code-caption.html content="myServer/src/MyServer.ts" %}
 ```js
 import {Context} from '@loopback/context';
 import {Server} from '@loopback/core';
 
-class MyServer extends Context implements Server {
+export class MyServer extends Context implements Server {
   constructor(
     @inject(CoreBindings.APPLICATION_INSTANCE) app: Application,
-    @inject(MyServerBindings.CONFIG) options?: RestServerConfig
+    @inject(MyServerBindings.CONFIG) options?: MyServerConfig
   ) {
     super(app);
   }  
+}
+
+export interface MyServerConfig {
+  x: string;
+  y: number;
 }
 ```
 
 ### Implement Server Functions
 
-You may already notice that the server declaration above also implements a `Server` interface.
-There are two basic functions a server must have:
+You may already notice that the Server declaration above also implements a `Server` interface.
+There are two basic functions a Server must have:
 
 - start()
 - stop()
@@ -69,42 +86,52 @@ and could interact with events from outside.
 
 {% include code-caption.html content="myServer/src/MyServer.ts" %}
 ```js
-import {Context} from '@loopback/context';
-import {Server} from '@loopback/core';
+import {Server, Application, CoreBindings} from '@loopback/core';
+import {Context, inject} from '@loopback/context';
 import * as EventEmitter from 'events';
 
 class MyEmitter extends EventEmitter {}
 
-class MyServer extends Context implements Server {
+export class MyServer extends Context implements Server {
   myEmitter: MyEmitter;
+
   constructor(
-    @inject(CoreBindings.APPLICATION_INSTANCE) app: Application,
-    @inject(MyServerBindings.CONFIG) options?: RestServerConfig
+      @inject(CoreBindings.APPLICATION_INSTANCE) public app: Application
   ) {
     super(app);
+    // an Emitter listen on a certain event whose name is returned by `getEventName()`
     this.myEmitter = new MyEmitter();
-  }  
+  }
+
+  // return the event name
+  async getEventName(): Promise<string> {
+    return await this.get('MyServerBindings.EMITTER_EVENT');
+  }
 
   async start(): Promise<void> {
-    this.myEmitter.on('event', () => {
-      console.log('an event occurred!');
+    console.log('Your server startted!');
+    let eventName = await this.getEventName();
+    // listen on an event and give some response
+    this.myEmitter.on(eventName, () => {
+      console.log('Your server is responding to an event...');
     });
     return Promise.resolve();
   }
 
   async stop(): Promise<void> {
-    this.myEmitter.removeListener('event', ()=>{
-      return Promise.resolve();
-    });
+    let eventName = await this.getEventName();
+    // remove listeners
+    this.myEmitter.removeAllListeners(eventName);
+    console.log('Your server stopped!');
+    return Promise.resolve();
   }
 }
 ```
 
-### Bind a Server
+### Bind a Server to Application
 
-Now you can create a server instance and bind it to your app. 
-When instanciating the `Application`, `start()` and `stop()` will be 
-mounted to `Application` level functions.
+Now you can bind a Server to your Application by calling `this.server()` 
+in the Application constructor.
 
 {% include code-caption.html content="application.ts" %}
 ```js
@@ -119,35 +146,48 @@ export class MyApp extends Application {
 }
 ```
 
-There are several ways to bind a server to an app
+When instanciating `MyApp`, a server instance named `serverFoo` will be 
+bound to the app, `start()` and `stop()` will be mounted to `Application` level functions.
+
+There are several ways to bind a server to an app, the strategy in the example above
+is the most staightforward one:
 
 ```js
-app.server(RestServer, 'nameOfYourServer');
+app.server(MyServer, 'nameOfYourServer');
 ```
 
-Or levarage a component to do it:
+Or you can levarage a component to do it:
 
 ```js
 const app = new MyApplication({
-  components: [RestComponent]
+  components: [MyComponent]
 });
 // OR
-app.component(RestComponent);
+app.component(MyComponent);
 ```
 
-Finally you can start your server/app by:
-(This part needs more clarification about one app with multiple server)
+Please see [Creating components](Creating-components.htm) and our implementation of 
+[RestComponent](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/rest-component.ts)
+for details.
+
+Finally you can create an app instance, start your app by `app.start()`,
+and stop it by `app.strop()`:
 
 {% include code-caption.html content="index.ts" %}
-```js
+```ts
 import { MyApp } from './src/application';
+import {MyServer} from './src/server/myServer/src/MyServer';
+
 const app = new MyApp();
+const eventName = 'working';
+app.bind('MyServerBindings.EMITTER_EVENT').to(eventName);
+
 app
   .start()
   .then(async () => {
-    const server = await app.getServer('jannyserverFoo');
-    // it doesn't work
-    server.myEmitter.emit('event');
+    // Or `const server = <MyServer> await app.getServer('serverFoo');`
+    const server = await app.getServer(MyServer);
+    setInterval(()=> {server.myEmitter.emit(eventName);}, 500);
   })
   .catch(err => {
     console.error(`Unable to start application: ${err}`);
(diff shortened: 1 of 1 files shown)


pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Server.html
+summary:
+---
+
+Server is a system that enters a working state of doing something by starting it, 
+and exits the working state by stopping it.
+Usually a Server is responsible for interacting with events outside, process incoming things
+and give corresponding responses.
+
```
shimks: I'm going to grammar nazi all over this, so please don't mind me.
This should be:
`Usually a Server is responsible for interacting with events outside, processing incoming things
and giving corresponding responses.`

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+export interface MyServerConfig {
+  x: string;
+  y: number;
+}
+```
+
+### Implement Server Functions
+
+You may already notice that the Server declaration above also implements a `Server` interface.
```
shimks: `You may have already noticed`

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+// OR
+app.component(MyComponent);
+```
+
+Please see [Creating components](Creating-components.htm) and our implementation of 
+[RestComponent](https://github.com/strongloop/loopback-next/blob/master/packages/rest/src/rest-component.ts)
+for details.
+
+Finally you can create an app instance, start your app by `app.start()`,
+and stop it by `app.strop()`:
```
shimks: `app.stop()`

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+
+process.on('SIGTERM', stopServer);
+process.on('SIGINT', stopServer);
+```
+
+*Please note, currently each app can only have one server.*
+
+## REST Server
+
+LoopBack implements a REST HTTP server for you to quickly get REST apis to work.
```
shimks: `REST APIs`

shimks: Sorry for polluting up the reviews, didn't know they'd be separated like this 😅

shimks (commented):
Makes sense to me, I think.

jannyHou:
@shimks applied, thanks!

shimks (approved):

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+title: 'Server'
+keywords: LoopBack 4.0, LoopBack 4
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Server.html
+summary:
+---
+
+Server is a system that enters a working state of doing something by starting it, 
+and exits the working state by stopping it.
```
kjdelisle, line 12: This is somewhat awkwardly phrased and implies the existence of concepts like "state", which we do not track or provide.

In LoopBack, a Server is anything that has an async start and stop function.

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+Usually a Server is responsible for interacting with events outside, processing incoming things
+and giving corresponding responses.
```
kjdelisle, line 14: A typical server will listen for requests on a port, handle those requests, and return appropriate responses.

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Implement a Server
+
+A Server is essencially a Context, see [Context](Context.htm) to learn more before
```
kjdelisle, line 18: This is in direct contradiction to what's written above. Servers don't require context. Technically, you can have a server that handles traffic from the outside without using any context-based functionality! Making your Server extend Context and accept a parent Context as an argument for its base constructor allows you to inject from bindings made on the parent (which is useful), but it's not required.

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+you design a Server.
+
+User can extend the context system to implement their own Server and integrate it with their app.
```
kjdelisle, line 21: `User can extend the context system` implies that you're expanding the functionality of Context. I know that `extends` is the keyword used in TypeScript, but this should be more explicit to avoid confusion.

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+A Server module contains the following parts:
+
+- a Server Context class
+- a namespace for Context Binding
```
kjdelisle, line 25: I would drop lines 22-25 as they're inaccurate and misleading. The section below does a much better job of explaining why you might want context, and also does a good job of saying that we _recommend_ it (instead of implying that it **does** contain context).

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Context Binding
+
+It is recommended that each Server implementation creates its own child
+Context, which inherits from the parent Application context. This way,
+any Server-specific bindings will remain local to the Server instance,
+and will avoid polluting its parent module scope.
+
+A Server module can export its own Bindings as `namespace` and
```
kjdelisle, line 34: Bindings are _not_ namespaced! Any bindings you create on a child context that share a key with a parent context's binding will replace that binding at the child level.

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Context Binding
+
+It is recommended that each Server implementation creates its own child
+Context, which inherits from the parent Application context. This way,
+any Server-specific bindings will remain local to the Server instance,
+and will avoid polluting its parent module scope.
+
+A Server module can export its own Bindings as `namespace` and 
+receive app level config from `coreBindings`.
```
kjdelisle, line 35: This should be `CoreBindings`. Currently, we're not exporting any documentation on those bindings, but we probably should given its usefulness.

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Server.html
+summary:
+---
+
+Server is a system that enters a working state of doing something by starting it, 
+and exits the working state by stopping it.
+Usually a Server is responsible for interacting with events outside, processing incoming things
+and giving corresponding responses.
+
```
raymondfeng, line 15: This paragraph is too generic. It would be better to describe `Server` in the context of LoopBack 4. We introduce `Server` to represent inbound transports and/or protocols. For example:

- REST over http/https 
- gRPC over http2
- GraphQL over http/https
- jsonrpc over http/https
- MQTT

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+### Context Binding
+
+It is recommended that each Server implementation creates its own child
+Context, which inherits from the parent Application context. This way,
+any Server-specific bindings will remain local to the Server instance,
+and will avoid polluting its parent module scope.
+
+A Server module can export its own Bindings as `namespace` and 
+receive app level config from `coreBindings`.
+Let's start with building a server that listen on some event:
```
kjdelisle, line 36: You're saying that we'll start with building a server, but the code snippet is just an extension of CoreBindings with our own bindings (which we don't want to do!)

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+
+export namespace MyServerBindings {
+  export const CONFIG = `${CoreBindings.APPLICATION_CONFIG}#myserver`;
+  export const EMITTER_EVENT = 'myserver.emitter.event';
+}
+```
+
+### Extend Context Class
+
+The Server class extends `Context` from `@loopback/context`.
```
kjdelisle, line 50: "Our" server class. I would make up an example server class and make it clear that it's one we've invented for demonstrative purposes. The repeated assertions about what `Server`/Server does, contains, etc are incredibly confusing.

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+
+{% include code-caption.html content="application.ts" %}
+```js
+import { Application, ApplicationConfig } from '@loopback/core';
+import {MyServer} from './myServer/src/MyServer';
+
+export class MyApp extends Application {
+  constructor(options?: ApplicationConfig) {
+  super(options);
+  this.server(MyServer, 'serverFoo');
```
kjdelisle, line 144: Change the name of the instance. All instances bound in this way will have a key `servers.{name}` where `{name}` is the string provided. This would give you a key of `servers.serverFoo` which is unnecessary.

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,220 @@
...
+```
+
+Or you can levarage a component to do it:
+
+```js
+const app = new MyApplication({
+  components: [MyComponent]
+});
+// OR
+app.component(MyComponent);
```
kjdelisle, line 166: Components _can_ setup servers, but that's only because components can do literally _anything_ the developer writes them to do!

I would avoid mentioning this, especially given that there's a good chance we'll remove the automatic instantiation of a RestServer from RestComponent (and possibly do away with RestComponent entirely!)

Hidden instantiation of servers behind components isn't necessarily an anti-pattern, but it doesn't do enough to let you see what's happening and could lead to unforeseen consequences.

kjdelisle (changes requested):
The general concept of what a Server _is_ should be done as a separate document, covered by [this task](https://github.com/strongloop/loopback-next/issues/680), while this document should describe what you want to do to make your own implementation.

This document doesn't have a logical flow to follow, and while the example of a custom Server is a good start, it doesn't do enough to cover the fundamentals of how to make a useful server implementation.

There was no mention of the core concept of a router in this Pull Request, which we discussed yesterday as being the only "best practice" we need to mention (since people will most likely use controllers, we need a router to work with that concept).

Overall, what I'd like to see is something broken down along these lines:
- Making a basic server (BasicServer)
- Making a server with connectivity (AdvancedServer)
- Adding routing to your AdvancedServer

Additional notes:
- do not say that Server _is_ a Context. Your implementations of Server _can extend_ Context, but nothing requires it.
- do not mention "state". Servers do not require state tracking, or possess it as an innate quality, and this will only serve to confuse others.

pages/en/lb4/Server.md
kjdelisle: @raymondfeng Part of why there's an explanation regarding Servers as a simpler concept here is that there's nothing to stop a user from creating worker-like implementations that follow this interface. 

As an example, creating processors that convert documents from one format to another in the background do not need to run their own HTTP/S servers (though they'll likely connect to a datasource of some kind in the background).

I asked @jannyHou to cover off this concept since leveraging it gives you two things for free:
- automatic starting/stopping of those instances along with any other Servers
- automatic naming of bound instances for later use

That being said, the general concept of what a Server is shouldn't be in this document.

pages/en/lb4/Server.md
kjdelisle: Quick note: I attempted to document these bindings in the TypeScript, but our doc generation doesn't pick up any of the annotations (so I'm not sure what to do about that right now).

bajtos:
@kjdelisle great comments. I think you know much more about Servers than I do, I am happy to let you drive this one.

> do not say that Server is a Context. Your implementations of Server can extend Context, but nothing requires it.

Slightly related: I think it's important to explicitly say that the routing implementation _must_ create a new child context for each request handled, and explain why is this important (each request has different request-specific values bound to the context).

A per-server context is optional, a per-request context is required.

bajtos:
Few more comments:

 - Could you please update the pull request title to make it more clear that this is documentation for people writing servers, not for consumers (application developers)?
 - Please add the new page to the sidebar. I think it should be somehow related or close to http://loopback.io/doc/en/lb4/Extending-LoopBack-4.html and http://loopback.io/doc/en/lb4/Creating-components.html.

pages/en/lb4/Server.md
jannyHou: @raymondfeng @kjdelisle @bajtos Shall we agree on the concept of `Server` first?

jannyHou:
> The general concept of what a Server is should be done as a separate document, covered by this task, 

I am not sure why the same concept's "definition" and "implementation" should be documented separately.

> while this document should describe what you want to do to make your own implementation.

The whole things I wrote is my understanding of what I need to do to make my own implementation.

> This document doesn't have a logical flow to follow,

Have you really carefully read what I wrote? There is a flow: each sub title means a step, and I document them in the same order as how you built the timer app.

>  and while the example of a custom Server is a good start, it doesn't do enough to cover the fundamentals of how to make a useful server implementation.

after making a Server's concept very generic as a worker, I don't know what are the other "useful" things I need to include in the implementation. "fundamentals of how to make a useful server implementation" I don't know what do you mean by "fundamentals"  here.

> There was no mention of the core concept of a router in this Pull Request, which we discussed yesterday as being the only "best practice" we need to mention (since people will most likely use controllers, we need a router to work with that concept).

"route" is not a must-have concept in a generic Server, if it is a must-have concept of some Server , I would suggest someone put down the definition of the "some Server" and also put down the elements to build in a best practice.

From my understanding, "some Server" is a **Web Server**, and the elements are Sequence, Router. 

And moreover, I think at the end of the meeting we don't have an agreement of the best practise due to the complexity of sample app's implementation.

> Making a server with connectivity (AdvancedServer)
Adding routing to your AdvancedServer

Same concern as above. 

> do not say that Server is a Context. Your implementations of Server can extend Context, but nothing requires it.

OK fine, this is the most fundamental thing I can see to start your Server impt with, if it's not, then I really have no idea how are you going to define your Server.

> do not mention "state". Servers do not require state tracking, or possess it as an innate quality, and this will only serve to confuse others.

Again, this is my understanding, and if that's not correct, someone please give a clear definition of Server.

pages/en/lb4/Server.md
jannyHou: > In LoopBack, a Server is anything that has an async start and stop function.

Are you saying Server is just a javascript/typescript interface? Then what is the point of documenting a js/ts interface? From you description I don't see any relation between the interface and Loopback 4

pages/en/lb4/Server.md
jannyHou: > A typical server will listen for requests on a port, handle those requests, and return appropriate responses.

"A typical server"..... do you know how huge the gap is between the definition of "anything that has an async start and stop function." and " listen for requests on a port, handle those requests, and return appropriate responses"?

IMO, I cannot explain the thing like "Human is a creature that has body and head and can breath. A typical human does coding everyday". That is a typical programmer, not a typical human

pages/en/lb4/Server.md
jannyHou: I explained in https://github.com/strongloop/loopback.io/pull/512#issuecomment-341739216

pages/en/lb4/Server.md
jannyHou: ok

pages/en/lb4/Server.md
jannyHou: ok I will dig more, I don't find any document of Binding and was not confident when wrote this line. thanks for correcting.

pages/en/lb4/Server.md
jannyHou: ok I can change it to "Let's start to build a server..." or just move it up to "##Implement Server" section.

pages/en/lb4/Server.md
jannyHou: ok

jannyHou:
I am closing this PR since most of its content doesn't seem on the right track.

PR #513, opened 2017-11-03
lb4: Application

kjdelisle:
connected to https://github.com/strongloop/loopback-next/issues/650

Document the Application class as a concept, along with its purpose and usage.

### DIFF
--- a/pages/en/lb4/Application.md
+++ b/pages/en/lb4/Application.md
@@ -1,0 +1,211 @@
+---
+lang: en
+title: 'Application'
+keywords: LoopBack 4.0, LoopBack 4
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Application.html
+summary:
+---
+
+## What is an Application?
+
+In LoopBack 4, the [`Application`](http://apidocs.strongloop.com/@loopback%2fcore/#Application)
+class is the central class for setting up all of your module's components,
+controllers, servers and bindings. The `Application` class extends 
+[Context](Context.html), and provides the controls for starting and stopping
+itself and its associated servers.
+
+When using LoopBack 4, we strongly encourage you to create your own subclass
+of `Application` to better organize your configuration and setup.
+
+## Making your own application class
+
+By making your own application class, you can perform several additional
+tasks as a part of your setup:
+- Pass configuration into the base class constructor
+- Perform some asynchronous wireup before application start
+- Perform some graceful cleanup on application stop 
+
+{% include code-caption.html content="src/widget-application.ts" %}
+```ts
+import {Application} from '@loopback/core';
+import {RestComponent, RestServer} from '@loopback/rest';
+import {SamoflangeController, DoohickeyController} from './src/controllers';
+import {WidgetApi} from './apidef/';
+
+export class WidgetApplication extends Application {
+  constructor() {
+    // This is where you would pass configuration to the base constructor
+    // (as well as handle your own!)
+    super({
+      components: [RestComponent],
+    });
+    const app = this; // For clarity.
+    // You can bind to the Application-level context here.
+    // app.bind('foo').to(bar);
+    app.controller(SamoflangeController);
+    app.controller(DoohickeyController);
+  }
+
+  async start() {
+    // This is where you would asynchronously retrieve servers, providers and
+    // other components to configure them before launch.
+    const server = await app.getServer(RestServer);
+    server.bind('rest.port').to(8080);
+    server.api(WidgetApi);
+    return await super.start();
+  }
+
+  async stop() {
+    // This is where you would do whatever is necessary before stopping your
+    // app (graceful closing of connections, flushing buffers, etc)
+    console.log('Widget application is shutting down...')
+    await super.stop();
+  }
+}
+
+```
+
+## Configuring your application
+Your application can be configured with constructor arguments, bindings, or
+a combination of both.
+
+### Binding configuration
+Binding is the most commonly-demonstrated form of application configuration
+throughout our examples, and in most cases, is the recommended method for
+setting up your application.
+
+In addition to the binding functions provided by [Context](Context.html),
+the `Application` class also provides some sugar functions for commonly used
+bindings, like `component`, `server` and `controller`:
+
+```ts
+export class MyApplication extends Application {
+  constructor() {
+    super();
+  }
+  this.component(MagicSuite);
+  this.server(RestServer, 'public');
+  this.server(RestServer, 'private');
+
+  this.controller(FooController);
+  this.controller(BarController);
+  this.controller(BazController);
+}
+```
+
+You can find a complete list of these functions on the
+[`Application`](http://apidocs.loopback.io/@loopback%2fcore/#Application) API
+docs page.
+
+Additionally, you can use more advanced forms of binding to fine-tune your
+application's configuration:
+
+```ts
+export class MyApplication extends Application {
+  constructor() {
+    super();
+  }
+  this.server(RestServer);
+  this.controller(FooController);
+  this.bind('fooCorp.logger').toProvider(LogProvider);
+  this.bind('repositories.widget')
+    .toClass(WidgetRepository)
+    .inScope(BindingScope.SINGLETON);
+}
+```
+In the above example:
+- injection calls for `fooCorp.logger` will be handled by the `LogProvider`
+  class.
+- injection calls for `repositories.widget` will be handled by a singleton
+instance of the `WidgetRepository` class.
+
+### Constructor configuration
+
+The `Application` class constructor also accepts an
+[`ApplicationConfig`](http://apidocs.strongloop.com/@loopback%2fcore/#ApplicationConfig)
+object which contains three collections: `components`, `controllers` and `servers`.
+It will automatically create bindings for each of these collections as a part
+of defining the application instance.
+
+{% include note.html content="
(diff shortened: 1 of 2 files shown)


pages/en/lb4/Application.md
```diff
@@ -0,0 +1,223 @@
...
+
+## What is an Application?
+
+In LoopBack 4, the [`Application`](http://apidocs.strongloop.com/@loopback%2fcore/#Application)
+class is the central class for setting up all of your module's components,
+controllers, servers and bindings. The `Application` class extends 
+[Context](Context.html), and provides the controls for starting and stopping
+itself and its associated servers.
+
+When using LoopBack 4, we strongly encourage you to create your own child class
```
raymondfeng: `subclass` instead of `child class`?

pages/en/lb4/Application.md
```diff
@@ -0,0 +1,223 @@
...
+of `Application` to better organize your configuration and setup.
+
+## Making your own application
+
+By making your own application instance, you can perform several additional
```
raymondfeng: `own application class`?

pages/en/lb4/Application.md
```diff
@@ -0,0 +1,223 @@
...
+" %}
+
+Many bindings defined by the framework itself are utilized under the
+assumption that there will only be one instance of an `Application` class.
+
+### Use caution with singleton binding scopes
+By default, bindings for controllers will instantiate a new instance whenever
+they are injected or retrieved from their binding. Your application should only
+set singleton binding scopes on controllers when it makes sense to do so.
+
```
raymondfeng: I don't think it's true for LoopBack 4. Most of internal services should NOT be assumed as singletons. When we bind a service class to a given context, new instances will be created. Even BindingScope.SINGLETON means it's a singleton within the context subtree.

raymondfeng (changes requested):
Please address my comments.

pages/en/lb4/Application.md
kjdelisle: Umm, that sentence says "don't make your bindings singletons unless it makes sense to do it". I think we're in agreement?

raymondfeng: Weird. My comment was against `+### Do not create multiple Application instances` paragraph.

kjdelisle: If we're talking about that section, the issue is that decorators like `@api`, are creating metadata entries using `Reflector.defineMetadata`, which is our singleton instance of our NamedspacedReflect. If two instances of Application are both importing that same shared reference, will they not be polluting each other's "global" binding scope?

https://github.com/strongloop/loopback-next/blob/master/packages/context/src/reflect.ts#L182

kjdelisle: Edited previous comment to correct some definitions and names, and to add a link to what I'm referring to.

raymondfeng: Decorators such as `@api` add metadata to a class/property/method/param. These metadata should be independent of the hosting app. Multiple app instances should be able to share the same metadata. App instance specific information need to be captured somewhere else outside the decorations. 

BTW, you should be able to create app1 and app2 to partition controllers/repositories. This scheme might be more useful for platforms such as appconnect that builds on top of LoopBack. For applications, one `app` is the case for most of the time, especially in the spirit of microservices.

kjdelisle: You can bind controllers to server-level context to achieve this.

My concern is that partitioning may occur in areas like controllers, but devs might be tempted to recycle common definitions of things like Models, which will definitely run afoul of this limitation since they also make use of the `Reflector` instance: https://github.com/strongloop/loopback-next/blob/8c7f474a72df23a1fa2d33bee9d36ba41031fd63/packages/repository/src/decorators/model.ts#L30

kjdelisle: I think the nuance of documenting this minefield is only going to open devs to problems down the road, and I don't see the value gained in doing so when compared to just separating them into different microservices.

pages/en/lb4/Application.md
```diff
@@ -0,0 +1,223 @@
...
+import {WidgetApi} from './apidef/';
+
+export class WidgetApplication extends Application {
+  constructor() {
+    // This is where you would pass configuration to the base constructor
+    // (as well as handle your own!)
+    super({
+      components: [RestComponent],
+    });
+    // You can bind to the Application-level context here.
```
b-admike: Can we show them an example of how? Even if it is commented, something like:
```
//app.bind('foo').to('bar');
```

pages/en/lb4/Application.md
```diff
@@ -0,0 +1,223 @@
...
+
+The `Application` class constructor also accepts an
+[`ApplicationConfig`](http://apidocs.strongloop.com/@loopback%2fcore/#ApplicationConfig)
+object which contains three collections: `components`, `controllers` and `servers`.
+It will automatically create bindings for each of these collections as a part
+of defining the application instance.
+
+{% include note.html content="
+  More advanced binding configuration such as provider binding, or binding scopes
+  are not possible with the constructor-based configuration approach.
```
b-admike: In this configuration method, should we mention what the default binding scopes would be?

pages/en/lb4/Application.md
```diff
@@ -0,0 +1,223 @@
...
+
+
+
+## Tips for application setup
+Here are some tips to help avoid common pitfalls and mistakes.
+
+### Use unique bindings
+Use binding names that are prefixed with a unique string that does not overlap
+with loopback's bindings. As an example, if your application is built for
+your employer FooCorp, you can prefix your binding with `fooCorp`.
```
b-admike: `prefix your bindings...`

pages/en/lb4/Application.md
```diff
@@ -0,0 +1,223 @@
...
+    'public': RestServer,
+    'private': RestServer,
+  },
+});
+```
+In the above example, the two server instances would be bound to the Application
+context under the keys `servers.public`, and `servers.private` respectively.
+
+
+
```
virkt25: Too much whitespace here (3 lines).

crandmck (approved):
I'll copy edit after this lands.

pages/en/lb4/Application.md
raymondfeng: My point is that we don't have to take a position in this doc. IMO, it would be better to remove this paragraph.

kjdelisle:
@raymondfeng Please take another look. :)

b-admike (approved):

pages/en/lb4/Application.md
```diff
@@ -0,0 +1,221 @@
...
+We provide the [`getSync`](http://apidocs.loopback.io/@loopback%2fcontext/#getSync)
+function for scenarios where you cannot asynchronously retrieve your bindings,
+such as in constructor bodies.
+
+However, the number of scenarios in which you must do this are limited, and you
+should avoid potential race conditions and retrieve your bindings asynchronously
+using the [`get`](http://apidocs.loopback.io/@loopback%2fcontext/#get) function
+whenever possible.
+
+### Do not create multiple Application instances
```
raymondfeng: I suggest that we remove this paragraph based on my previous comments.

raymondfeng (approved):

pages/en/lb4/Application.md
```diff
@@ -0,0 +1,211 @@
...
+{% include code-caption.html content="src/widget-application.ts" %}
+```ts
+import {Application} from '@loopback/core';
+import {RestComponent, RestServer} from '@loopback/rest';
+import {SamoflangeController, DoohickeyController} from './src/controllers';
+import {WidgetApi} from './apidef/';
+
+export class WidgetApplication extends Application {
+  constructor() {
+    // This is where you would pass configuration to the base constructor
```
jannyHou, line 39: nitpick: I am a little confused what kind of "configuration" to pass here, would like to see a one-line example, like [`app.bind('foo').to(bar)`](https://github.com/strongloop/loopback.io/pull/513/files#diff-21ab553137dc3efeb2a9c3b17acd308cR46)

pages/en/lb4/Application.md
```diff
@@ -0,0 +1,211 @@
...
+    app.controller(DoohickeyController);
+  }
+
+  async start() {
+    // This is where you would asynchronously retrieve servers, providers and
+    // other components to configure them before launch.
+    const server = await app.getServer(RestServer);
+    server.bind('rest.port').to(8080);
+    server.api(WidgetApi);
+    return await super.start();
```
bajtos, line 57: Perhaps add a comment why is it important to call `super.start()`? (In order to let the base method to actually start the servers.)

Similar comment applies to `await super.stop()` below.

pages/en/lb4/Application.md
```diff
@@ -0,0 +1,211 @@
...
+In addition to the binding functions provided by [Context](Context.html),
+the `Application` class also provides some sugar functions for commonly used
+bindings, like `component`, `server` and `controller`:
+
+```ts
+export class MyApplication extends Application {
+  constructor() {
+    super();
+  }
+  this.component(MagicSuite);
```
bajtos, line 88: This is outside of `constructor` function, inside class declaration - looks like a typo to me? Please check that the snippet compiles.

pages/en/lb4/Application.md
kjdelisle: The very next line is demonstrating the passthrough of a config object to the super call (in this case, the `RestComponent`.

The individual binding examples are not what this comment is describing.

PR #521, opened 2017-11-08
rework key concepts

shimks:
- Added summary on Components with links to Using Components and Creating Components.
- Reordered key concepts from:
  - sequence->routes->schema->controllers->context->dependency injection->repositories->decorators
  - to
  - ~routes->controllers->sequence->decorators->context->dependency injection->schema->repositories~
  - application->server->context->dependency injection->controllers->routes->sequences->servers->schema->repositories->decorators
- connected to https://github.com/strongloop/loopback-next/issues/636

### DIFF
--- a/_data/sidebars/lb4_sidebar.yml
+++ b/_data/sidebars/lb4_sidebar.yml
@@ -40,28 +40,28 @@ children:
     url: Server.html
     output: 'web, pdf'
 
-  - title: 'Sequence'
-    url: Sequence.html
-    output: 'web, pdf'
-
-  - title: 'Routes'
-    url: Routes.html
-    output: 'web, pdf'
-
-  - title: 'Schemas'
-    url: Schemas.html
-    output: 'web, pdf'
-
-  - title: 'Controllers'
-    url: Controllers.html
-    output: 'web, pdf'
-
   - title: 'Context'
     url: Context.html
     output: 'web, pdf'
 
   - title: 'Dependency injection'
     url: Dependency-injection.html
+    output: 'web, pdf'
+
+  - title: 'Controllers'
+    url: Controllers.html
+    output: 'web, pdf'
+
+  - title: 'Routes'
+    url: Routes.html
+    output: 'web, pdf'
+
+  - title: 'Sequence'
+    url: Sequence.html
+    output: 'web, pdf'
+
+  - title: 'Schemas'
+    url: Schemas.html
     output: 'web, pdf'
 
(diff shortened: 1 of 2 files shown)


b-admike (approved):
I don't really have a strong opinion on the ordering of the concepts.

raymondfeng (approved):

crandmck (approved):
LGTM

jannyHou:
I also don't have very strong opinion on their sequence, just a suggestion(feel free to ignore):

We are missing two important concepts "Server" and "Application", after those two docs check in, I would suggest list concepts as:

Dependency Injection --> Context -->  Application -->  --> Controller --> Server --> Route --> Sequence --> Schema --> Repositories  --> Decorators 

- Dependency Injection: A very important design concept people have to learn first to understand examples containing `@inject()` afterwards.

- [Context --> Application] : I feel Context is the most fundamental concept that users can learn first. Then introduce Application(which extends Context and is the highest level Context in your app.)

- [Server --> Route --> Sequence]: I would suggest put these three concepts adjacent since at least our rest server is implemented in a way that leverages Route and Sequence.

- [Schema --> Repositories]: Would be better to organize these 4 concepts together: Schema, Model, DataSource, Repositories.

- [Decorators]: It is more like a Language concept we apply in LB4, but since we explain it by listing our usage of it, readers will have difficulty understand those sections if they are not familiar with Application, Route, Repositories, etc....

crandmck:
> [Decorators]: It is more like a Language concept we apply in LB4, but since we explain it by listing our usage of it, readers will have difficulty understand those sections if they are not familiar with Application, Route, Repositories, etc....

That reminds me: I think Decorators would go better under [Language-related concepts](http://loopback.io/doc/en/lb4/Language-related-concepts.html) where we already have Mixin.

If these concepts are required to understand everything else, then perhaps we could move "Language-related concepts" up in the sidebar (perhaps between "Getting started" and "Key concepts").

Also, wouldn't dependency injection technically be a "language concept," too?

kjdelisle:
@shimks I think once @b-admike 's PR land, you're going to need to rebase and move his new file to the correct location. Once that's done, LGTM

kjdelisle:
Ah! Almost forgot! My PR for `Application` landed as well, which will also cause you headaches. Sorry. :(

b-admike:
@shimks The server concept PR has landed as well!

jannyHou (approved):

b-admike (approved):

PR #522, opened 2017-11-08
Add server concept page

b-admike:
Closes https://github.com/strongloop/loopback-next/issues/680

### DIFF
--- a/pages/en/lb4/Concepts.md
+++ b/pages/en/lb4/Concepts.md
@@ -23,6 +23,7 @@ servers.
 - [**Dependency Injection**](Dependency Injection.html): Technique that separates the construction of dependencies of a class or function from its behavior, to keep the code loosely coupled.
 - [**Repository**](Repositories.html): Type of Service that represents a collection of data within a DataSource.
 - [**Decorator**](Decorators.html): Enables you to annotate or modify your class declarations and members with metadata.
+- [**Server**](Server.html): Represents implementation for inbound transports and/or protocols such as REST over http, gRPC over http2, and graphQL over https. It typically listens for requests on a specific port, handle them, and return appropriate responses.
 
 {% include note.html title="Review Note" content="_Perhaps this should include some of the material in <a href='Thinking-in-LoopBack.html'> Thinking in LoopBack</a>_.
 " %}
--- a/pages/en/lb4/Server.md
+++ b/pages/en/lb4/Server.md
@@ -1,0 +1,76 @@
+---
+lang: en
+title: 'Server'
+keywords: LoopBack 4.0, LoopBack 4
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Server.html
+summary:
+---
+
+## Overview
+
+The [Server](https://apidocs.strongloop.com/@loopback%2fcore/#Server) interface defines the minimal required functions (start and stop) to implement for a LoopBack application. Servers in LoopBack 4 are used to represent implementations for inbound transports and/or protocols such as REST over http, gRPC over http2, graphQL over https, etc. They typically listen for requests on a specific port, handle them, and return appropriate responses. A single application can have multiple server instances listening on different ports and working with different protocols.
+
+
+## Usage
+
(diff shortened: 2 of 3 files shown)


pages/en/lb4/Concepts.md
```diff
@@ -18,6 +18,7 @@ LoopBack 4 introduces some new concepts that are important to understand:
 - [**Dependency Injection**](Dependency Injection.html): Technique that separates the construction of dependencies of a class or function from its behavior, to keep the code loosely coupled.
 - [**Repository**](Repositories.html): Type of Service that represents a collection of data within a DataSource.
 - [**Decorator**](Decorators.html): Enables you to annotate or modify your class declarations and members with metadata.
+- [**Server**](Server.html): Placeholder.
```
crandmck: Suggest something like this (adapted from your page below).

Represents implementation for inbound transports and/or protocols such as REST over http, gRPC over http2, graphQL over https. They typically listen for requests on a specific port, handle them, and return appropriate responses. A single application can have multiple server instances listening on different ports and working with different protocols.

crandmck (changes requested):

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,40 @@
...
+keywords: LoopBack 4.0, LoopBack 4
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Server.html
+summary:
+---
+
+## Overview
+
+The [Server](https://apidocs.strongloop.com/@loopback%2fcore/#Server) interface defines the minimal required functions (start and stop) to implement for a LoopBack application. Servers in LoopBack 4 are used to represent implementations for inbound transports and/or protocols such as REST over http, gRPC over http2, graphQL over https, etc. They typically listen for requests on a specific port, handle them, and return appropriate responses. A single application can have multiple server instances listening on different ports and working with different protocols.
```
jannyHou, line 13: Based on all discussions I have participated, I think

> Servers in LoopBack 4 are used to represent implementations for inbound transports and/or protocols such as REST over http, gRPC over http2, graphQL over https, etc. They typically listen for requests on a specific port, handle them, and return appropriate responses.

is a proper definition with no contradiction. 

> A single application can have multiple server instances listening on different ports and working with different protocols.

IIRC in the current release we decide to avoid dealing with multiple servers bound to one app, focus on the use case "one app only has one server". Better confirm with @kjdelisle about it....

jannyHou (commented):
@b-admike Thank you so much for taking over the concept part. Just a few comment.

bajtos (commented):
Good stuff! I don't see any obvious problems, I'll leave it up to others to review the details.

shimks (approved):

pages/en/lb4/Concepts.md
kjdelisle: Excellent suggestion! :100:

pages/en/lb4/Server.md
kjdelisle: The current implementation allows multiple "servers"; it's the RestServer's implementation of things like `@api` and other tags that don't quite fit with the current implementation (storing their values in the top-level metadata container is why this is an issue)

kjdelisle (changes requested):
It's almost done!

Things to add:
- How to add servers to your app: should be a small snippet or two showing how to use `app.server(...)` (I know you do that in the large example, but we should highlight it so people don't overlook it)
- Link back to Creating-Components.html so that devs can learn about how to make their own servers.

pages/en/lb4/Server.md
b-admike: Thanks @jannyHou @kjdelisle. I'll keep it as is, but if you feel strongly about removing it since we don't have concrete examples for it, I don't mind doing so.

crandmck (approved):

jannyHou (approved):

pages/en/lb4/Server.md
```diff
@@ -0,0 +1,76 @@
...
+  }
+  // This server instance will be bound under "servers.fooServer".
+  this.server(RestServer, 'fooServer');
+  // Creates a binding for "servers.MQTTServer" and a binding for
+  // "servers.SOAPServer";
+  this.servers([MQTTServer, SOAPServer]);
+}
+```
+
+You can also add multiple servers in the constructor of your application class as shown [Here](Application.html#servers).
```
kjdelisle: change the case of `Here` to `here` since it's in the middle of a sentence.

kjdelisle (dismissed):
One minor nit, then LGTM

PR #523, opened 2017-11-09
docs(lb4): Creating components - custom servers

kjdelisle:
connected to https://github.com/strongloop/loopback-next/issues/681

### DIFF
--- a/pages/en/lb4/Creating-components.md
+++ b/pages/en/lb4/Creating-components.md
@@ -25,15 +25,15 @@ export class MyComponent {
 }
 ```
 
-When a component is mounted to an application, a new instance of the component class is created and then
- - each Controller class is registered via `app.controller()`,
- - each Provider is bound to its key in `providers` object.
+When a component is mounted to an application, a new instance of the component class is created and then:
+ - Each Controller class is registered via `app.controller()`,
+ - Each Provider is bound to its key in `providers` object.
 
 The example `MyComponent` above will add `MyController` to application's API and create a new binding `my.value` that will be resolved using `MyValueProvider`.
 
 ## Providers
 
-Providers are the mechanism allowing components to export values that can be used by the target application or other components. A provider is a class that provides `value()` function, this function is called by [Context](Context.html) whenever another entity requests a value to be injected.
+Providers enable components to export values that can be used by the target application or other components. The `Provider`  class provides a `value()` function called by [Context](Context.html) when another entity requests a value to be injected.
 
 ```js
 import {Provider} from '@loopback/context';
@@ -63,7 +63,7 @@ export class MyComponent {
 
 ### Accessing values from Providers
 
-Applications should use `@inject` decorators to access the value provided by our exported provider.
+Applications can use `@inject` decorators to access the value of an exported Provider.
 If you’re not familiar with decorators in TypeScript, see [Key Concepts: Decorators](Decorators.html)
 
 ```js
@@ -84,7 +84,8 @@ class MyController {
 }
 ```
 
-A note on binding names: In order to avoid name conflicts, always add a unique prefix to your binding key (`my-component.` in the example above). See [Reserved binding keys](Reserved-binding-keys.html) for the list of keys reserved for the framework use.
+{% include note.html title="A note on binding names" content="To avoid name conflicts, add a unique prefix to your binding key (for example, `my-component.` in the example above). See [Reserved binding keys](Reserved-binding-keys.html) for the list of keys reserved for the framework use.
+" %}
 
 ### Asynchronous providers
 
@@ -107,7 +108,7 @@ In this case, LoopBack will wait until the promise returned by `value()` is reso
 
 ### Working with HTTP request/response
 
-In some cases, the provider may depend on other parts of LoopBack, for example the current `request` object. Such dependencies should be listed in provider's constructor and annotated with `@inject` keyword, so that LoopBack runtime can resolve them automatically for you.
+In some cases, the Provider may depend on other parts of LoopBack; for example the current `request` object. The Provider's constructor should list such dependencies annotated with `@inject` keyword, so that LoopBack runtime can resolve them automatically.
 
 ```js
 const uuid = require('uuid/v4');
@@ -125,7 +126,7 @@ class CorrelationIdProvider {
 
 ## Modifying request handling logic
 
-A frequent use case for components is to modify the way how all requests are handled. For example, the authentication component needs to verify user credentials before the actual handler can be invoked, a logger component needs to record start time and write a log entry when the request has been handled.
+A frequent use case for components is to modify the way requests are handled. For example, the authentication component needs to verify user credentials before the actual handler can be invoked; or a logger component needs to record start time and write a log entry when the request has been handled.
 
 The idiomatic solution has two parts:
 
@@ -179,7 +180,7 @@ The idiomatic solution has two parts:
 
 ### Accessing Elements contributed by other Sequence Actions
 
-When writing a custom sequence action, you need to access Elements contributed by other actions run in the sequence. For example, `authenticate()` action needs information about the invoked route in order to decide whether and how to authenticate the request.
+When writing a custom sequence action, you need to access Elements contributed by other actions run in the sequence. For example, `authenticate()` action needs information about the invoked route to decide whether and how to authenticate the request.
 
 Because all Actions are resolved before the Sequence `handle` function is run, Elements contributed by Actions are not available for injection yet. To solve this problem, use `@inject.getter` decorator to obtain a getter function instead of the actual value. This allows you to defer resolution of your dependency only until the sequence action contributing this value has already finished.
 
@@ -232,14 +233,13 @@ export class AuthenticationProvider {
 ## Extends Application with Mixin
 
 When binding a component to an app, you may want to extend the app with the component's
-properties and methods.
-This can be achieved by using mixins.
+properties and methods by using mixins.
 
-If you are not familiar with the mixin concept, check [Mixin](Mixin.html) to learn more.
+For an overview of mixins, see [Mixin](Mixin.html).
 
-An example of how a mixin leverages component would be `RepositoryMixin`:
-Suppose an app has multiple components with repositories bound to each of them,
-you can use function `RepositoryMixin` to mount those repositories to application level context.
+An example of how a mixin leverages a component is `RepositoryMixin`.
+Suppose an app has multiple components with repositories bound to each of them. 
+You can use function `RepositoryMixin()` to mount those repositories to application level context.
 
 The following snippet is an abbreviated function
 [`RepositoryMixin`](https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/repository-mixin.ts):
@@ -291,7 +291,7 @@ app.find('repositories.*');
 
 ## Configuring components
 
-More often than not, the component may want to offer different value providers depending on the configuration. For example, a component providing Email API may offer different transports (stub, SMTP, etc.).
+More often than not, the component may want to offer different value providers depending on the configuration. For example, a component providing an email API may offer different transports (stub, SMTP, and so on).
 
 Components should use constructor-level [Dependency Injection](Context.html#dependency-injection) to receive the configuration from the application.
 
@@ -307,3 +307,79 @@ class EmailComponent {
   }
 }
 ```
+
+## Creating your own servers 
+
+LoopBack 4 has the concept of a Server, which you can use to create your own
+implementations of REST, SOAP, gRPC, MQTT and more. For an overview, see 
+[Server](Server.html).
+
+Typically, you'll want server instances that listen for traffic on one or more
+ports (this is why they're called "servers", after all). This leads into a key
+concept to leverage for creating your custom servers.
+
+### Controllers and routing
+LoopBack 4 developers are strongly encouraged to use controllers for their
+modules, and this naturally leads to the concept of routing.
+
+No matter what protocol you intend to use for your custom server, you'll need
+to use some algorithm to determine _which_ controller and function to send
+request data to, and that means you need a router.
+
+For example, consider a "toy protocol" similar to the JSON RPC
+specification (but nowhere near as complete or robust).
+
+The toy protocol will require a JSON payload with three properties: `controller`,
+`method`, and `input`.
+
+An example request would look something like this:
+```json
+{
+  "controller": "GreetController",
+  "method": "basicHello",
+  "input": {
+    "name": "world",
+  }
+}
+```
+
+You can find the code for our sample RPC server implementation
+[over here](https://github.com/strongloop/loopback4-example-rpc-server).
+
+### Trying it out
+First, install your dependencies and then start the application:
+```
+npm i && npm start
+```
+
+Now, try it out: start the server and run a few REST requests. Feel free to use
+whatever REST client you'd prefer (this example will use `curl`).
+```sh
+# Basic Greeting Calls
+$ curl -X POST -d '{ "controller": "GreetController", "method": "basicHello" }' -H "Content-Type: application/json" http://localhost:3000/
+Hello, World!
+$ curl -X POST -d '{ "controller": "GreetController", "method": "basicHello", "input": { "name": "Nadine" } }' -H "Content-Type: application/json" http://localhost:3000/
+Hello, Nadine!
+# Advanced Greeting Calls
+$ curl -X POST -d '{ "controller": "GreetController", "method": "hobbyHello", "input": { "name": "Nadine" } }' -H "Content-Type: application/json" http://localhost:3000/
+Hello, Nadine! I heard you like underwater basket weaving.
+$ curl -X POST -d '{ "controller": "GreetController", "method": "hobbyHello", "input": { "name": "Nadine", "hobby": "extreme mountain biking" } }' -H "Content-Type: application/json" http://localhost:3000/
+Hello, Nadine! I heard you like extreme mountain biking.
+```
+
+While a typical protocol server would be a lot more involved in the
+implementation of both its router and server, the general concept remains
+the same, and you can use these tools to make whatever server you'd like.
+
+### Other considerations
+Some additional concepts to add to your server could include:
+- Pre-processing of requests (changing content types, checking the request body,
+etc)
+- Post-processing of responses (removing sensitive/useless information)
+- Caching
+- Logging
+- Automatic creation of default endpoints
+- and more...
+
+LoopBack 4's modularity allows for custom servers of all kinds, while still
+providing key utilities like context and injection to make your work easier.


pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,296 @@ class EmailComponent {
   }
 }
 ```
+
+## Creating your own servers 
+
+LoopBack 4 has the concept of a Server, which you can use to create your own
```
virkt25, line 313: `which can be used` might sound better

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,296 @@ class EmailComponent {
...
+implementations of REST, SOAP, gRPC, MQTT and more. For an overview, see the
+[Server](#Server.html) section.
+
+### A Basic Example
+The concept of servers in LoopBack doesn't actually require listening on ports;
+you can use this pattern to create workers that process data, write logs, or
```
virkt25: or **do** whatever you'd like.

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,296 @@ class EmailComponent {
...
+  async start() {
+    // Run the report processor every 2 minutes.
+    let window = Date.now();
+    interval = setInterval(async () => {
+      const results = await db.find({
+        createdDate: {
+          $gt: window,
+        },
+      });
+      if (results && results.length > 0) {
```
virkt25: `results.length > 0` isn't needed. `forEach` won't throw on an empty array.

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,296 @@ class EmailComponent {
...
+    return Promise.resolve();
+  }
+}
+```
+In this example, the `ReportProcessingServer` will load new reports every two
+minutes from the database (using the provided repository), process those entries
+and then update the database.
+
+### Using your server
+To use your server, simply bind it to your application's context.
```
virkt25: Do we want to call using methods like `.controller`, `.server`, etc. as bindings or does it make sense to say **To use your server, register it with your application via `.server()`**

raymondfeng:
Good stuff. I wonder why you don't use [jsonrpc-2.0](http://www.jsonrpc.org/specification)? The spec is close to what you illustrate.

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,296 @@ class EmailComponent {
...
+  process.exit(0);
+}
+// Event handlers for when we are killed; these aren't *required*.
+process.on('SIGTERM', shutdown);
+process.on('SIGINT', shutdown);
+```
+
+### Trying it out
+Let's start our server up and run a few REST requests. Feel free to use
+whatever REST client you'd prefer (this example will use `curl`).
```
virkt25: Shouldn't it be **in** this example

kjdelisle:
Funny, @bajtos said the same thing when I showed him this earlier. This came from an example I made on-the-fly with the team. 

Wouldn't making it a jsonrpc-2.0 server complicate the Router portion quite a bit?
You'd need to compile a list of all the methods available on the controller definitions to determine which one is appropriate handler, whereas my contrived example lets you skip that part.

kjdelisle:
@jonathan-casarrubias PTAL

I want to know if this answers some of the questions you have/may have had while working on the gRPC server, as well as what you might want to see more of in this document.

raymondfeng:
> Wouldn't making it a jsonrpc-2.0 server complicate the Router portion quite a bit?
You'd need to compile a list of all the methods available on the controller definitions to determine which one is appropriate handler, whereas my contrived example lets you skip that part.

We can use `ControllerName.methodName` as the `method` for json-rpc 2.0.

crandmck:
@kjdelisle I just made some copy edits directly to the branch, rather than via comments.  Hope that's OK.

crandmck (approved):

bajtos:
> I want to know if this answers some of the questions you have/may have had while working on the gRPC server, as well as what you might want to see more of in this document.

Adding @akashjarad to the loop, he volunteered to implement MQTT support - see https://github.com/strongloop/loopback-next/issues/710 and https://github.com/strongloop/loopback-next/issues/647#issuecomment-341017299

kjdelisle:
Rebased, squashed @crandmck 's changes into the first commit

pages/en/lb4/Creating-components.md
kjdelisle: @crandmck Do we want to avoid addressing the user? I can switch to passive voice if we want.

pages/en/lb4/Creating-components.md
kjdelisle: Why call it when I can check if it's got elements?
Also, it could end up being defined, but not a collection with a length property.
Mostly just defensive coding, really.

kjdelisle: Additional point-of-note: The advanced example is real code that I've tested, but this one was largely a thought experiment. It's definitely not robust (there are plenty of holes in the way it looks for reports to process, for example), and it might not even run at all.

pages/en/lb4/Creating-components.md
kjdelisle: Actually, we already have plenty of examples where we're addressing the user:
https://loopback.io/doc/en/lb4/Schemas.html

kjdelisle:
>We can use ControllerName.methodName as the method for json-rpc 2.0.

What if there are two controllers with the same method name?
Though, I guess you could just document that as a limitation of the server...

Hmm. Maybe I should just write a JsonRpcServer, and document the process? :P

pages/en/lb4/Creating-components.md
b-admike: Do we want to make this code snippet something users can copy and paste and run? It'd be cool to get them started, especially since this is our basic example of a server.

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,297 @@ class EmailComponent {
...
+    this.config = config || {};
+  }
+
+  async start(): Promise<void> {
+    this.expressServer = express();
+    this.router = new RPCRouter(this);
+    // Currently, there is a limitation with core node libraries
+    // that prevents the easy promisification of some functions, like
+    // http.Server's listen method.
+    this._server = this.expressServer.listen(
```
b-admike: Can we wrap listen() call with try/catch block?

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,297 @@ class EmailComponent {
...
+      (this.config && this.config.port) || 3000
+    );
+    return Promise.resolve();
+  }
+  async stop(): Promise<void> {
+    this._server.close();
```
b-admike: ditto

pages/en/lb4/Creating-components.md
kjdelisle: If we want this to be copy-pastable, I'd have to expand it to include all of the same boilerplate snippets contained in the advanced example (`index.ts`, `/src/application.ts`, etc)

pages/en/lb4/Creating-components.md
kjdelisle: It's meant to be demo code. I don't want to impact its readability with loads of boilerplate safeguarding and error handling.

It's the same reason it's not riddled with debug statements. :P

pages/en/lb4/Creating-components.md
kjdelisle: see above for response

b-admike:
Okay LGTM

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,297 @@ class EmailComponent {
...
+```ts
+import {Server} from '@loopback/core';
+import {processReports} from '../util';
+export class ReportProcessingServer implements Server {
+  @inject('database') db: DefaultCrudRepository<Report, int>;
+  interval: NodeJS.Timer;
+  async start() {
+    // Run the report processor every 2 minutes.
+    let window = Date.now();
+    interval = setInterval(async () => {
```
bajtos: I think this created a global `interval` property, did you mean `this.interval`?

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,297 @@ class EmailComponent {
...
+Let's create an example server! Please note that this is not a complete example:
+
+```ts
+import {Server} from '@loopback/core';
+import {processReports} from '../util';
+export class ReportProcessingServer implements Server {
+  @inject('database') db: DefaultCrudRepository<Report, int>;
+  interval: NodeJS.Timer;
+  async start() {
+    // Run the report processor every 2 minutes.
```
bajtos: Uh oh, I find this example rather complex and difficult to understand. Can we come up with a simpler example? If not, then please at least refactor the code and extract small named functions that will make it easier to comprehend what's going on. For example:

```ts
export class ReportProcessingServer implements Server {
  // ...
  async start() {
    // Run the report processor every 2 minutes.
    this.interval = setInterval(
      () => this.processReports().catch(err => console.warn('Report processing failed:', err)),
      1000 * 120);
  }

  async processReports() {
    const results = await this.getRecentlyAddedRecords();
    if (!results || !results.length) return;
    await Promise.all(results.map(rep => this.processResult(rep)));
  }

  // ...
}
```

pages/en/lb4/Creating-components.md
bajtos: Also note that errors (promise rejections) returned by this function are ignored and trigger [`process.on('unhandledRejection')`](https://nodejs.org/api/process.html#process_event_unhandledrejection), which is IMO not the right way for handling errors in background processing. You should catch the errors, see e.g. the code snippet in my comment above.

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,297 @@ class EmailComponent {
...
+    let window = Date.now();
```
bajtos: I see two problems here:

 - `window` can be confused with the global variable provided by browsers
 - In my experience, window usually refers to an interval of time (how many seconds do we want to go into past), not a point (what is the first timestamp we want to check).

I am proposing to use a better name, e.g. `timeOfLastRun`.

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,297 @@ class EmailComponent {
...
+    interval = setInterval(async () => {
+      const results = await db.find({
+        createdDate: {
+          $gt: window,
+        },
+      });
+      if (results && results.length > 0) {
+        results.forEach(async (rep) => {
+          await db.updateById(rep.id, await processReport(rep));
```
bajtos: This is horrible in the sense that it does not wait for the update operations to finish and completely ignore any errors that may happen. Please consider using `Promise.all` and `.map` instead, see one of my earlier comments above.

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,297 @@ class EmailComponent {
...
+        }); 
+      }
+      window = Date.now();
```
bajtos:
```ts
const results = await db.find({
        createdDate: {
         $gt: window,
       },
     });
// ...
window = Date.now();
```

This creates a race condition. What if some records have been added after the database finished the query but before we have received all the matching records (and processed them)? Those will be silently ignored.

I understand you want to keep this example simple, but I don't think that's a valid excuse for teaching people to write subtly broken code.

Assuming `processReport` is idempotent (produces the same result when executed multiple times on the same record), the problem can be fixed by storing the current time _before_ querying the database.

```ts
class {
  // ...
  async processReports() {
    const since = this.window; // please, use a better name than "window"
    this.window = Date.now();

    const results = await db.find({
      createdDate: {
        $gt: since,
      },
    });
    // etc. (no changes there)
  }
}
```

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,297 @@ class EmailComponent {
...
+
+### Defining the controller
+Here is the `GreetController` and the `Person` type:
+
+{% include code-caption.html content="/src/controllers/greet.controller.ts" %}
+
+```ts
+export class GreetController {
+  basicHello(input: Person) {
+    return `Hello, ${input && input.name || 'World'}!`;
```
bajtos: It's a bad practice to mix business logic (what's the name to print) with views/templates. Please split this line into two:

```ts
const name = input && input.name || 'World';
return `Hello, ${name}!`;
```

It's not that bad here, but `hobbyHello` below is much more difficult to read in the current form. Here is what I am proposing for `hobbyHello`:

```ts
const hobby = input && input.hobby || 'underwater basket weaving';
return `${this.basicHello(input)} I heard you like ${hobby}.`;
```

(I think this will also make the code easier to debug, I am not sure if V8/DevTools Inspector can step through template parts.)

bajtos (changes requested):
I reviewed about two thirds of the patch up to "Defining our router and server", PTAL at my comment. I'll take a look at the remaining content later, hopefully still today.

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,297 @@ class EmailComponent {
...
+import * as express from 'express';
+import * as parser from 'body-parser';
+
+export class RPCRouter {
+  routing: express.Router;
+  constructor(private server: RPCServer) {
+    this.routing = express.Router();
+    const jsonParser = parser.json();
+    this.routing.post('*', jsonParser, async (request, response) => {
+      console.log(JSON.stringify(request.body));
```
bajtos: What if this line throws an error? The top-level async function will return a rejected promise, which will be ignored because express does not report async middleware/route functions.

Please consider cleaning up this code and making it more robust.

```ts
class {
  constructor(...) {
    this.routing.post('*', jsonParser, (request, response) => {
      this.handleRequest(request, response)
        .catch(err => this.handleError(request, response, error));
    });
  }

  async handleRequest(request, response) {
    console.log(JSON.stringify(request.body));
    // etc.
  }

  handleError(request, response, err) {
     console.error('Cannot handle %s %s:', request.method, request.path, err);
     if (response.headersSent) return;
     response.statusCode = 500;
     response.end();
  });
}
```

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,297 @@ class EmailComponent {
...
+      // Aliasing these values for clarity and brevity.
+      const ctrl = request.body.controller;
+      const method = request.body.method;
+      const input = request.body.input;
+      try {
+        const controller = await this.server.get(`controllers.${ctrl}`);
+        if (!controller) {
+          response.statusCode = 400;
+          response.send(`No controller was found with name "${ctrl}".`);
```
bajtos: I'd personally `throw new HttpErrors.BadRequest`. However, since the controller was not found, it's probably better to send 404? Also, if this RPC is JSON based, shouldn't we return a JSON formatted error, possibly with 200 status code?

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,297 @@ class EmailComponent {
...
+          return;
+        }
+        if (!controller[method]) {
+          response.statusCode = 400;
+          response.send(
+            `No method was found on controller "${ctrl}" with name "${method}".`
+          );
```
bajtos: Ditto.

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,297 @@ class EmailComponent {
...
+        response.send(await controller[method](input));
+      } catch (err) {
+        // You'll probably want more robust error handling in a real server!
+        console.error(err);
+        response.statusCode = 500;
+        response.send(err);
+      }
+    });
+    // Use our router!
+    this.server.expressServer.use(this.routing);
```
bajtos: Why is this custom router a class when all it does is to create an express router and mount it on the server? I find this design confusing. Also why are you creating a new Router instance, when there is only a single route set up? 

Here is a possibly simpler solution:

```ts
export function mountRpcRoute(server) {
  server.expressServer.post('*', jsonParser, async (request, response) => {
    handleRequest(server, request, response)
        .catch(err => this.handleError(server, request, response, error));
  });
}
```

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,297 @@ class EmailComponent {
...
+  async start(): Promise<void> {
+    this.expressServer = express();
+    this.router = new RPCRouter(this);
+    // Currently, there is a limitation with core node libraries
+    // that prevents the easy promisification of some functions, like
+    // http.Server's listen method.
+    this._server = this.expressServer.listen(
+      (this.config && this.config.port) || 3000
+    );
+    return Promise.resolve();
```
bajtos: This line is not needed at all. When an `async` function does not return anything, the runtime will convert it into a promise resolving to `undefined` under the hood.

Please remove the other usages of `return Promise.resolve()` too.

pages/en/lb4/Creating-components.md
bajtos: I am actually quite concerned about this too. People look up to our docs for examples they will be copy-n-pasting into their code and which may eventually get into production.

I agree we should keep this readable, but not at the cost of showing bad practices!

Here in particular we should use `p-event` to ensure `error` event reported by the server is correctly handled.

```ts
this._server = this.expressServer.listen(...);
return await pEvent(this._server, 'listen');
```

pages/en/lb4/Creating-components.md
bajtos: There is actually no need to wrap this in a try/catch block here. If `close()` throws, the error will be converted into promise rejection and the promise returned by the async `stop()` function will be rejected with that.

However, it's probably a good practice to wait until the server is closed, and maybe return an error when it cannot be closed (although I am not sure what the caller is expected to do in such case, as there is little to do when the server cannot be closed).

```ts
this._server.close();
return await pEvent(this._server, 'close');
```

pages/en/lb4/Creating-components.md
bajtos: It is also important to wait for `listen` event before returning from `start()`, otherwise our app can print "Server is running on port XYZ" before the server has actually started to listen 😱

What's worse, if there are tests calling `app.start()` and making HTTP requests immediately after `start` returns, then they can sporadically fail when the timing is wrong and `listen` takes more ticks of event-loop/promise micro-queue than usually and the request is started before the server started listening.

pages/en/lb4/Creating-components.md
```diff
@@ -307,3 +307,297 @@ class EmailComponent {
...
+const app = new MyApplication();
+
+app.start().catch(err => {
+  console.error(`Unable to start application: ${err}`);
+});
+async function shutdown() {
+  await app.stop();
+  process.exit(0);
+}
+// Event handlers for when we are killed; these aren't *required*.
```
bajtos: Well, if we want to keep our examples in this page short and readable, then I think this section should be removed, since it's not really needed? If you want to show how to use stop, then perhaps use an end-to-end test as the example and call start/stop from before/after hooks.

bajtos (dismissed):
Few more comments regarding the code snippets.

At high level, I am missing any mention of Sequences. I think it's important to explain why we use the Sequence concept in REST, why extension developers should use the Sequence concept for their own transports too, and also show an example how to implement a new Sequence with new sequence actions for a custom transport.

Thoughts?

pages/en/lb4/Creating-components.md
bajtos: >  it might not even run at all

Please, let's avoid code that looks like it should run but that does not actually run. If you only want to outline the implementation, then use some sort of a pseudo code.

You can also stick a big fat warning saying the file does not compile, but that would look pretty lame, at least to me. We are writing these code examples to be read by many people, I think the time needed to verify that the code examples actually work is very well worth it!

raymondfeng:
It might be better off if we create a github repo to host the sample code and README. The doc site can just make a reference to the README.

pages/en/lb4/Creating-components.md
kjdelisle: Including `pEvent` seems a bit odd. I didn't even know it was a thing (I think you mentioned it in your presentation, but I missed the start!). Not everyone will know what pEvent is/does, and I think it detracts from the example, which is why I stuck the comment up there instead.

>It is also important to wait for listen event before returning from start(), otherwise our app can print "Server is running on port XYZ" before the server has actually started to listen 😱

I had a version that handled this using old-school promises:
```ts
return new Promise((resolve, reject) => {
  // using this ugly form to call resolve or reject on the event.
});
```

I wasn't a fan because it detracted from the core logic. I think I'll just use p-event.

kjdelisle:
>At high level, I am missing any mention of Sequences. I think it's important to explain why we use the Sequence concept in REST, why extension developers should use the Sequence concept for their own transports too, and also show an example how to implement a new Sequence with new sequence actions for a custom transport.

My thoughts are that Sequence is not demonstrably a best practice, and that it would only serve to complicate a developer's implementation if they originally intended to use middleware chaining or some other pattern.

In our case, a router is a natural consequence of our framework's opinion regarding the use of controllers but the same can't be said for something like sequence.

kjdelisle:
@raymondfeng I'm starting to think so as well. I actually have the code handy, so I might just create a new repository for it and edit the PR to point there.

pages/en/lb4/Creating-components.md
kjdelisle: It was mostly to draw attention to the concept by isolating it as a class.

kjdelisle: Additionally, it makes it way simpler to mock pieces of the code if it's a class instead of a function.

kjdelisle:
@bajtos I've moved the code to a different repo, PTAL

jannyHou (approved):

virkt25 (approved):

bschrammIBM:
Do you still me to review this? There seems to be a long history.

kjdelisle:
@bschrammIBM It's already landed, but if there are concerns or comments you'd like addressed, you can still perform a review after the fact and I can try to address them in a follow-up PR.

jonathan-casarrubias:
@kjdelisle sorry for the late response...

Anyway I think this is a good start but  my first impression is that still is too basic for more complex extensions, I'm not sure how to improve the docs without creating a tutorial instead, but the Server example is not quite the same as the REST and gRPC components are actually built...

Somehow the Server examples demonstrate how to extend an Application but for me it tastes more like an application developer level, rather than an extension developer level example

PR #530, opened 2017-11-10
Add a hosted swagger UI for LoopBack apps

raymondfeng:
The PR enables https://loopback.io/api-explorer as the hosted UI for LoopBack 4 api explorer.


api-explorer/index.html
```diff
@@ -0,0 +1,101 @@
...
+<div id="swagger-ui"></div>
+
+<script src="./swagger-ui-bundle.js"> </script>
+<script src="./swagger-ui-standalone-preset.js"> </script>
+<script>
+window.onload = function() {
+  
+  // Build a system
+  const ui = SwaggerUIBundle({
+    url: "http://petstore.swagger.io/v2/swagger.json",
```
virkt25: Would it be better to point this to `localhost:3000/swagger.json` OR an example also hosted on `loopback.io` just because someone might wonder why we're pointing to petstore.

raymondfeng: The page will be only used for redirect with `?url=...`.

kjdelisle:
I think we should be making this an offline module rather than re-hosting it on our own domain.

Don't get me wrong, I like that this will solve the problem of us pointing to "petstore", but I would much prefer packaging it as an installable module for the monorepo (maybe make it a Server?)

api-explorer/index.html
bajtos: @raymondfeng could you please add a comment capturing this discussion? This will make it easier for future readers to understand where did the petstore URL come from.

bajtos (changes requested):
Great stuff!

I don't have bandwidth to verify whether the front-end works as expected.

I have one concern though: what are the instructions for upgrading swagger-ui to a newer version? Are there any customizations we need to preserve (restore after upgrade)?

@raymondfeng please add a file describing the upgrade process - where to get swagger-ui bundle, what files to copy over, what customizations to preserve. Perhaps `api-explorer/README.md` would be a good place for this?

bajtos:
> I think we should be making this an offline module rather than re-hosting it on our own domain.
> 
> Don't get me wrong, I like that this will solve the problem of us pointing to "petstore", but I would much prefer packaging it as an installable module for the monorepo (maybe make it a Server?)

As I understand it, the plan is to provide a hosted version of swagger-ui as a short-term solution, until we have bandwidth to implement a proper API Explorer that is capable to run locally (perhaps as part of the target application - we are doing this in loopback-component-explorer).

kjdelisle (approved):
I'm okay with it as a stop-gap, but I'll defer to @bajtos

raymondfeng:
@bajtos Comments addressed. PTAL.

bschrammIBM (commented):
AFAIK this can be merged. RFeng confirmed there will be no link on the loopback.io doc site. I request some introductory text for how this will work for the developer. Is it opened automatically?

api-explorer/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "saveSpec": null,
+    "fetchSpec": "^3.4.4"
+  },
+  "_requiredBy": [
+    "#DEV:/"
+  ],
+  "_resolved": "https://registry.npmjs.org/swagger-ui-dist/-/swagger-ui-dist-3.4.4.tgz",
+  "_shasum": "7dd0d7f7884e4387b9d10d4f4f7b0f93436a11d2",
+  "_spec": "swagger-ui-dist@^3.4.4",
+  "_where": "/Users/rfeng/Projects/docs/loopback.io",
```
bajtos: Hmm, this refers to your local checkout, I find that suspicious.

This whole `package.json` file looks to me like it's coming from `node_modules/swagger-ui` after running `npm install swagger-ui`. Is that intentional?

api-explorer/README.md
```diff
@@ -0,0 +1,10 @@
+# Hosted API Explorer
+
+This directory contains [swagger-ui](https://github.com/swagger-api/swagger-ui) assets that serve as a hosted API Explorer for LoopBack 4 applications. The UI will be available at https://loopback.io/api-explorer?url=<swagger-spec-url>.
```
bajtos: I think you should use different characters than `<` and `>` here, because angle brackets are interpreted as HTML markdown, effectively hiding `<swagger-spec-url>` part from the URL.  See [the rendered README](https://github.com/strongloop/loopback.io/blob/3c7c493e56e354dae2862d2e3f1301d85a5989f8/api-explorer/README.md).

api-explorer/package.json
bajtos: Can we modify `update-swagger-ui.sh` to skip copying `package.json`?

api-explorer/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+To upgrade swagger-ui to a newer version:
+
+1. Update the [`swagger-ui-dist`](https://www.npmjs.com/package/swagger-ui-dist) version in [package.json](https://github.com/strongloop/loopback.io/blob/gh-pages/package.json). You can use `npm update swagger-ui-dist`.
```
bajtos: I love the simplicity of an update workflow that leverages `npm` 👍

update-swagger-ui.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# This script copies files from `swagger-ui-dist` module into `api-explorer`
```
bajtos: Shell/bash scripts won't work on Windows, could you please rewrite this in Node.js?

Here are few modules you may find useful:

 - [rimraf](https://www.npmjs.com/search?q=rimraf) instead of `rm -rf` (if needed) 
 - [cpr](https://www.npmjs.com/package/cpr) instead of `cp`
 - [fs-extra](https://www.npmjs.com/package/fs-extra) provides both copy & remove
 - [shelljs](https://www.npmjs.com/package/shelljs) (similar to fs-extra)

api-explorer/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+2. Run `npm install` which in turn executes `./update-swagger-ui.sh` to copy files into `api-explorer`.
+
```
bajtos, line 10: IIUC, the third step is to add any new or modified files to git and send a PR with the changes. Can you mention this in the instructions please?

Ideally, we should be deleting files that are no longer part of swagger-ui too - I think that's something to handle in `update-swagger-ui` script, for example by removing everything from the `api-explorer` directory except "our" files (README.md, index.loopback.html).

bajtos (approved):
Much better!

raymondfeng:
@bajtos Reworked the script using Node and addressed some of your comments.

PR #539, opened 2017-11-15
Added necessary quotes on column name list

buokae:


slnode:

bschrammIBM (approved):

bajtos:
@jannyHou @kjdelisle could you PTAL if this is correct? I am not familiar MySQL indexes.

bschrammIBM:
Thank you for submitting this edit. We are in the process of reviewing it.

kjdelisle:
Sorry, was neck deep in other tasks.

I decided to try out a fresh app with a composite index using this approach, but without the backticks to escape the column names and it works just fine.

I created a Note template app and then added two new models:
- NoteUser, so I could create relations against the User table
- NoteShare, which is our through model

#### note-share.json
```json
{
  "name": "NoteShare",
  "base": "PersistedModel",
  "options": {
    "validateUpsert": true
  },
  "properties": {
    "userId": {
      "type": "string",
      "required": true
    },
    "noteId": {
      "type": "number",
      "required": true
    },
    "createdOn": {
      "type": "date"
    }
  },
  "validations": [],
  "indexes": {
    "NOTE_SHARE_INDEX": {
      "columns": "userId,noteId",
      "kind": "unique"
    }
  },
  "relations": {
    "note": {
      "type": "belongsTo",
      "model": "Note",
      "foreignKey": "noteId"
    },
    "noteUser": {
      "type": "belongsTo",
      "model": "NoteUser",
      "foreignKey": "noteUserId"
    }
  },
  "acls": [],
  "methods": {}
}

```

I then ran a script as a part of app boot to automigrate the table definitions to the MySQL instance, then I jumped into the container hosting this database and ran a query against the schema of the NoteShare table to check the indices (indexes).


#### Console Output from mysql command line
```
mysql> SHOW INDEX from NoteShare;
+-----------+------------+------------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| Table     | Non_unique | Key_name         | Seq_in_index | Column_name | Collation | Cardinality | Sub_part | Packed | Null | Index_type | Comment | Index_comment |
+-----------+------------+------------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| NoteShare |          0 | PRIMARY          |            1 | id          | A         |           0 |     NULL | NULL   |      | BTREE      |         |               |
| NoteShare |          0 | NOTE_SHARE_INDEX |            1 | userId      | A         |           0 |     NULL | NULL   |      | BTREE      |         |               |
| NoteShare |          0 | NOTE_SHARE_INDEX |            2 | noteId      | A         |           0 |     NULL | NULL   |      | BTREE      |         |               |
+-----------+------------+------------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
```

So technically, you don't need to do this for the composite index to work. ~~I'd advise doing it if you have a columnName that requires escaping, but the tutorial example is fine as it is, IMO.~~

EDIT: After another look, the backticks work just as well and will help people avoid having issues surrounding escape characters.

kjdelisle:
@buokae After another look, the backticks work just as well and will help people avoid having issues surrounding escape characters.

If you could update this PR to add a quick blurb explaining why backticks are being used in the column name, that would be all we need to land this change.

buokae:
@kjdelisle Sorry for assuming that the backticks would be necessary in every case, and thanks for showing me they aren't.

I get what you're asking and I would gladly do it, but:

In my current project I have the following index on a model:
```
  "indexes": {
    "UK_File_account_order_filename": {
      "columns": "account,order,filename",
      "kind": "unique",
      "description": "Constraint to keep the combination of account, order and filename unique."
    }
  },
```

As you see, no backticks, three columns in the index. When in run autoupdate on that model it gives the following error:
ErrorError: ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'order,filename)' at line 1

It does work however when I put the columns in with backticks like this:
```
      "columns": "`account`,`order`,`filename`",
```
So, I guess I'll have to do a little test with a minimal code base myself to check what's really wrong here.

I'm on 
    "loopback": "^3.16.2",
    "loopback-boot": "^2.27.0",
    "loopback-connector-mysql": "^5.2.0",

I'll get back to you.

PR #542, opened 2017-11-21
translated some pages to Japanese

t-eimizu:
- translated some pages.
- deleted unnecessary files.
- fixed invalid encoding (cp932 => UTF-8)
- fixed layout (translate => page)
- added lang parameter for include.


slnode:

bschrammIBM (approved):
Your translations are approved.

t-eimizu:
Thanks a lot!

bajtos:
Hello, thank you @t-eimizu for the pull request and sorry for the delay of landing it.

I am actually quite worried about those `if include.lang == "ja"` directives in `_includes` files. If we use this approach for each translation, then the shared (included) bits would become way too long and difficult to maintain.

Is it possible to come up with a different approach where there is one file for each language in `_includes` too? For example, we can have `_includes/content/angular-methods-caveat.html` for English and `_includes/content/angular-methods-caveat.ja.html` for Japanese.  `_includes/content/ja/angular-methods-caveat.html` would be even better, if that's supported by Jekyll.

t-eimizu:
Hi, @bajtos. Thank you for suggestion. 
I think that the approach you propose is good too.
I will try a method to split the file to be included for each language.

bschrammIBM (changes requested):
please make the change requested by the eng team

PR #544, opened 2017-11-21
Update HasManyThrough-relations.md

Tobmai:


slnode:

bschrammIBM (changes requested):
Can you please provide an explanation for why this change is needed?

dhmlau:
@jannyHou , PTAL?

jannyHou (approved):
LGTM

bschrammIBM:
Here is link to the page:
https://loopback.io/doc/en/lb3/HasManyThrough-relations.html

bschrammIBM (approved):

jannyHou:
Hi @bschrammIBM, I think the reason is we created the instance with `{patientId: 1}`: https://github.com/strongloop/loopback.io/pull/544/files#diff-7b5f81ba00268175654aba0eb0bd4e26L324, so it makes sense that we also query with the same value `1`

Tobmai:
@jannyHou exactly. 
Additionally in a sample response (https://loopback.io/doc/en/lb3/HasManyThrough-relations.html) we got "patientId: 1", not "patientId: 2" like in GET query.
```
[
  {
    "appointmentDate": "2014-06-01",
    "id": 1,
    **"patientId": 1,**
    "physicianId": 1,
    "patient": {
      "name": "Jane Smith",
      "id": 1
    }
  }
]
```

PR #547, opened 2017-11-23
Testing extensions

virkt25:
First draft for testing your extensions. Tried to be a high level summary of what to test for various extension artifacts as well as the types of extensions.

### DIFF
--- a/pages/en/lb4/Testing-Your-Extensions.md
+++ b/pages/en/lb4/Testing-Your-Extensions.md
@@ -1,0 +1,280 @@
+---
+lang: en
+title: 'Testing your extension'
+keywords: LoopBack 4.0, LoopBack 4
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Testing-your-extension.html
+summary:
+---
+
+## Overview
+
+LoopBack 4 extensions are often used by other teams. A thorough test suite for your extension brings powerful benefits to all your users, including:
+
+* Validating the behavior of the extension
+* Preventing unwanted changes to the API or functionality of the extension
+* Providing working samples and code snippets that serve as functional documentation for your users
+
+## Project Setup
+
+We recommend that you use `@loopback/cli` to create the extension, as it installs several tools you can use for testing, such as `mocha`, assertion libraries, and a swagger validator.
+
+The `@loopback/cli` includes the `mocha` automated test runner and a
+`test` folder containing recommended folders for various types of tests.
+`Mocha` is enabled by default if `@loopback/cli` is used to
+create the extension project. The `@loopback/cli` installs and configures `mocha`, creates the `test` folder, and also enters a `test` command in your `package.json`.
+
+Assertion libraries such as [ShouldJS](http://shouldjs.github.io/) (as `expect`),
+[SinonJS](http://sinonjs.org/), and a swagger validator are made available
+through the convenient `@loopback/testlab` package. The `testlab` is also installed by `@loopback/cli`.
+
+### Manual Setup - Using Mocha
+
+* Install `mocha` by running `npm i --save-dev mocha`. This will save the `mocha` package in
+`package.json` as well.
+* Under `scripts` in `package.json` add the following:
+`test: npm run build && mocha --recursive ./dist/test`
+
+## Types of tests
+
+A comprehensive test suite tests many aspects of your code. We recommend that you write unit, integration, and acceptance tests to test your application from a variety of perspectives. Comprehensive testing
+ensures correctness, integration, and future compatibility.
+
+You may use any development methodology you want to write your extension; the
+important thing is to test it with an automated test suite. In Traditional development methodology, you write the code first and then write the tests. In Test-driven development methodology, you write the tests first, see them fail, then write the code to pass the tests.
+
+### Unit Tests
+
+A unit test tests the smallest unit of code possible, which in this case is a function.
+Unit tests ensure variable and state changes by outside actors don't affect the
+results. [Test doubles](https://en.wikipedia.org/wiki/Test_double) should be
+used to substitute function dependencies. You can learn more about test doubles
+and Unit testing here: [Testing your Application: Unit testing](http://loopback.io/doc/en/lb4/Testing-your-application.html#unit-testing).
+
+#### Controllers
+
+At its core, a controller is a simple class that is responsible for related
+actions on an object. Performing unit tests on a controller in an extension is the same as
+performing unit tests on a controller in an application.
+
+To test a controller, you instantiate a new instance of your controller
+class and test a function, providing a test double for constructor arguments as
+needed. Following are examples that illustrate how to perform a unit test on a controller class:
+
+**`src/controllers/ping.controller.ts`**
+```ts
+export class PingController {
+  @get('/ping')
+  ping(msg?: string) {
+    return `You pinged with ${msg}`;
+  }
+}
+```
+
+**`test/unit/controllers/ping.controller.ts`**
+```ts
+import {PingController} from '../../..';
+import {expect} from '@loopback/testlab';
+
+describe('PingController() unit', () => {
+  it('pings with no input', () => {
+    const controller = new PingController();
+    const result = controller.ping();
+    expect(result).to.equal('You pinged with undefined');
+  });
+
+  it('pings with msg \'hello\'', () => {
+    const controller = new PingController();
+    const result = controller.ping('hello');
+    expect(result).to.equal('You pinged with hello');
+  });
+});
+```
+
+You can find an advanced example on testing controllers in [Unit test your Controllers](http://loopback.io/doc/en/lb4/Testing-your-application.html#unit-test-your-controllers).
+
+#### Decorators
+
+The recommended usage of a decorator is to store metadata about a class or a
+class method. The decorator implementation usually provides a function
+to retrieve the related metadata based on the class name and method name.
+For a unit test for a decorator, it is important to test that that it stores and retrieves the correct metadata. *The retrieval gets tested as a
+result of validating whether the metadata was stored or not.*
+
+Following is an example for testing a decorator:
+
+**`src/decorators/test.decorator.ts`**
+```ts
+export function test(file: string) {
+  return function(target: Object, methodName: string): void {
+    Reflector.defineMetadata(
+      'example.msg.decorator.metadata.key',
+      {file},
+      target,
+      methodName,
+    );
+  };
+}
+
+export function getTestMetadata(
+  controllerClass: Constructor<{}>,
+  methodName: string,
+): {file: string} {
+  return Reflector.getMetadata(
+    'example.msg.decorator.metadata.key',
+    controllerClass.prototype,
+    methodName,
+  );
+}
+```
+
+**`test/unit/decorators/test.decorator.ts`**
+```ts
+import {test, getTestMetadata} from '../../..';
+import {expect} from '@loopback/testlab';
+
+describe('test.decorator (unit)', () => {
+  it('can store test name via a decorator', () => {
+    class TestClass {
+      @test('me.test.ts')
+      me() {}
+    }
+
+    const metadata = getTestMetadata(TestClass, 'me');
+    expect(metadata).to.be.a.Object();
+    expect(metadata.file).to.be.eql('me.test.ts');
+  });
+});
+```
+
+#### Mixins
+
+A Mixin is a TypeScript function that extends the `Application` Class, adding
+new constructor properties, methods, etc. It is difficult to write a unit test
+for a Mixin without the `Application` Class dependency. The recommended practice
+is to write an integration test is described in [Mixin Integration Tests](#mixin-integration-tests).
+
+#### Providers
+
+A Provider is a Class that implements the `Provider` interface. This interface
+requires the Class to have a `value()` function. A unit test for a provider
+should test the `value()` function by instantiating a new `Provider` class, using
+a test double for any constructor arguments.
+
+**`src/providers/random-number.provider.ts`**
+```ts
+import {Provider} from '@loopback/context';
+
+export class RandomNumberProvider implements Provider<number> {
+  value(): number {
+    return (max: number): number => {
+      return Math.floor(Math.random() * max) + 1;
+    };
+  }
+}
+```
+
+**`test/unit/providers/random-number.unit.test.ts`**
+```ts
+import {RandomNumberProvider} from '../../..';
+import {expect} from '@loopback/testlab';
+
+describe('RandomNumberProvider (unit)', () => {
+  it('generates a random number within range', () => {
+    const provider = new RandomNumberProvider().value();
+    const random: number = provider(3);
+
+    expect(random).to.be.a.Number();
+    expect(random).to.equalOneOf([1, 2, 3]);
+  })
+})
+```
+
+#### Repositories
+
+*This section will be provided in a future version.*
+
+### Integration Tests
+
+An integration test plays an important part in your test suite by ensuring your
+extension artifacts work together as well as `@loopback`. It is
+recommended to test two items together and substitute other integrations as test doubles so it becomes apparent where the integration errors may occur.
+
+#### Mixin Integration Tests
+
+A Mixin extends a base Class by returning an anonymous class. Thus, a Mixin is tested by actually
+using the Mixin with its base Class. Since this requires two Classes to work
+together, an integration test is needed. A Mixin test checks that
+new or overridden methods exist and work as expected in the new Mixed class. Following is an example for an
+integration test for a Mixin:
+
+**`src/mixins/time.mixin.ts`**
+```ts
+import {Constructor} from '@loopback/context';
+export function TimeMixin<T extends Constructor<any>>(superClass: T) {
+  return class extends superClass {
+    constructor(...args: any[]) {
+      super(...args);
+      if (!this.options) this.options = {};
+
+      if (typeof this.options.timeAsString !== 'boolean') {
+        this.options.timeAsString = false;
+      }
+    }
+
+    time() {
+      if (this.options.timeAsString) {
+        return new Date().toString();
+      }
+      return new Date();
+    }
+  };
+}
+```
+
+**`test/integration/mixins/time.intg.test.ts`**
+```ts
+import {expect} from '@loopback/testlab';
+import {Application} from '@loopback/core';
+import {TimeMixin} from '../../..';
+
+describe('TimeMixin (integration)', () => {
+  it('mixed class has .time()', () => {
+    const myApp = new AppWithTime();
+    expect(typeof myApp.time).to.be.eql('function');
+  });
+
+  it('returns time as string', () => {
+    const myApp = new AppWithLogLevel({
+      timeAsString: true,
+    });
+
+    const time = myApp.time();
+    expect(time).to.be.a.String();
+  });
+
+  it('returns time as Date', () => {
+    const myApp = new AppWithLogLevel();
+
+    const time = myApp.time();
+    expect(time).to.be.a.Date();
+  });
+
+  class AppWithTime extends TimeMixin(Application) {}
+});
+```
+
+### Acceptance Test
+
+An Acceptance test for an extension is a comprehensive test written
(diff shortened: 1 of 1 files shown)


pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,305 @@
...
+create the extension project. This will install and configure `mocha` as well
+as create a `test` command in your `package.json`.
+
+Assertion libraries such as [ShouldJS](http://shouldjs.github.io/) (as `expect`),
+[SinonJS](http://sinonjs.org/), and a swagger validator are made available
+through the convenitent `@loopback/testlab` package. This is installed if the
+project was created using the CLI.
+
+### Manual Setup - Using Mocha
+- Install `mocha` by running `npm i mocha`. This will save the package in
```
b-admike: `npm i --save-dev mocha` here to save it as a dev dep?

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,305 @@
...
+A unit test checks the smallest unit of code possible, in this case a function.
+This helps to ensure variable and state changes by outside actors don't affect
+the result. If the function has dependencies, you should substitute them with
+[test doubles](https://en.wikipedia.org/wiki/Test_double) to ensure only the
+function you are testing affects the results. SinonJS is our recommended
+framework to create test doubles in the form of spies, stubs, and mocks.
+
+#### Controllers
+
+At it's core, a controller is a simple class that is responsible for related
```
b-admike: nitpick: `its`

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,305 @@
...
+actions on a object. Unit testing a controller in an extension is the same as
```
b-admike: nitpick: `an` object

pages/en/lb4/Testing-Your-Extensions.md
kjdelisle: ^

pages/en/lb4/Testing-Your-Extensions.md
kjdelisle: ^

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,305 @@
...
+code to pass the tests
+
+### Unit Tests
+
+A unit test checks the smallest unit of code possible, in this case a function.
+This helps to ensure variable and state changes by outside actors don't affect
+the result. If the function has dependencies, you should substitute them with
+[test doubles](https://en.wikipedia.org/wiki/Test_double) to ensure only the
+function you are testing affects the results. SinonJS is our recommended
+framework to create test doubles in the form of spies, stubs, and mocks.
```
kjdelisle: I think that we could probably make this more succinct, since this is a repetition of what we cover in the other testing guide. I'd just link back to it as a reference.

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,305 @@
...
+}
+```
+
+**`test/unit/controllers/ping.controller.ts`**
+```ts
+import {PingController} from '../../..';
+import {expect} from '@loopback/testlab';
+
+describe('PingController() unit', () => {
+  it('.ping() (no arg)', () => {
```
kjdelisle: I thought our test practice was to use descriptive names for the test labels.
Something like `it('pings with no input', ...)`

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,305 @@
...
+```
+
+You can find a more advanced example on testing controllers in [Unit test your Controllers](http://loopback.io/doc/en/lb4/Testing-your-application.html#unit-test-your-controllers).
+
+#### Decorators
+
+The recommended usage of a decorator is to store metadata about a class or a
+class method. The decorator implementation will usually also provide a function
+to retrieve the metadata stored by it based on the className and method name.
+So for a unit test, it is important to test that we can store metadata via a
```
kjdelisle: Drop the "so" to avoid passive voice.

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,305 @@
...
+  });
+});
+```
+
+#### Mixins
+
+A Mixin is a function that extends a base Class (usually `Application` Class)
+and returns an anonymous class, adding new contructor properties, methods, etc.
+to the base Class. Due to the fact that the Mixin returns an anonymous class,
+it is not possible to write a unit test in isolation for a Mixin. The
```
kjdelisle: I would rephrase this to be more specific to TypeScript. You technically can write a unit test for a JavaScript mixin, since you can mix your artifacts into an empty class/object. Our mixins will most likely be constrained to classes that extend Application so it's not a *universal* truth, but it's still something important to keep in mind.

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,305 @@
...
+recommended practice is to write an integration test for a Mixin as shown [here: Integrations Tests Mixins]().
+
+#### Providers
+
+A Provider is a Class, similar to a Controller but it implements the Provider
```
kjdelisle: I would avoid the comparison to controllers.
Providers are meant to resolve artifacts without given input; I'd like to think of them as "default factories".
Controllers are meant to handle requests and return responses, which is always an input-driven/context-driven operation (not to be confused with `Context`).

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+`test: npm run build && mocha --recursive ./dist/test`
+
+## Types of tests
+
+A comprehensive test suite tests many aspects of your code. It is encouraged to
+test as many aspects of your application from a variety of perspectives to
+ensure correctness, integration and ensuring future compatibility. To accomplish
+this you should write unit, integration and acceptance tests.
+
+You may use any development methodology you want to write your extension, the
```
shimks: `your extension,`
use a semicolon instead

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+important thing is to have an automated test suite.
+- **Traditional development:** write the code first and then write the tests
+- **Test-driven development:** write tests first, see them fail, then write the
+code to pass the tests
+
+### Unit Tests
+
+A unit test checks the smallest unit of code possible, in this case a function.
+This ensures variable and state changes by outside actors don't affect the
```
shimks: `actors`
factors

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+results. [Test doubles](https://en.wikipedia.org/wiki/Test_double) should be
+used to substitue function dependencies. You can learn more about test doubles
```
shimks: `substitue`
substitute

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+  });
+});
+```
+
+#### Mixins
+
+A Mixin is a TypeScript function that extends the `Application` Class, adding
+new constructor properties, methods, etc. It is difficult to write a unit test
+for a Mixin without the `Application` Class dependency. The recommended practice
+is to write an integration test ofr a Mixin as show below.
```
jannyHou: nitpick: typo: 
> an integration test ofr a Mixin as show below.

`ofr` --> `of`

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+
+You can find a more advanced example on testing controllers in [Unit test your Controllers](http://loopback.io/doc/en/lb4/Testing-your-application.html#unit-test-your-controllers).
+
+#### Decorators
+
+The recommended usage of a decorator is to store metadata about a class or a
+class method. The decorator implementation will usually also provide a function
+to retrieve the metadata stored by it based on the className and method name.
+For a unit test, it is important to test that we can store metadata via a
+decorator and retrieve the correct metadata. *The retrieval gets tested as a
```
shimks: `The retrieval gets tested as a consequence of validating the metadata was stored`
I think this sounds better: 'the retrieval gets tested as a consequence of validating **whether** the metadata was stored **or not**'. Alternative could be 'validating metadata storage'

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Integration Tests
+
+An integration test plays an important part in your test suite by ensuring your
+extension artifacts play well with each other as well as `@loopback`. It is
+recommended to test two items together and other integrations can be
+substituted by test doubles so it becomes apparent where the integration breaks
+(if it does).
+
+#### Controllers
```
jannyHou: maybe we can direct users to http://loopback.io/doc/en/lb4/Testing-your-application.html#integration-testing if we haven't decided the content here.

pages/en/lb4/Testing-Your-Extensions.md
shimks: `ofr`
of

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+
+#### Decorators
+
+*TBD ... How / What?*
+
+#### Mixins
+
+As mentioned under unit/mixins section, a Mixin extends a base Class by returning
+an anonymous class. As such to test a Mixin, it should be done by actually
+using the mixin with it's base Class. Since this requires two Classes to work
```
shimks: `it's`
its

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+
+*TBD ... How / What?*
+
+#### Repositories
+
+*TBD ... How / What*
+
+### Acceptance Test
+
+An Acceptance test is the ultimate test. This is a comprehensive test, written
```
shimks: `This is a comprehensive test, written end-to-end that is responsible ...`
Grammar: This is a comprehensive test written end-to-end **which** is reponsible ...
The comma needs to be either moved after 'end-to-end' or gotten rid of entirely (I'm not sure)

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+end-to-end that is responsible for covering the user scenario. This acceptance
+test should make use of all extension artifacts such as decorators, mixins,
+providers, repositories, etc. No test doubles should need to be used for an
+Acceptance Test. This is a black box test where you don't know / care about the
+internals of the extensions. You will be thinking like the consumer of the
+extension (using it, and it doing a thing as expected).
```
shimks: `(using it, and it doing a thing as expected)`
Awkward phrasing; stuff in parenthesis should make sense grammatically even without them

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+
+Due to the complexity of an acceptance test, there is no example given here.
+Have a look at a [loopback4-example-log-extension](https://github.com/strongloop/loopback4-example-log-extension)
+to understand the extension artifacts and usage. Then you can see the acceptance
```
shimks: `and usage`
and their usage
doesn't matter too much imo

shimks: `Then you can see the acceptance ...`
Awkward phrasing. I'd go with something like 'You can see its acceptance test here'

shimks (approved):
grammar galore

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+
+    expect(random).to.be.a.Number();
+    expect(random).to.equalOneOf([1, 2, 3]);
+  })
+})
+```
+
+#### Repositories
+
+*TBD ... How / What?*
```
kjdelisle: This is a bit of a weird one. In our typical case, people will be using whatever repository implementation is provided by their utilities (either the way legacy-juggler builds Repository instances, or some other ORM).

In the case of [loopback4-extension-typeorm](https://github.com/strongloop/loopback4-extension-typeorm), it provides the functions you need to leverage the same concepts so that you don't have to build the repositories yourself as a user at runtime. What you'd test would be whether or not that binding is occurring, rather than how it gets made (in TypeORM's case, the _how_ is TypeORM's problem, not ours).

If an extension were to provide functions that directly produce objects that are "repositories", then that should be unit tested accordingly.

The TL;DR here is that if someone provides some form of repository functionality in their extension, they should either try to limit its responsibility to wiring up the repositories. If they can't, then they should write the usual unit and integration tests to prove that it works.

(Even talking about this stuff feels like it's something we shouldn't be writing an _opinion piece_ on, but I digress)

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Testing-your-extension.html
+summary:
+---
+
+## Overview
+
+An extension for LoopBack 4 is likely to be used by thousands and as such it's
+important to have a thorough test suite for an extension because:
```
dhmlau: is "to" better than "because" here?

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+- Ensure your extension works correctly
+- Prevent regressions when new features are added or bugs are fixed
+- Helps demonstrate usage
```
dhmlau: "Helps" --> "Help"?

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+used to substitue function dependencies. You can learn more about test doubles
+and Unit testing [here: Testing your Application: Unit testing](http://loopback.io/doc/en/lb4/Testing-your-application.html#unit-testing).
+
+#### Controllers
+
+At its core, a controller is a simple class that is responsible for related
+actions on an object. Unit testing a controller in an extension is the same as
+unit testing a controller for an application.
+
+To test a controller you want to instantiate a new instance of your controller
```
dhmlau: Suggestion:
"To test a controller, you can instantiate ...."

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+});
+```
+
+You can find a more advanced example on testing controllers in [Unit test your Controllers](http://loopback.io/doc/en/lb4/Testing-your-application.html#unit-test-your-controllers).
+
+#### Decorators
+
+The recommended usage of a decorator is to store metadata about a class or a
+class method. The decorator implementation will usually also provide a function
+to retrieve the metadata stored by it based on the className and method name.
```
dhmlau: Instead of "based on the className and method name", how about "based on the class name and method name"?

pages/en/lb4/Testing-Your-Extensions.md
dhmlau: also "as shown" instead of "as show"

pages/en/lb4/Testing-Your-Extensions.md
dhmlau: From the readers' perspective, i think it's better to either remove the section or put "coming soon" (something like that) under it.   I'd actually prefer to remove it and create a separate ticket to track the outstanding items (whether we will write later or we decide we should write or not).

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+
+An integration test plays an important part in your test suite by ensuring your
+extension artifacts play well with each other as well as `@loopback`. It is
+recommended to test two items together and other integrations can be
+substituted by test doubles so it becomes apparent where the integration breaks
+(if it does).
+
+#### Controllers
+
+*TBD ... Repositories? Other Services?*
```
dhmlau: same comments from above "TBD" section.

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,303 @@
...
+keywords: LoopBack 4.0, LoopBack 4
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Testing-your-extension.html
+summary:
+---
+
+## Overview
+
+An extension for LoopBack 4 is likely to be used by thousands and as such it's
```
kjdelisle: I'd rephrase this to not specify a numeric quantity, since it comes across as pretentious.
Reading that, it sounds like:
"You must make code that we think is good because we're popular and we don't want you to make us look bad."

pages/en/lb4/Testing-Your-Extensions.md
kjdelisle: `because it`

All of the bullet points should be natural extensions of the leading sentence.

`because ensure your extension works correctly`
vs
`because it ensures your extension works correctly`

All that said, tests don't necessarily ensure that your extension works correctly, only that the existing user cases you've turned into tests are being satisfied.

kjdelisle (changes requested):
I'm on the move, so I'll review more soon, but I've left a couple comments.

pages/en/lb4/Testing-Your-Extensions.md
virkt25: I'm hesitant to do that because integration tests for an extension might be different than those for a application. I also don't want to be too prescriptive because a user should be the one deciding how / what tests they want to write for their extension

pages/en/lb4/Testing-Your-Extensions.md
virkt25: I definitely meant actors (referring to dependencies) ... but I can change it if it doesn't make sense.

pages/en/lb4/Testing-Your-Extensions.md
virkt25: ~Switching to `to` instead of `because` cause the points become natural extensions that way. Rolls off the tongue smoother~

Switched it to `because it` since I didn't read `to` in full context of the sentence

bschrammIBM (approved):
I edited the text for readability. I approve with my suggested edits and any additional edits suggested by the team.

pages/en/lb4/Testing-Your-Extensions.md
```diff
@@ -0,0 +1,281 @@
...
+keywords: LoopBack 4.0, LoopBack 4
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Testing-your-extension.html
+summary:
+---
+
+## Overview
+
+LoopBack 4 extensions are often used by other teams. For this reason, it is important to have a
```
kjdelisle: I'd say that this sentence is weak; it doesn't really tie together the idea that an extension being used by _anyone_ is a reason to test it.

IMO, this should just say something like:
```
A thorough test suite for your extension brings powerful benefits, like:
- validating extension behavior
- preventing unwanted changes to the API or functionality of your extension
- providing working samples and snippets of code that serve as functional documentation to your users
```

The exact phrasing of the bullet points doesn't have to change since the current list still captures the same ideas, but you'll want to reword them slightly to fit with the change to the previous sentence.

kjdelisle (approved):
I'd prefer if you implemented that minor change, but it's not a blocker.
LGTM

pages/en/lb4/Testing-Your-Extensions.md
bschrammIBM: I rewrote the intro slightly differently, but very close to K's text above:

LoopBack 4 extensions are often used by other teams. A thorough test suite for your extension brings powerful benefits to all your users, including:

* Validating the behavior of the extension
* Preventing unwanted changes to the API or functionality of the extension
* Providing working samples and code snippets that serve as functional documentation for your users

PR #548, opened 2017-11-27
Add controller CLI

shimks:
Docs to go with CLI creating an empty controller template
- connected to https://github.com/strongloop/loopback-next/pull/762
- connected to https://github.com/strongloop/loopback-next/issues/738

### DIFF
--- a/pages/en/lb4/includes/CLI-std-options.md
+++ b/pages/en/lb4/includes/CLI-std-options.md
@@ -1,0 +1,11 @@
+{% if include.title != "no" %}
+**Standard options**
+{% endif %}
+`-h, --help`
+: Print the generator's options and usage.
+
+`--skip-cache`
+: Do not remember prompt answers. Default is false.
+
+`--skip-install`
+: Do not automatically install dependencies. Default is false.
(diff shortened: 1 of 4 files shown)


pages/en/lb4/includes/CLI-std-options.md
```diff
@@ -0,0 +1,11 @@
+{% if include.title != "no" %}
+**Standard options**
+{% endif %}
+`-h, --help`
```
dhmlau, line 4: Do we need an extra line? 
When I was viewing it after clicking the "View" button, there's no new line before the `-h`.

dhmlau (approved):
Please make sure the format / spacing is correct.  Other than that, LGTM.

pages/en/lb4/includes/CLI-std-options.md
shimks: I copied the file from the lb3 folder, so if we need a new line before the `-h`, we'd need to change both files. The code snippet works in loopback.io without formatting issue if that's what you're concerned about.

dhmlau: ok. If it works fine in loopback.io, then it's ok.  The "View" in github doesn't always show exactly on the web site.

bschrammIBM (commented):
I am sorry, but I don't have time to parse through the history of this issue. If the rest of the team approves it, then I am fine with it.

Issue #434, opened 2017-08-05
extend operational hook examples


atherdon:
## URL of the page 
https://loopback.io/doc/en/lb3/Operation-hooks.html#before-delete
This section don't have an examples, that suits to situation when you want to change things in related models (simple version of 'cascade' delete).

## Suggested resolution
I propose to extend this section with example from [link](https://stackoverflow.com/questions/28607543/how-to-access-the-modal-instances-that-will-be-deleted-in-the-before-delete)
I think this will help other researches. For me it was very helpful

Thanks

crandmck:
Yes, please open a pull request.  I think an example of a cascading delete would be great here.

crandmck:
@atherdon ping...?

lancedouglas1:
Submitted a PR with a simple "see also" link: https://github.com/strongloop/loopback.io/pull/570

PR #570, opened 2017-12-19
Issue 434

lancedouglas1:
Linking to StackOverflow example of a nested delete.
https://github.com/strongloop/loopback.io/issues/434

### DIFF
--- a/pages/en/lb3/Operation-hooks.md
+++ b/pages/en/lb3/Operation-hooks.md
@@ -744,6 +744,10 @@ if (subscriptions.length > 0) {
 }
 ```
 
+{% include see-also.html content="
+Interesting extended example that performs a nested delete: https://stackoverflow.com/a/28990626/3038071
+" %}
+
 ### after delete
 
 {% include important.html content="


slnode:

bschrammIBM:
Assigning to Eng team for a review.

dhmlau:
@bschrammIBM , quick question.. Do we usually directly reference posts from SO in official documentation?

bschrammIBM:
No it does not appear that there are other direct references to an example on www.StackOverflow.com. IMO not a good idea as we do not have control over the referenced page. There is, however, a reference to SO as a place to ask questions on this page: https://loopback.io/doc/en/contrib/Reporting-issues.html and as a community forum for lb on this page: https://loopback.io/doc/en/contrib/

In this case, could we rewrite the example to include it in the loopback.io?

dhmlau:
@lancedouglas1 , per comment from @bschrammIBM , it would be a better approach if we could get a snippet of the example instead of referencing a link in the documentation.  Please feel free to submit a PR for the concrete example.  
With this reason, I'm closing this PR.  Thanks.

PR #571, opened 2017-12-20
fix small errors in Key Concepts

shimks:
- fixed some obvious mistakes and outdated code examples in Key Concepts, so that when the code is copy-pasted it works.

### DIFF
--- a/pages/en/lb4/Sequence.md
+++ b/pages/en/lb4/Sequence.md
@@ -13,30 +13,6 @@ summary:
 A `Sequence` is a stateless grouping of [Actions](#actions) that control how a
 `Server` responds to requests.
 
-```js
-const server = await app.getServer(RestServer);
-server.handler((sequence, request, response) => {
-  const spec = {parameters: {name: 'string', source: 'query'}};
-  const params = sequence.parseParams(request, spec);
-  sequence.send(response, `hello ${params.name}`);
-});
-```
-
-In the following example, we define a `handler` function that uses the
-`Sequence` to respond to every HTTP request our server instance receives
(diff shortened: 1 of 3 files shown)


pages/en/lb4/Sequence.md
```diff
@@ -23,10 +47,10 @@ instances handle requests and responses. The `DefaultSequence` looks like this:
 -->
 ```js
 class DefaultSequence {
-  async handle(request: ParsedRequest, response: ServerResponse) {
+  async handle(request, response) {
```
b-admike: any reason for removing the types here?

shimks: This is weird, the diffs are switched; what's in red is what i introduced and what's in green is stuff that I changed. Originally, I accidentally pushed a commit onto the master branch, so I ended up reverting that and then introduced this PR. I'll see if I can fix this.

pages/en/lb4/Sequence.md
```diff
@@ -13,30 +13,6 @@ summary:
 A `Sequence` is a stateless grouping of [Actions](#actions) that control how a
 `Server` responds to requests.
 
-```js
-const server = await app.getServer(RestServer);
```
b-admike, line 17: was this duplicate code?

shimks: The code didn't work when I tried to copy paste it; the spec given in `sequence.handler` has the wrong type to be a `parseParams` parameter supposedly, and it seems like the code was refactored so that this wouldn't work anymore.

bschrammIBM (approved):

PR #587, opened 2018-01-20
Important documentation change to avoid mistakes.

TwistedLogic:
When passing the host without a port the connector isn't able to create instances of redis client.

### DIFF
--- a/pages/en/lb3/Redis-connector.md
+++ b/pages/en/lb3/Redis-connector.md
@@ -67,7 +67,10 @@ Edit `datasources.json` to add other properties that enable you to connect the
     <tr>
       <td>host</td>
       <td>String</td>
-      <td>Database host name. When this property is set, also the property **port** needs to be set.</td>
+      <td>
+        Database host name.
+        When this property is set, also the property **port** needs to be set.
+      </td>
     </tr>
     <tr>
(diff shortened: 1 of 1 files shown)


bschrammIBM:
Thanks for this addition - I am forwarding it to the Eng team to review.
Can you also fix your Commit message so the PR passes all checks?
1 problems found:
**    4eba5a9 - Important documentation change to avoid mistakes.: line 3 longer than 72 characters.
**

dhmlau:
@TwistedLogic , thanks for submitting the PR.  While I agree that the `port` is mandatory if `host` is used, I believe it is the same for the other properties `username`, `password` and `database` as well.    
I'm not familiar with Redis connector.  But if it works similarly as other connectors like mongodb and cloudant connectors, we'll need to provide all the information that is not included in the `url` property.   Are you thinking along the same line or I misinterpret the intent of your PR?

TwistedLogic:
@dhmlau the connector doesn't manage settings very well when you pass the `host` but not the `port`. I already submitted a PR to fallback to the default port in case only the `host` was given.

`username` isn't managed at all. The `password` field is optional, depending on you Redis auth configuration. I also believe that `database` param should be mandatory, now the connector treats it as optional.

PR #588, opened 2018-01-21
Improve context docs

raymondfeng:
Doc updates for `@loopback/context`.

### DIFF
--- a/pages/en/lb4/Creating-decorators.md
+++ b/pages/en/lb4/Creating-decorators.md
@@ -1,0 +1,13 @@
+---
+lang: en
+title: 'Creating decorators'
+keywords: LoopBack 4.0, LoopBack 4
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Creating-decorators.html
+summary: A tutorial to use `@loopback/metadata` module to create new decorators
+---
+
+[Create decorators](https://github.com/strongloop/loopback-next/blob/master/packages/metadata/README.md)
+
+---
--- a/pages/en/lb4/Creating-decorators.md
+++ b/pages/en/lb4/Creating-decorators.md
@@ -2,12 +2,10 @@
 lang: en
 title: 'Creating decorators'
 keywords: LoopBack 4.0, LoopBack 4
+layout: readme
+source: metadata
 tags:
 sidebar: lb4_sidebar
 permalink: /doc/en/lb4/Creating-decorators.html
 summary: A tutorial to use `@loopback/metadata` module to create new decorators
----
-
-[Create decorators](https://github.com/strongloop/loopback-next/blob/master/packages/metadata/README.md)
-
 ---
--- a/pages/en/lb4/Decorators.md
+++ b/pages/en/lb4/Decorators.md
@@ -237,6 +237,57 @@ class WidgetController {
 }
 ```
 
+A few variants of `@inject` are provided to declare special forms of dependencies:
+
+- @inject.getter: inject a getter function that returns a promise of the bound value of the key
+
+```ts
+class HelloController {
+  constructor(
+    @inject.getter('authentication.currentUser')
+    private userGetter: Getter<UserProfile>) {}
+
+  async greet() {
+    const user = await this.userGetter();
+    return `Hello, ${user.name}`;
+  }
+}
+```
+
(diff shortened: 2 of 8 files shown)


pages/en/lb4/Decorators.md
```diff
@@ -237,6 +237,57 @@ class WidgetController {
 }
 ```
 
+A few variants of `@inject` are provided to declare special forms of dependencies:
+
+- @inject.getter: inject a getter function that returns a promise of the bound value of the key
```
bajtos: Missing backticks?

```markdown
 - `@inject.getter`: inject a getter function etc.
```

pages/en/lb4/Decorators.md
```diff
@@ -237,6 +237,57 @@ class WidgetController {
...
+    private userGetter: Getter<UserProfile>) {}
+
+  async greet() {
+    const user = await this.userGetter();
+    return `Hello, ${user.name}`;
+  }
+}
+```
+
+- @inject.setter: inject a setter function to set bound value of the key
```
bajtos: Missing backticks?

```markdown
 - `@inject.setter`: inject a setter function etc.
```

pages/en/lb4/Decorators.md
```diff
@@ -237,6 +237,57 @@ class WidgetController {
...
+    @inject.setter('greeting')
+    private greetingSetter: Setter<string>) {}
+
+  greet() {
+    greetingSetter('my-greeting-message');
+  }
+}
+```
+
+- @inject.tag: inject an array of values by a pattern or regexp to match bindng tags
```
bajtos: ditto - missing backticks

pages/en/lb4/Creating-decorators.md
```diff
@@ -0,0 +1,13 @@
...
+lang: en
+title: 'Creating decorators'
+keywords: LoopBack 4.0, LoopBack 4
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Creating-decorators.html
+summary: A tutorial to use `@loopback/metadata` module to create new decorators
+---
+
+[Create decorators](https://github.com/strongloop/loopback-next/blob/master/packages/metadata/README.md)
```
bajtos: An empty page with a single link to external README is suboptimal. Could you please use the mechanism we have already in place for sourcing content from README files?

See http://loopback.io/doc/en/contrib/Including-READMEs.html and http://loopback.io/doc/en/contrib/readmes.html

bajtos (changes requested):
Nice!

pages/en/lb4/Creating-decorators.md
raymondfeng: Yes, I was trying to do so but the current tooling does not seem to support a monorepo (loopback-next).

bajtos: > the current tooling does not seem to support a monorepo (loopback-next).

Could you please be more specific? What is the current problem and how much work would it be to fix it? If you prefer to land this PR first, then I am ok with that as long as there is a follow-up issue so that we don't forget to fix this problem.

bajtos: Is this discussion actually relevant, considering 7e95ae0 and 934c889?

bajtos (changes requested):
Assuming that the new script `update-v4-readmes.sh` works as intended, you need to update [Jenkinsfile](https://github.com/strongloop/loopback.io/blob/1e7877fb568620b735ad4dcadcaa1c0b3e3062a9/Jenkinsfile) to fetch these new READMEs and commit them to github.

Other than that, the patch LGTM.

raymondfeng:
@bajtos I added v4 readme steps in Jenkinsfile. I'll merge the PR to see it works as expected as I cannot tell it locally.

raymondfeng:
For the record, http://loopback.io/doc/en/lb4/Creating-decorators.html is live now. :-)

bajtos:
> For the record, http://loopback.io/doc/en/lb4/Creating-decorators.html is live now. :-)

👏 

There is one glitch to fix though: the note "This page was generated from the metadata README." is pointing to a wrong GitHub URL: https://github.com/strongloop/metadata

Ideally, the note should say `@loopback/metadata` instead of `metadata`, and point to loopback-next's `packages/metadata` directory.

@raymondfeng ☝️

PR #599, opened 2018-02-01
Adding more detail to the Access Type, property and end-point definitions

frankorama:
Using Loopback, I often forget the syntax of a property when I'm adding relations to my models.  To delete an instance it's `deleteById`, but to delete a relation instance it's `__destroyById`.  I mapped out the relation between access types, properties and their corresponding end-points as a reminder to myself and i hope it can help other users.


frankorama:
Formatted the HTML section property.

bschrammIBM:
Thank you for these edits. I will pass them along t the technical team to review. In the meantime can you please fix the lint checks that failed?

pages/en/lb3/Controlling-data-access.md
```diff
@@ -355,3 +355,165 @@ To upgrade an application from an earlier version of LoopBack 3.x, follow these
...
+    </tr>
+    <tr>
+      <td>&nbsp;</td>
+      <td>findById</td>
+      <td>GET</td>
+      <td>/model-name-plural/{id}</td>
+    </tr>
+    <tr>
+      <td>&nbsp;</td>
+      <td>exists</td>
```
jannyHou, line 382: hi @frankorama , there are two `exists`

pages/en/lb3/Controlling-data-access.md
```diff
@@ -355,3 +355,165 @@ To upgrade an application from an earlier version of LoopBack 3.x, follow these
...
+    </tr>
+    <tr>
+      <td>WRITE</td>
+      <td>upsert</td>
+      <td>PATCH</td>
+      <td>/model-name-plural</td>
+    </tr>
+    <tr>
+      <td>&nbsp;</td>
+      <td>upsert</td>
```
jannyHou, line 442: correct me if I am wrong, my understanding is the `property` field should be unique, but here `upsert` maps to two endpoints: `PATCH/model-name-plural`and `PUT/model-name-plural`
I feel `replace` is the proper word for `PUT/model-name-plural`.

jannyHou (commented):
@frankorama Thank you for contribution.
I left two comments. And can you elaborate more about which code do you generate the table from?
Is it https://github.com/strongloop/loopback/blob/master/lib/model.js#L400 ?

frankorama:
For the generation of the table, it was good old-fashioned elbow-grease; clicking through all the available end-points and looking at the console output while in debug mode with the debug command 'loopback:security:access-context'.

pages/en/lb3/Controlling-data-access.md
frankorama: I just checked in the loopback explorer and there are in fact 2 end-points that map to the `exists` property as described in my documentation

pages/en/lb3/Controlling-data-access.md
frankorama: In loopback 2, both `PATCH` and `PUT` map to the upsert property.  

I'm in the process of migrating our application to Loopback 3 and noticed that the behaviour of PUT has been changed.  I can update the loopback 3 documentation with the new information as soon as I've successfully completed the migration

jannyHou: @frankorama sounds good! I didn't notice it's for loopback 2.

jannyHou (approved):
@frankorama Thank you for adding the map between endpoint and dao method! The PR looks good to me, could you rebase? I will land it for you.

dhmlau:
@jannyHou @virkt25 , since the PR is good, it would be good if we could rebase and fix the commit msg for @frankorama .  thanks.

virkt25:
Thanks for the contribution @frankorama. Your changes have landed! 🎉

PR #600, opened 2018-02-01
docs(lb4): customizing sequence actions

### LINKED ISSUE #863: [Docs]: How to create and register a custom sequence action (e.g. send)
## Overview

We have removed "Configurable response types/formats #436" from MVP because custom responses can be produced via a custom send action. However, our documentation does not explain how to implement such custom send actions or how to register them in an application.

We need to do better! 😄 

See https://github.com/strongloop/loopback-next/issues/436#issuecomment-349073924, https://github.com/strongloop/loopback-next/issues/436#issuecomment-349329054 and https://github.com/strongloop/loopback-next/issues/436#issuecomment-349387123. I am cross-posting the content of the last comment below.

## Acceptance Criteria
- [x] Take the snippet below and convert it into documentation describing the creation of custom sequences
- [x] Update loopback.io with this documentation (http://loopback.io/doc/en/lb4/Sequence.html#using-sequence-actions)
---

Registering your custom send action involves injecting on the `RestBindings.SequenceActions.SEND` key with a custom Provider for Send.

**application.ts**
```ts
export class YourApp extends RepositoryMixin(Application) {
  constructor() {
    super();
    // Assume your controller setup and other items are in here as well.
    this.bind(RestBindings.SequenceActions.SEND).toProvider(CustomSendProvider);
  }
```

**custom-send-provider.ts**
```ts
import { Send } from "@loopback/rest";
import { Provider, BoundValue } from "@loopback/context";
import { writeResultToResponse } from "@loopback/rest";

export class CustomSendProvider implements Provider<BoundValue>{
  value(): Send | Promise<Send> {
    return writeResultToResponse; // Replace this with your own implementation.
  }
}
```

If you're implementing _multiple servers_ in your application, binding this key at the Application-level context will make that binding the default for all of the servers connected to it. This means that if you wanted different custom providers for different RestServer instances, you'd want to perform those bindings elsewhere (like in an async override of the `start` method).

In this example, PrivateSendProvider might allow stack traces to be returned via the response, whereas PublicSendProvider will not.

**in application.ts**
```ts
async start() {
   const publicServer = await this.getServer('public');
   const privateServer = await this.getServer('private');
   publicServer.bind(RestBindings.SequenceActions.SEND).toProvider(PublicSendProvider);
   privateServer.bind(RestBindings.SequenceActions.SEND).toProvider(PrivateSendProvider);
   await super.start(); // Don't forget to call start!
}
```

kjdelisle:
closes https://github.com/strongloop/loopback-next/issues/863

- Change the current section to specify that we're customizing the sequence actions
- Add a brief explanation of what we're doing
- Show sample code
- Brief conclusion about what it lets you do, as well as mentioning that this works for all listed sequence actions

### DIFF
--- a/pages/en/lb4/Sequence.md
+++ b/pages/en/lb4/Sequence.md
@@ -141,11 +141,58 @@ class MySequence extends DefaultSequence {
 }
 ```
 
-### Using Sequence actions
+### Customizing Sequence Actions
 
-{% include content/tbd.html %}
+There might be scenarios where the default sequence _ordering_ is not something
+you want to change, but rather the individual actions that the sequence will
+execute.
 
-How to use non-core Sequence Actions (authenticate, authorize, log, debug, etc).
+To do this, you'll need to override one or more of the sequence action bindings
+used by the `RestServer`, under the `RestBindings.SequenceActions` constants.
+
+As an example, we'll implement a custom sequence action to replace the default
+"send" action. This action is responsible for returning the response from a
+controller to the client making the request.
+
+To do this, we'll register a custom send action by binding a
+[Provider](http://apidocs.strongloop.com/@loopback%2fcontext/#Provider) to the
+`RestBindings.SequenceActions.SEND` key.
+
+First, let's create our `CustomSendProvider` class, which will provide the
+send function upon injection.
+
+{% include code-caption.html content="/src/providers/custom-send-provider.ts" %}
+**custom-send-provider.ts**
+```ts
+import { Send } from "@loopback/rest";
+import { Provider, BoundValue } from "@loopback/context";
+import { writeResultToResponse } from "@loopback/rest";
+
+export class CustomSendProvider implements Provider<BoundValue>{
+  value(): Send | Promise<Send> {
+    return writeResultToResponse; // Replace this with your own implementation.
+  }
+}
+```
+
+Next, in our application class, we'll inject this provider on the
+`RestBindings.SequenceActions.SEND` key.
+
+{% include code-caption.html content="/src/application.ts" %}
+```ts
+export class YourApp extends RepositoryMixin(Application) {
+  constructor() {
+    super();
+    // Assume your controller setup and other items are in here as well.
+    this.bind(RestBindings.SequenceActions.SEND).toProvider(CustomSendProvider);
+  }
+```
+
+As a result, whenever the send action of the
+[`DefaultSequence`](http://apidocs.strongloop.com/@loopback%2frest/#DefaultSequence)
+is called, it will make use of your function instead! You can use this approach
+to override any of the actions listed under the `RestBindings.SequenceActions`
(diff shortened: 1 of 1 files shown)


pages/en/lb4/Sequence.md
```diff
@@ -141,11 +141,58 @@ class MySequence extends DefaultSequence {
...
+  }
+}
+```
+
+Next, in our application class, we'll inject this provider on the
+`RestBindings.SequenceActions.SEND` key.
+
+{% include code-caption.html content="/src/application.ts" %}
+```ts
+export class YourApp extends RepositoryMixin(Application) {
```
b-admike, line 183: I suggest importing `RepositoriyMixin` and `Application` here to be consistent with the snippet above.

jannyHou: And import `CustomSendProvider` as well :-p

pages/en/lb4/Sequence.md
```diff
@@ -141,11 +141,58 @@ class MySequence extends DefaultSequence {
...
+{% include code-caption.html content="/src/providers/custom-send-provider.ts" %}
+**custom-send-provider.ts**
+```ts
+import { Send } from "@loopback/rest";
+import { Provider, BoundValue } from "@loopback/context";
+import { writeResultToResponse } from "@loopback/rest";
+
+export class CustomSendProvider implements Provider<BoundValue>{
+  value(): Send | Promise<Send> {
+    return writeResultToResponse; // Replace this with your own implementation.
```
bajtos, line 173: I am proposing to enhance this code snippet to show our users how to best structure their code.

```ts
export class CustomSendProvider implements Provider<BoundValue>{
  constructor(
    // specify dependencies to inject 
  ) {}

  value(): Send | Promise<Send> {
    return (response, result) => this.action(response, result);
  }

  action(response: Response, result: OperationRetval): void {
    // put your implementation here
    // access dependencies injected to the constructor via "this" properties
  }
}
```

bajtos (approved):
👏

bajtos:
strongloop/loopback-next#863 is mentioning configuration of custom sequence actions in multi-server setup. I know multiple servers are out of scope of MVP, but I think our documentation should at least briefly mention what will happen in such case (all servers will use the same customization) and that instructions for customizing sequence actions per-server instance will be provided later.

pages/en/lb4/Sequence.md
bajtos, line 183: Should we extend `RestApplication` instead of `Application`?

shimks (approved):
LGTM once you address @bajtos 's comments

bschrammIBM (approved):

kjdelisle:
@bschrammIBM I needed to update this pull request with a few changes based on @bajtos 's recommendations. I'll open a follow-up PR to address that feedback instead.

PR #601, opened 2018-02-01
docs(lb4): reorganize current tutorial materials

kjdelisle:
connected to https://github.com/strongloop/loopback-next/issues/729

This is a rework of the existing tutorial materials and layout
that uses the example-getting-started materials as a starting
point for new users.

THIS IS STILL A WORK-IN-PROGRESS!
Feel free to review it anyway. :)

### DIFF
--- a/_data/sidebars/lb4_sidebar.yml
+++ b/_data/sidebars/lb4_sidebar.yml
@@ -15,15 +15,19 @@ children:
   url: Installation.html
   output: 'web, pdf'
 
-- title: 'Getting started'
-  url: Getting-started.html
+- title: 'Examples and tutorials'
+  url: Examples-and-tutorials.html
   output: 'web, pdf'
   children:
 
-  - title: 'Examples and tutorials'
-    url: Examples-and-tutorials.html
+  - title: 'Hello World!'
+    url: Hello-World.html
     output: 'web, pdf'
 
+  - title: 'Getting Started'
+    url: Getting-started.html
+    output: 'web, pdf'
+  
 - title: 'Key concepts'
   url: Concepts.html
   output: 'web, pdf'
--- a/pages/en/lb4/Examples-and-tutorials.md
+++ b/pages/en/lb4/Examples-and-tutorials.md
@@ -8,13 +8,15 @@ permalink: /doc/en/lb4/Examples-and-tutorials.html
 summary:
 ---
 
-LoopBack 4 comes with the following example projects:
+Here's a list of tutorial resources to help you jump into LoopBack 4!
 
-- **[hello-world](https://github.com/strongloop/loopback-next/tree/master/packages/example-hello-world)**:
-  Tutorial on setting up a simple hello-world application using LoopBack 4.
+* **[Hello World](Hello-World.html)**:
+  Tutorial on setting up a simple hello-world application using LoopBack 4 with
+  a REST server and a controller.
 
-- **[getting-started](https://github.com/strongloop/loopback-next/tree/master/packages/example-getting-started)**:
-  Tutorial on building a simple application with LoopBack 4 key concepts.
+* **[Getting Started](Getting-started.html)**:
+  Tutorial on building a Todo API that leverages key LoopBack 4 concepts like
+  controllers, sequences, repositories, datasources, context and more!
 
 - **[log-extension](https://github.com/strongloop/loopback-next/tree/master/packages/example-log-extension)**:
   Tutorial on building a log extension.
@@ -23,7 +25,6 @@ LoopBack 4 comes with the following example projects:
   An example showing how to implement a made-up RPC protocol.
 
 You can download any of the example projects usig our CLI tool `lb4`:
-
 
 ```
 $ lb4 example
--- a/pages/en/lb4/Getting-started.md
+++ b/pages/en/lb4/Getting-started.md
@@ -1,90 +1,12 @@
 ---
+title: "Getting Started"
 lang: en
-title: 'Getting started'
-keywords: LoopBack 4.0, LoopBack 4
-tags:
+layout: readme
+source: loopback-next 
+file: packages/example-getting-started/README.md
+keywords: LoopBack, LoopBack 4 
+tags: [tutorial, getting-started]
 sidebar: lb4_sidebar
 permalink: /doc/en/lb4/Getting-started.html
-summary: Write and run a LoopBack 4 "Hello World" project in JavaScript and TypeScript.
+summary: The introductory tutorial for learning LoopBack 4.
 ---
-## Prerequisites
-
-Install [Node.js](https://nodejs.org/en/download/) (version 8.x.x or higher) if
-not already installed on your machine.
-
-## Install LoopBack 4 CLI
-
-The LoopBack 4 CLI is a command-line interface that can scaffold a project or
-extension with more features under development. CLI provides the fastest way to
-get started with a LoopBack 4 project that adheres to best practices.
-
-Install the CLI globally by running
-```sh
-npm i -g @loopback/cli
-```
-
-## Create a new project
-
-The CLI tool will scaffold the project, configure TypeScript compiler and
-install all the required dependencies. To create a new project, run the CLI as
-follows and answer the prompts.
-```sh
-lb4 app
-```
-
-Answer the prompts as follows:
-```sh
-? Project name: getting-started
-? Project description: Getting started tutorial
-? Project root directory: (getting-started)
-? Application class name: StarterApplication
-? Select project build settings: Enable tslint, Enable prettier, Enable mocha, Enable loopbackBuild
-```
-
-### Starting the project
-
-The project comes with a "ping" route to test the project. Let's try it out by running the project.
-```sh
-cd getting-started
-npm start
-```
-
-In a browser, visit [http://127.0.0.1:3000/ping](http://127.0.0.1:3000/ping).
-
-## Adding your own controller
-
-Now that we have a basic project created, it's time to add our own [controller](Controllers.html).
-Let's add a simple "Hello World" controller as follows:
-
-* Create a new file in `/src/controllers/` called `hello.controller.ts`.
-
-* Paste the following contents into the file:
-  ```ts
-  import {get} from '@loopback/rest';
-
-  export class HelloController {
-    @get('/hello')
-    hello(): string {
-      return 'Hello world!';
-    }
-  }
-  ```
-
-* Update `/src/application.ts` to load the controller:
-    * Import `HelloController` at the top of the file
-      ```ts
-      import {HelloController} from './controllers/hello.controller';
-      ```
-
-    * Add controller in `setupControllers()`
-      ```ts
-      setupControllers() {
-        this.controller(PingController);
-        this.controller(HelloController);
-      }
-      ```
-
-* Start the application using `npm start`.
-    * *Note: If your application is still running, press **CTRL+C** to stop it before restarting it*
-
-* Visit [http://127.0.0.1:3000/hello](http://127.0.0.1:3000/hello) to see `Hello world!`
--- a/pages/en/lb4/Hello-World.md
+++ b/pages/en/lb4/Hello-World.md
@@ -1,0 +1,99 @@
+---
+lang: en
+title: 'Hello World!'
+keywords: LoopBack 4.0, LoopBack 4
+tags:
+sidebar: lb4_sidebar
+permalink: /doc/en/lb4/Hello-World.html
+summary: Write and run a LoopBack 4 "Hello World" project in JavaScript and TypeScript.
+---
+
+## Prerequisites
+
+Install [Node.js](https://nodejs.org/en/download/) (version 8.x.x or higher) if
+not already installed on your machine.
+
+## Install LoopBack 4 CLI
+
+The LoopBack 4 CLI is a command-line interface that can scaffold a project or
+extension with more features under development. CLI provides the fastest way to
+get started with a LoopBack 4 project that adheres to best practices.
+
+Install the CLI globally by running
+
+```sh
+npm i -g @loopback/cli
+```
+
+## Create a new project
+
+The CLI tool will scaffold the project, configure TypeScript compiler and
+install all the required dependencies. To create a new project, run the CLI as
+follows and answer the prompts.
+
+```sh
+lb4 app
+```
+
+Answer the prompts as follows:
+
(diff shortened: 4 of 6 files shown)


kjdelisle:
Right now, the big problem I see with this tutorial is that developers will need a REST client to test it out because we don't bundle a front-end with it.

However, we pulled the static file server from the MVP, so I'm alright with keeping the tutorial as a back-end only affair for now, with a view to using a front-end in the tutorial before we release.

pages/en/lb4/readmes/example-getting-started.md
```diff
@@ -0,0 +1,424 @@
...
+
+{% include code-caption.html content="src/application.ts" %}
+
+```ts
+import {Application, ApplicationConfig} from '@loopback/core';
+import {RestComponent} from '@loopback/rest';
+import {PingController} from './controllers/ping-controller';
+import {Class, Repository, RepositoryMixin} from '@loopback/repository';
+
+export class TodoApplication extends RepositoryMixin(Application) {
```
shimks: Could you please refactor the tutorial to use `RestApplication`? Refer to the latest `master` branch for consistency.

shimks: I've updated the README in this PR https://github.com/strongloop/loopback-next/pull/961, so you can just copy paste the diffs

pages/en/lb4/readmes/example-getting-started.md
```diff
@@ -0,0 +1,424 @@
...
+import {
+  Class,
+  Repository,
+  RepositoryMixin,
+  DataSourceConstructor,
+} from '@loopback/repository';
+import {db} from './datasources/db.datasource';
+import {TodoRepository} from './repositories';
+
+export class TodoApplication extends RepositoryMixin(Application) {
```
shimks: Same as above

shimks (commented):
Haven't taken a detailed look, but is there a for us to see the diff between this and the README in the package itself? Also, do we know of a way somehow keep the tutorial updated as further updates are made to the `example-getting-started` repo? What I'm afraid of is that we'll always forget to update this page as changes to the `example-getting-started` packages are made.

_data/sidebars/lb4_sidebar.yml
```diff
@@ -15,13 +15,13 @@ children:
...
-  url: Getting-started.html
+- title: 'Tutorial'
+  url: Tutorial.html
   output: 'web, pdf'
   children:
 
-  - title: 'Examples and tutorials'
-    url: Examples-and-tutorials.html
+  - title: 'More examples and tutorials'
+    url: More-examples-and-tutorials.html
```
bajtos: This will break any pages referring to the old URL `/More-examples-and-tutorials.html`. Can we preserve the old URL please?

pages/en/lb4/readmes/example-getting-started.md
```diff
@@ -0,0 +1,424 @@
+## Making A Todo App
```
bajtos: I believe the files in `readmes` are automatically fetched from GitHub, see http://loopback.io/doc/en/contrib/Including-READMEs.html and https://github.com/strongloop/loopback.io/blob/e26a44be2e910740a4eaea2927845b46812a2fd1/update-v4-readmes.sh.

IMO, you need to make these changes in our main monorepo - https://github.com/strongloop/loopback-next/blob/a2368cee598c131a826ee42e347266c6e25ae582/packages/example-getting-started/README.md

If you don't want this content to be fetched from example-getting-started's README, then please put the content to a different file.

Unless I am missing something here?

bajtos (changes requested):

_data/sidebars/lb4_sidebar.yml
kjdelisle: My reason for renaming this URL is that preserving the current name makes for some confusing UX.
The navbar ends up looking like this:
```
+ Tutorial
+ Examples and Tutorials
```

If we're going to separate them to distinguish between our default introductory tutorial and our other resources, we should make sure there's no confusion about what is what.

What if we rename the route to something like "Additional Learning"?

pages/en/lb4/readmes/example-getting-started.md
kjdelisle: I'm not exactly sure how it's supposed to work, but I know that the current template we use for transplanting the READMEs into the documents is currently hardcoded to path towards `strongloop/{repo-name}`, which won't work for our use-case. It looks like we're refreshing those doc pages by _committing the updates here_ rather than scraping them dynamically, since we have a whole folder full of READMEs for those connectors in the loopback.io repository: https://github.com/strongloop/loopback.io/tree/gh-pages/pages/en/lb3/readmes

Sidenote: It also looks like many of them haven't been updated in forever; was this script run manually by Rand or someone else on a regular basis?

Regardless, I wouldn't expect there to already be something in place that copies READMEs over from lb4, and until there is, we need to make sure that there's something in place here.

kjdelisle: Quick addendum: I do think that figuring out an automatic refresh of these documents falls within MVP scope, but only if everyone agrees:

@bajtos @raymondfeng @dhmlau

shimks (approved):

b-admike (approved):

_data/sidebars/lb4_sidebar.yml
```diff
@@ -15,13 +15,13 @@ children:
   url: Installation.html
   output: 'web, pdf'
 
-- title: 'Getting started'
```
dhmlau, line 18: Are we removing the Getting Started?  I kind of like it though.  
The garden path tutorial might be able to guide people through.  But the Getting Started can at least quickly show what LB can do? 

Not sure what you're planning for the tutorial.  If it's simple enough for people to "Get started", i'm good with removing the getting started.

_data/sidebars/lb4_sidebar.yml
dhmlau: What about `other examples and tutorials`? 
Or rename the `Tutorial` to something else? e.g. step-by-step tutorial?

pages/en/lb4/readmes/example-getting-started.md
dhmlau: I remember some of the READMEs in LB3 got copied over to the official documentation.  Not sure how easy/difficult to do that.  Perhaps @bschrammIBM can help us?

pages/en/lb4/readmes/example-getting-started.md
```diff
@@ -0,0 +1,422 @@
...
+" %}
+
+Let's get started building our very first application using LoopBack 4!
+We're going to build a Todo application that provides a CRUD API for Todo lists!
+
+## Prerequisites
+
+Before we can begin, you'll need to make sure you have some things installed:
+
+* [Node.js](https://nodejs.org/en/) at v6.x or greater
```
dhmlau: If we're dropping Node 6, it would be v8 or higher?

_data/sidebars/lb4_sidebar.yml
bajtos: IMO, the URL does not necessarily have to be the same as the page title, e.g. we can have "More examples and tutorials" at `lb4/Examples-and-tutorials.html`. Although I agree it's better when they match each other.

The reason why I am concerned about this change is that my upcoming blog post is referring to `Examples-and-tutorials.html` URL and I'd rather avoid synchronization between landing my PR adding the blog post and this PR breaking old URLs.

On the second thought, I think we should simply leverage [jekyll-redirect-from plugin](https://github.com/jekyll/jekyll-redirect-from#usage) that's already configured for loopback.io and add redirect from `Examples-and-tutorials.html` to `More-examples-and-tutorials.html`.

pages/en/lb4/readmes/example-getting-started.md
bajtos: We are already fetching some READMEs from loopback-next repository, see https://github.com/strongloop/loopback.io/pull/588 for an example how to do that, and also Raymond's recent PR https://github.com/strongloop/loopback.io/pull/608 that's fixing some aspects of this process.

> Sidenote: It also looks like many of them haven't been updated in forever; was this script run manually by Rand or someone else on a regular basis?

I believe there is a Jenkins pipeline (see [Jenkinsfile](https://github.com/strongloop/loopback.io/blob/gh-pages/Jenkinsfile) in this repo) that is supposed to update README whenever they change. I think it's based on some GitHub web hooks, but I am not sure. @rmg could you please clarify?

> Regardless, I wouldn't expect there to already be something in place that copies READMEs over from lb4, and until there is, we need to make sure that there's something in place here.

Uff, I am don't understand, could you please rephrase your sentence for more clarity?

Since we already have some tooling for including READMEs from loopback-next packages, I think we should leverage it for your work too, even if this process has flaws right now. At worst, you can run [update-v4-readmes.sh](https://github.com/strongloop/loopback.io/blob/gh-pages/update-v4-readmes.sh) manually to ensure the copy of example-getting-started's README is stored in this repo.

If you disagree and think this is more work that we want to invest for MVP, then please apply the second approach I proposed:

> If you don't want this content to be fetched from example-getting-started's README, then please put the content to a different file.

I really don't want to create a situation when it looks like `pages/en/lb4/readmes/example-getting-started.md` is something that's supposed to be automatically mirroring content of `example-getting-started/README.md`, while it is not a mirror but a completely different content instead.

_data/sidebars/lb4_sidebar.yml
virkt25: +1 to keeping Getting started as the page describing the process of ensure Node 8 + TypeScript is installed (does this need to be installed?), installing the CLI, and creating an app (Just show the built-in `ping controller`). A bare bones, I made a thing with LB4. 

Tutorial imo then should go into a more complex app.

pages/en/lb4/Examples-and-tutorials.md
```diff
@@ -4,26 +4,23 @@ title: Examples and tutorials
...
 
-- **[hello-world](https://github.com/strongloop/loopback-next/tree/master/packages/example-hello-world)**:
+* **[hello-world](https://github.com/strongloop/loopback-next/tree/master/packages/example-hello-world)**:
   Tutorial on setting up a simple hello-world application using LoopBack 4.
 
-- **[getting-started](https://github.com/strongloop/loopback-next/tree/master/packages/example-getting-started)**:
-  Tutorial on building a simple application with LoopBack 4 key concepts.
-
-- **[log-extension](https://github.com/strongloop/loopback-next/tree/master/packages/example-log-extension)**:
+* **[log-extension](https://github.com/strongloop/loopback-next/tree/master/packages/example-log-extension)**:
```
virkt25: Can we add headers for each one showing what the tutorial is for. `## Application`, `## Extensions`, `## Servers`

_data/sidebars/lb4_sidebar.yml
virkt25: How about making `Examples and Tutorials` a top level item listing all examples / tutorials including the Garden Path tutorial ... and under it the first item can be the Garden Path Tutorial. We can link directly to the `Tutorial` page from `Getting-Started`?

_data/sidebars/lb4_sidebar.yml
virkt25: I think the Getting Started page can / should fold in the installation page.

kjdelisle:
The issue I have with the "Getting Started" section is that it conflicts with the `example-getting-started` package, which is not a "Hello World" example, but rather a full application tutorial.

IMO, we have two options:
**Option A**
- Rename the entire `example-getting-started` package to something like `example-todo`
- Use the "Getting Started" section as the introduction to installation *AND* the "Hello World" tutorial (how it's currently written without this PR's changes)
- Add a new section under the "Examples and Tutorials" header for the new todo tutorial (with auto-updated example linkup)

**Option B**
- Keep the `example-getting-started` package as it is
- Rename the "Getting Started" section to "Hello World"
- Have a new and separate "Getting Started" section immediately afterwards that is the full application tutorial.

I'm heavily leaning towards Option A in this case, even though it's a considerably larger effort because it:
- Allows us to keep the useful "Getting Started" wording for the actual idea of...you know...getting started.
- Makes it clearer what the example package in question actually does (it's a Todo app!)

Option B's obvious advantage is that we don't need to refactor code, but I think it isn't as clear as we'd like it to be and clarity should be our primary goal with any sort of tutorials or documentation!

shimks:
I'm for option 1 as well, reason being that the naming convention would actually reflect the repositories' purposes with that option.

Additionally, I'm thinking that we rename the repo to `example-tutorial` to make it clear even in the github that it's meant to be THE tutorial. I'm also thinking that the docs for `example-getting-started` should have their titles changed to to have the word 'tutorial' in them to really hammer in the point that THIS section is THE TUTORIAL

virkt25:
+1 for Option A

raymondfeng:
+1 to rename `example-getting-started` to `example-todo`. The term `getting-started` is much more opinionated and it should be tied to the name of our examples.

dhmlau:
+1 on renaming and `Getting Started` content.

> Add a new section under the "Examples and Tutorials" header for the new todo tutorial 

Is the new todo tutorial the garden path tutorials? because i'm wondering what's so special of that tutorial that we need to make a new section instead of putting it on the same page.

bajtos:
I don't have a strong opinion, I am happy to follow the crowd.

kjdelisle:
Once https://github.com/strongloop/loopback-next/pull/1001 lands, I'll be able to refactor this PR to include the documentation from that package as individual pieces, and then wrap that content with appropriate links back and forth between them.

_data/sidebars/lb4_sidebar.yml
dhmlau: ah. just realize @virkt25 has the same comment! :)
see https://github.com/strongloop/loopback-next/issues/1004

_data/sidebars/lb4_sidebar.yml
```diff
@@ -15,15 +15,19 @@ children:
...
   output: 'web, pdf'
   children:
 
-  - title: 'Examples and tutorials'
-    url: Examples-and-tutorials.html
+  - title: 'Hello World!'
+    url: Hello-World.html
     output: 'web, pdf'
 
+  - title: 'Getting Started'
```
dhmlau, line 27: I'd actually prefer Getting Started being at the top level.  That's usually the first place to look when people are starting out.

virkt25:
Closing as no longer needed.

Issue #605, opened 2018-02-05
Broken link on https://loopback.io/doc/en/lb4/Decorators.html


bschrammIBM:
## URL of the page 

https://loopback.io/doc/en/lb4/Decorators.html

## Nature of the issue 
There is a broken link to: https://loopback.io/lb3/Model-definition-JSON-file.html 
The link is in italics below.

Model Decorator
Syntax: @model(definition?: ModelDefinitionSyntax)

Model decorator is a Class decorator. In LoopBack 4, we inherit the model definition format from LoopBack 3, you can find it in *Model definition JSON* file. For usage examples, see Define Models

## Expected behavior

Link would open the correct page

## Actual behavior

page not found

## Suggested resolution

Is there another url or does the topic not exist?
Need to change link or create the page.

bschrammIBM:
I found the topic in LB3, this is the correct url: https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html
Need to check why the link doesn't work
The url needs to updated. Wrong url: https://loopback.io/lb3/Model-definition-JSON-file.html

shimks:
Issue is solved with #606

PR #606, opened 2018-02-05
fix broken link in decorators

shimks:
- connected to #605

### DIFF
--- a/pages/en/lb4/Decorators.md
+++ b/pages/en/lb4/Decorators.md
@@ -354,7 +354,7 @@ For more information, see the [Dependency Injection](Dependency-Injection.htm) s
 
   Model decorator is a Class decorator.
   In LoopBack 4, we inherit the model definition format from LoopBack 3,
-  you can find it in [Model definition JSON file](/lb3/Model-definition-JSON-file.html).
+  you can find it in [Model definition JSON file](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html).
   For usage examples, see [Define Models](Repositories.html#define-models)
 
   *Please note we will elaborate more about model and model definition in #Model.htm,*


virkt25 (approved):

bschrammIBM:
Thank you! Merging...

pages/en/lb4/Decorators.md
```diff
@@ -354,7 +354,7 @@ For more information, see the [Dependency Injection](Dependency-Injection.htm) s
 
   Model decorator is a Class decorator.
   In LoopBack 4, we inherit the model definition format from LoopBack 3,
-  you can find it in [Model definition JSON file](/lb3/Model-definition-JSON-file.html).
+  you can find it in [Model definition JSON file](https://loopback.io/doc/en/lb3/Model-definition-JSON-file.html).
```
bajtos, line 357: I think we should be using relative paths. That way the links will keep working after we clone content, e.g. as had been done in the past when we created `lb3` as a copy of `lb2`.

```diff
-  you can find it in [Model definition JSON file](/lb3/Model-definition-JSON-file.html).
+  you can find it in [Model definition JSON file](Model-definition-JSON-file.html).
```

shimks: I thought that we should be using relative paths as well. But since appending `/lb3` breaks the link, should we be creating a copy of the file for `lb4`?

virkt25: Can you try `../lb3/...`

shimks: Wow, that worked. I'll be creating a PR to apply this fix

bajtos: Oh, I did not realize this is a `lb4` file referring to `lb3` content. I like @virkt25's proposal 👍

PR #607, opened 2018-02-05
fix(lb4): update custom sequence action example

kjdelisle:
continues the work of #600

### DIFF
--- a/pages/en/lb4/Sequence.md
+++ b/pages/en/lb4/Sequence.md
@@ -161,26 +161,72 @@ send function upon injection.
 {% include code-caption.html content="/src/providers/custom-send-provider.ts" %}
 **custom-send-provider.ts**
 ```ts
-import { Send } from "@loopback/rest";
-import { Provider, BoundValue } from "@loopback/context";
-import { writeResultToResponse } from "@loopback/rest";
+import {Send, ServerResponse} from "@loopback/rest";
+import {Provider, BoundValue, inject} from "@loopback/context";
+import {writeResultToResponse, RestBindings} from "@loopback/rest";
 
-export class CustomSendProvider implements Provider<BoundValue>{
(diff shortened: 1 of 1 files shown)


bschrammIBM (approved):

pages/en/lb4/Sequence.md
```diff
@@ -161,26 +161,72 @@ send function upon injection.
...
+   * @param ServerResponse response The response object used to reply to the
+   * client.
+   * @param OperationRetVal result The result of the operation carried out by
+   * the controller's handling function.
+   */
+  action(response: ServerResponse, result: OperationRetVal) {
+    if (result) {
+      // Currently, the headers interface doesn't allow arbitrary string keys!
+      const headers = this.request.headers as any || {};
+      const header = headers.accept || 'application/json';
```
bajtos, line 196: Using names that differ only by the last letter (`headers` and `header`) is an anti-pattern to be avoided, it is too easy to confuse these two names for each other.

```ts
const requestHeaders = this.request.headers as any || {};
const accepts = requestHeaders.accept || 'application/json';
```

pages/en/lb4/Sequence.md
```diff
@@ -161,26 +161,72 @@ send function upon injection.
...
+      response.setHeader('Content-Type', header);
+      response.end(formattedResult);
+    } else {
+      response.end();
+    }
   }
 }
 ```
 
+Our custom provider will automatically read the `Accept` header from the request
```
bajtos, line 208: Please not that `Accept` header can contain multiple entries, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept:

```
Accept: text/html, application/xhtml+xml, application/xml;q=0.9, */*;q=0.8
```

Please consider using [accepts](https://www.npmjs.com/package/accepts) module to negotiate the best content type.

PR #609, opened 2018-02-06
fix: use relative paths

shimks:
- fixed previous links using hard paths to use relative-paths so that they link to local pages


virkt25 (approved):

bschrammIBM (approved):

bschrammIBM (changes requested):
need to resolve conflicts, as you are probably aware.

PR #610, opened 2018-02-07
Add docs for context module to reflect latest features

raymondfeng:
See https://github.com/strongloop/loopback-next/pull/946

### DIFF
--- a/pages/en/lb4/Context.md
+++ b/pages/en/lb4/Context.md
@@ -26,10 +26,23 @@ builtin/in-memory storage mechanism).
 - You have full access to updated/real-time application+request state at all
 times.
 
-LoopBack's context system allows an unlimited amount of Context instances,
-each of which may have a parent Context.
+## How to create a context?
 
-Typically, however, an application will have three "levels" of context: application-level, server-level and request-level.
+A context can be created with an optional parent and an optional name. If the
+name is not provided, a UUID will be generated as the value. Context instances
+can be chained using the `parent` to form a hierarchy. For example, the code
+below creates a chain of three contexts: `reqCtx -> serverCtx -> rootCtx`.
+
+```ts
+import {Context} from '@loopback/context';
+const rootCtx = new Context('root-ctx'); // No parent
+const serverCtx = new Context(rootCtx, 'server-ctx'); // rootCtx as the parent
+const reqCtx = new Context(serverCtx); // No explicit name, a UUID will be generated
+```
+
+LoopBack's context system allows an unlimited amount of Context instances,
+each of which may have a parent Context. However, an application typically
+has three "levels" of context: application-level, server-level and request-level.
 
 ## Application-level context (global)
 
@@ -42,6 +55,7 @@ Here is a simple example:
 
 ```js
 const Application = require('@loopback/core').Application;
+// Please note `Application` extends from `Context`
 const app = new Application(); // `app` is a "Context"
 class MyController { ... }
 app.controller(MyController);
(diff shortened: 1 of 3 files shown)


bschrammIBM (approved):
approved by me - also need a technical review

pages/en/lb4/Context.md
```diff
@@ -27,7 +27,8 @@ builtin/in-memory storage mechanism).
 times.
 
 LoopBack's context system allows an unlimited amount of Context instances,
-each of which may have a parent Context.
+each of which may have a parent Context. Each context has a name, which can
```
virkt25: Should there be an example showing users how to name a context?

raymondfeng: I was considering that but this page is for the concept. If there is a place to show an example, +1.

virkt25: Perhaps in the README for `@loopback/context`?

virkt25 (approved):

pages/en/lb4/Dependency-injection.md
```diff
@@ -140,28 +140,101 @@ class InfoController {
...
+Sometimes the dependencies are optional. For example, the logging level for a Logger provider can have a default value if it is not set (bound to the context). 
+
+To resolve an optional dependency, set `optional` flag to true:
+
+```ts
+const ctx = new Context();
+await ctx.get('optional-key', {optional: true}); // Return `undefined` instead of throwing an error
+```
+
+Here is anthoer example with dependency injection:
```
bajtos: typo: `anthoer`

pages/en/lb4/Dependency-injection.md
```diff
@@ -140,28 +140,101 @@ class InfoController {
...
+
+```ts
+export class LoggerProvider implements Provider<Logger> {
+  constructor(
+    // Log writer is an optional dependency and it falls back to `logToConsole`
+    @inject('log.writer', {optional: true})
+    private logWriter: LogWriterFn = logToConsole,
+
+    @inject('log.level', {optional: true})
+    private logLevel: string = 'WARN',
```
bajtos: AFAICT, this code example is exactly the same as the code example above. Is this intentional?

Did you perhaps want to show constructor-level and method-level injections as an addition to property-based injection shown previously?

pages/en/lb4/Dependency-injection.md
```diff
@@ -140,28 +140,101 @@ class InfoController {
...
+
+class TeamImpl implements Team {
+  constructor(@inject('project') public project: Project) {}
+}
+
+class ProjectImpl implements Project {
+  constructor(@inject('lead') public lead: Developer) {}
+}
+
+export function main() {
```
bajtos: So far, none of code examples are exporting a `main` function. To be honest, I don't understand why do you need it here anyways, since the body of `main` is synchronous code that can be placed directly in the main scope of the `.js`/`.ts` file.

I am proposing to reorganize this code to avoid `export function main`.

Please let me know if there are good reasons for keeping it, it's entirely possibly I am missing something here.

bajtos (changes requested):
Changes looks pretty good, I have few minor comments to address.

pages/en/lb4/Dependency-injection.md
raymondfeng: See https://github.com/raymondfeng/loopback4-example-di/blob/master/index.js. My demo app allows the user to choose which one to run from the command line.

pages/en/lb4/Dependency-injection.md
raymondfeng: Yes.

raymondfeng:
@bajtos Docs updated to address your comments. PTAL.

pages/en/lb4/Dependency-injection.md
bajtos: I am confused. The sentence before this code block says

> Optional dependencies can also be used with constructor and method injections. For example:

But the code example shows property-level injections only.

Could you please clarify?

raymondfeng: You probably didn't notice the 2nd example has `constructor(...)`.

b-admike (approved):

bajtos (approved):
:shipit:

PR #616, opened 2018-02-09
fix syntax errors in the readme layout

raymondfeng:
There were some syntax errors slipped into the code. This PR fixes them. I have verified with local `loopback.io` site.


bajtos (approved):
😄

PR #617, opened 2018-02-10
Users Page organized with standard image sizes plus added iLoveCoding…

azizali:
… logo


bschrammIBM (approved):
approved, thank you for this improvement!

PR #621, opened 2018-02-15
Move RestServer config out of `async start`

shimks:
This PR would utilize the new changes made in https://github.com/strongloop/loopback-next/pull/994 so that `async start` is not overridden if it can be overridden.

I left a couple of decisions undecided because I wasn't sure whether i should be changing some of these or not. Here are my main concerns:
- In the `WidgetApplication` example under `Application.md`, `server.api(WidgetApi)` is used in start. The function call can't be moved into the constructor since the app is only extending from `Application` and not `RestApplication`; `api()` function can't be called because it doesn't exist for the application. Should the example be left as it is?
- In `Context.md`, we configure multiple servers inside the start function. I think this should be left as is, but does anybody think otherwise?

### DIFF
--- a/pages/en/lb4/Application.md
+++ b/pages/en/lb4/Application.md
@@ -32,7 +32,6 @@ tasks as a part of your setup:
 import {Application} from '@loopback/core';
 import {RestComponent, RestServer} from '@loopback/rest';
 import {SamoflangeController, DoohickeyController} from './controllers';
-import {WidgetApi} from './apidef/';
 
 export class WidgetApplication extends Application {
   constructor() {
@@ -50,16 +49,6 @@ export class WidgetApplication extends Application {
     app.component(RestComponent);
     app.controller(SamoflangeController);
     app.controller(DoohickeyController);
-  }
-
-  async start() {
-    // This is where you would asynchronously retrieve servers, providers and
-    // other components to configure them before launch.
-    const server = await app.getServer(RestServer);
-    server.api(WidgetApi);
-    // The superclass start method will call start on all servers that are
-    // bound to the application.
-    return await super.start();
   }
 
   async stop() {
--- a/pages/en/lb4/Application.md
+++ b/pages/en/lb4/Application.md
@@ -38,7 +38,12 @@ export class WidgetApplication extends Application {
   constructor() {
     // This is where you would pass configuration to the base constructor
     // (as well as handle your own!)
-    super();
+    super({
+      rest: {
+        port: 8080
+      }
+    });
+
     const app = this; // For clarity.
     // You can bind to the Application-level context here.
     // app.bind('foo').to(bar);
@@ -51,7 +56,6 @@ export class WidgetApplication extends Application {
     // This is where you would asynchronously retrieve servers, providers and
     // other components to configure them before launch.
     const server = await app.getServer(RestServer);
-    server.bind('rest.port').to(8080);
     server.api(WidgetApi);
     // The superclass start method will call start on all servers that are
     // bound to the application.
--- a/pages/en/lb4/Application.md
+++ b/pages/en/lb4/Application.md
@@ -184,6 +184,15 @@ export class MyApplication extends RestApplication {
 ## Tips for application setup
 Here are some tips to help avoid common pitfalls and mistakes.
 
+### Extend from `RestApplication` when using `RestServer`
+If you want to use `RestServer` from our `@loopback/rest` package, we recommend you extend
+`RestApplication` in your app instead of manually binding `RestServer` or
+`RestComponent`. `RestApplication` already uses `RestComponent` and makes
+useful functions in `RestServer` like `handler()` available at the app level.
+This means you can call these `RestServer` functions to do all of your
+server-level setups in the app constructor without having to explicitly retrieve
+an instance of your server.
+
 ### Use unique bindings
 Use binding names that are prefixed with a unique string that does not overlap
 with loopback's bindings. As an example, if your application is built for
--- a/pages/en/lb4/Defining-and-validating-the-API.md
+++ b/pages/en/lb4/Defining-and-validating-the-API.md
@@ -289,7 +289,7 @@ _.merge(spec, CategoryAPI);
 export default spec;
 ```
 
-You can then bind the full spec to the application using `server.spec()`. This is done on the server level, because each server instance can expose a different (sub)set of API.
+You can then bind the full spec to the application using `app.api()`. Normally, this is done on the server level because each server instance can expose a different (sub)set of API, but since `RestApplication` uses only one REST server, you can bind the spec at application level.
 
 You also need to associate the controllers implementing the spec with the app using `app.controller(GreetController)`. This is not done on the server level because a controller may be used with multiple server instances, and types!
 
@@ -302,20 +302,18 @@ import { ProductController, DealController, CategoryController } from "./control
 export class YourMicroservice extends RestApplication {
 
   constructor() {
-    super();
+    super({
+      rest: {
+        port: 3001
+      }
+    });
     const app = this;
 
     app.controller(ProductController);
     app.controller(DealController);
     app.controller(CategoryController);
-
-  }
-  async start() {
-    const server = await app.getServer(RestServer);
-    // inject your spec here!
-    server.api(spec);
-    server.bind("rest.port").to(3001);
-    await super.start();
+    //inject your spec
+    app.api(spec);
   }
   // etc...
 }
--- a/pages/en/lb4/Defining-and-validating-the-API.md
+++ b/pages/en/lb4/Defining-and-validating-the-API.md
@@ -289,9 +289,17 @@ _.merge(spec, CategoryAPI);
 export default spec;
 ```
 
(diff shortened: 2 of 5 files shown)


pages/en/lb4/Defining-and-validating-the-API.md
```diff
@@ -284,7 +284,7 @@ _.merge(spec, CategoryAPI);
 export default spec;
 ```
 
-You can then bind the full spec to the application using `server.spec()`. This is done on the server level, because each server instance can expose a different (sub)set of API.
+You can then bind the full spec to the application using `app.api()`. Normally, this is done on the server level because each server instance can expose a different (sub)set of API, but since `RestApplication` uses only one REST server, you can bind the spec at application level.
```
bajtos: I am proposing a slightly different phrasing:

```markdown
You can then bind the full spec to the application using `app.api()`.
This works well for applications with a single REST server, because
there is only one API definition involved.

If you are building an application with multiple REST servers,
where each server provides a different API, then you need 
to call `server.api()` instead.
```

pages/en/lb4/Routes.md
```diff
@@ -74,7 +74,7 @@ server.api(spec);
 
 The example below defines a `Route` that will be matched for `GET /`. When the `Route` is matched, the `greet` Operation (above) will be called. It accepts an OpenAPI [OperationObject](https://github.com/OAI/OpenAPI-Specification/blob/0e51e2a1b2d668f434e44e5818a0cdad1be090b4/versions/2.0.md#operationObject) which is defined using `spec`.
 The route is then attached to a valid server context running underneath the
-application.
+application.
```
bajtos: Trailing space, please remove.

pages/en/lb4/Routes.md
```diff
@@ -96,9 +96,8 @@ function greet(name: string) {
 }
 
 (async function start() {
-  const server = await app.getServer(RestServer);
   const route = new Route('get', '/', spec, greet);
-  server.route(route);
+  app.route(route); // attaches route to RestServer
```
bajtos: Please move route registration out of async `start()` function.

I think we no longer need async start function and can call app.start() directly:

```ts
// greet is a basic operation
 function greet(name: string) {
   return `hello ${name}`;
 }

const route = new Route('get', '/', spec, greet);
app.route(route);

app.start();
```

pages/en/lb4/Sequence.md
```diff
@@ -86,18 +86,16 @@ class MySequence extends DefaultSequence {
...
 ```js
 import {RestApplication, RestServer} from '@loopback/rest';
 
 const app = new RestApplication();
 
-// or
 (async function start() {
-  const server = await app.getServer(RestServer);
-  server.sequence(MySequence);
+  app.sequence(MySequence);
```
bajtos: Ditto, please move this line out of async `start()` function and consider removing the start function completely.

bajtos (changes requested):
This is great!

shimks:
@bajtos May I get another review on this PR so that it could be landed? Thanks

b-admike (approved):
👏

virkt25:
> The function call can't be moved into the constructor since the app is only extending from Application and not RestApplication; api() function can't be called because it doesn't exist for the application. Should the example be left as it is?

I think we should remove this or create a new issue to track the removal. If it's simple just do it in this PR. This way we don't have misleading information in docs that will confuse users and provide a poor experience for anyone following the docs.

shimks:
You're thinking of just removing `server.api(WidgetApi)` entirely right? That would allow us to move things that are inside `start` to the constructor, but then the question becomes what is the `start` function for? What do we advocate users to put inside the `start` function? I'm not sure what this is myself after we decided on getting rid of multi server support.

virkt25:
Yes getting rid of `server.api()` if it doesn't work in the context of that example. 

As for what `start` is, it's not a function user's should need to override ... just call to start their `server(s)`

pages/en/lb4/Application.md
```diff
@@ -32,13 +32,17 @@ tasks as a part of your setup:
 import {Application} from '@loopback/core';
 import {RestComponent, RestServer} from '@loopback/rest';
 import {SamoflangeController, DoohickeyController} from './controllers';
-import {WidgetApi} from './apidef/';
 
 export class WidgetApplication extends Application {
```
bajtos, line 36: IMO, you should extend `RestApplication` here. The only reason why you would not want to, would be if your application was exposing multiple REST servers, which is not the case AFAICT.

This will also allow you to get rid of `app.component(RestComponent)` below.

pages/en/lb4/Application.md
```diff
@@ -47,17 +51,6 @@ export class WidgetApplication extends Application {
     app.controller(DoohickeyController);
   }
 
-  async start() {
-    // This is where you would asynchronously retrieve servers, providers and
-    // other components to configure them before launch.
-    const server = await app.getServer(RestServer);
-    server.bind('rest.port').to(8080);
-    server.api(WidgetApi);
```
bajtos, line 55: This is a breaking change in the context of the original example. The original example assumed api-first approach, where `WidgetApi` contained Swagger spec describing REST API, with extensions like `x-controller-name` and `x-method` used to link Swagger endpoints with controller method implementing them.

Since api-first support is out of scope of our upcoming release, I am ok to rework this example to code-first approach, where the REST API is described on controller classes/methods using decorators like `@get` and `@param`.

In other words, this change is fine with me too 👍

bajtos (approved):
LGTM, but please consider addressing my first comment to simplify the WidgetApplication example.

pages/en/lb4/Application.md
shimks: The only problem I have with that is the page the docs is about: `Application`. IMO an example shown in this page should be vanilla `Application` instead of `RestApplication`. What do you think?

bajtos: Fair enough, let's keep using `Application` then 👍 

Could you perhaps extend [Tips for application setup](https://github.com/strongloop/loopback.io/blob/d903c1b26965386f5aca8092206ca93b98eebb0b/pages/en/lb4/Application.md#tips-for-application-setup) to mention `RestApplication` + when and why to use it? It's ok to leave that out of this pull request though.

pages/en/lb4/Defining-and-validating-the-API.md
```diff
@@ -302,20 +310,18 @@ import { ProductController, DealController, CategoryController } from "./control
...
     app.controller(DealController);
     app.controller(CategoryController);
-
-  }
-  async start() {
-    const server = await app.getServer(RestServer);
-    // inject your spec here!
-    server.api(spec);
-    server.bind("rest.port").to(3001);
-    await super.start();
```
b-admike, line 318: Don't we still need to call `super.start()`?

pages/en/lb4/Routes.md
```diff
@@ -95,12 +94,11 @@ function greet(name: string) {
...
-  const server = await app.getServer(RestServer);
-  const route = new Route('get', '/', spec, greet);
-  server.route(route);
-  await app.start();
-})();
+const app = new RestApplication();
+const route = new Route('get', '/', spec, greet);
+app.route(route); // attaches route to RestServer
+
+app.start();
```
b-admike, line 101: Isn't app.start() returning a promise and thus should be called with `await`? (see https://github.com/strongloop/loopback-next/blob/master/packages/core/src/application.ts#L128-L136)

pages/en/lb4/Sequence.md
```diff
@@ -86,20 +86,16 @@ class MySequence extends DefaultSequence {
...
 const app = new RestApplication();
+app.sequence(MySequencce);
 
-// or
-(async function start() {
-  const server = await app.getServer(RestServer);
-  server.sequence(MySequence);
-  await app.start();
-})();
+app.start();
```
b-admike, line 98: Ditto.

pages/en/lb4/Defining-and-validating-the-API.md
jannyHou: @b-admike I think in this case it's enough to just inherit `start()` function from `server`. No need to override it.

jannyHou (commented):
Same comment as @b-admike : https://github.com/strongloop/loopback.io/pull/621/files#r171920566

Other than that LGTM 👍 awesome clarification of the current behavior of rest server.

pages/en/lb4/Routes.md
raymondfeng: No, that's a myth.

1. `await` can only be used within an `async` function. The `await` keyword helps VM to suspend the execution and wait for the resolution/rejection of the promise or value.

2. In this case, you cannot use `await` as the module will be wrapped into a function by Node.js but it's not `async`.

3. The code before the change declares an `async` function and execute it immediately.

4. It's perfectly fine to call an `async` function or a function returning a promise without `await`. The return value is Promise. You have to manually call `.then() or .catch()` to set up result handler.

b-admike: Thanks for the clarification Raymond. We don't really need to show `.then() or .catch()` handlers in this example. Disregard my comment!

b-admike (approved):

bschrammIBM (approved):

PR #627, opened 2018-02-20
docs: Remove LB4 installation page

dhmlau:
From this github issue https://github.com/strongloop/loopback-next/issues/1004: 
>In the official documentation, we have an installation page: http://loopback.io/doc/en/lb4/Installation.html. I'm not sure whether we need the section for Adding LoopBack 4 to an existing project, because we highly recommend users to scaffold a LB4 app using CLI.

We're going for this option: 
> - Remove the installation page completely
> - the prerequisite of installing Node already exists in Getting Started page.


dhmlau:
Interestingly, it includes another changes that are not in my branch.  Let me refresh and recreate the PR again.  Closing it.

PR #635, opened 2018-02-23
Update Testing your application

shimks:
A question for anyone that's familiar with test databases: should I be creating a separate config for test-specific databases? If so, where should the config file and the typescript file live and how should they be named?

- please review the first commit since the commit after introduces line breaks (https://github.com/strongloop/loopback.io/pull/635/commits/403489f8659992ec2ef398477580df5f4cbe1855)


pages/en/lb4/Testing-your-application.md
```diff
@@ -42,19 +43,15 @@ integrates these packages and makes them easy to use together with LoopBack.
...
 ### Set up testing infrastructure with LoopBack CLI
 
-{% include note.html content="The LoopBack CLI does not yet support LoopBack 4,
-so using the CLI is not an option with the beta release.
-" %}
-
-<!--
-If you are just getting started with LoopBack, use the LoopBack command-line tool (CLI)  `loopback-cli`. It's ready to use and installs simply by `npm install -g loopback-cli`. You don't need to do any extra steps for setup, and can head straight to the [next section](Testing-Your-Application#acceptance-testing.html).
--->
+LoopBack applications that have been generated by `lb4 app` or `lb4`
```
b-admike: You can add here that if the user selects all our add-ons such as `tslint` and `mocha` during app creation, then they get them added as deps too with configs.

pages/en/lb4/Testing-your-application.md
```diff
@@ -20,9 +20,10 @@ A thorough automated test suite is important because it:
...
 
-For an introduction to automated testing, see [Define your testing strategy](Thinking-in-LoopBack.html#define-your-testing-strategy); for a step-by-step tutorial, see [Incrementally implement features](Thinking-in-LoopBack.html#incrementally-implement-features).
+For an introduction to automated testing, see [Define your testing strategy](Defining-your-testing-strategy.html); for a step-by-step tutorial, see [Incrementally implement features](Implementing-features.html).
 
-{% include important.html content="A great test suite requires you to think smaller and favor fast, focused unit tests over slow application-wide end-to-end tests
+{% include important.html content="
+A great test suite requires you to think smaller and favor fast, focused unit tests over slow application-wide end-to-end tests.
```
kjdelisle: `over slow end-to-end tests`

Alternatively,

`over slow, application-wide, end-to-end test`

I would prefer the first, since end-to-end already implies that it covers the whole application, otherwise how could it be "end-to-end"?

pages/en/lb4/Testing-your-application.md
```diff
@@ -331,13 +330,13 @@ describe('Person (unit)', () => {
 });
 ```
 
-Writing a unit test for a custom repository methods is not straightforward because `CrudRepositoryImpl` is based on legacy loopback-datasource-juggler that was not designed with dependency injection in mind. Instead, use integration tests to verify the implementation of custom repository methods; see [Test your repositories against a real database](#test-your-repositories-against-a-real-database) in [Integration Testing](#integration-testing).
+Writing a unit test for a custom repository methods is not as straightforward because `CrudRepository` is based on legacy [loopback-datasource-juggler](https://github.com/strongloop/loopback-datasource-juggler) that was not designed with dependency injection in mind. Instead, use integration tests to verify the implementation of custom repository methods; see [Test your repositories against a real database](#test-your-repositories-against-a-real-database) in [Integration Testing](#integration-testing).
```
kjdelisle: Replace `...that was not designed with...`
with
`...which was not designed with...`

pages/en/lb4/Testing-your-application.md
```diff
@@ -331,13 +330,13 @@ describe('Person (unit)', () => {
...
 
 ### Unit test your Sequence
 
-While it's possible to test a custom Sequence class in isolation, it's better to rely on acceptance-level tests in this exceptional case. The reason is that a custom Sequence class typically has many dependencies (which makes test setup too long and complex), and at the same time it provides very little functionality on top of the injected sequence actions. Bugs are much more likely to caused by the way how the real sequence action implementations interact together (which is not covered by unit tests), instead of the Sequence code itself (which is the only thing covered).
+While it's possible to test a custom Sequence class in isolation, it's better to rely on acceptance-level tests in this exceptional case. The reason is that a custom Sequence class typically has many dependencies (which makes test setup too long and complex), and at the same time it provides very little functionality on top of the injected sequence actions. Bugs are much more likely to be caused by the way the real sequence action implementations interact together (which is not covered by unit tests), instead of the Sequence code itself (which is the only thing covered).
```
kjdelisle: `which makes test setup too long and complex` -> `which can make test setup too long or complex`

kjdelisle (commented):
Ugh. I just spent a decent amount of time reviewing this only to realize that most of what I'm proposing changes to is not actually a part of the changeset.

This is why I would prefer that we have 80 chars per line whenever possible; GitHub's diff wrapping seems to expect it!

shimks:
@bajtos Are we still advocating for attaching util-like functions to the model itself as seen in here? http://loopback.io/doc/en/lb4/Testing-your-application.html#unit-test-your-models-and-repositories

bajtos:
>  This is why I would prefer that we have 80 chars per line whenever possible; GitHub's diff wrapping seems to expect it!

I had the impression that GitHub's markdown renderer preserves line breaks, which is something we don't want 😞Because I could not find any documentation for this, I set out to find out the real behavior myself.

TL;DR: GitHub treats newlines as soft line breaks (i.e. a breakable space) when rendering markdown files (which is great!), but treats newlines as hard line breaks when rendering issue/pull request descriptions and possibly comments too (which I find confusing and often annoying).

https://github.github.com/gfm/#hard-line-breaks
> A line break (not in a code span or HTML tag) that is preceded by two or more spaces and does not occur at the end of a block is parsed as a hard line break (rendered in HTML as a <br /> tag).
>
> For a more visible alternative, a backslash before the line ending may be used instead of two spaces.

https://github.github.com/gfm/#soft-line-breaks
> A regular line break (not in a code span or HTML tag) that is not preceded by two or more spaces or a backslash is parsed as a softbreak. (A softbreak may be rendered in HTML either as a line ending or as a space. The result will be the same in browsers. In the examples here, a line ending will be used.)

A simple Markdown document to verify this behavior can be found here:

 - [preview](https://github.com/strongloop/loopback4-example-microservices/blob/4ff8d0f2e17d62b093039eca7d4d2f80c6a03847/TEST.md)
 - [raw source](https://raw.githubusercontent.com/strongloop/loopback4-example-microservices/4ff8d0f2e17d62b093039eca7d4d2f80c6a03847/TEST.md)

So! Based on this investigation, I fully support your push for 80-char-wide lines in Markdown documents now, @kjdelisle.

bschrammIBM (changes requested):
line 145 If this line is going to remain in the final version, then you need to remove the initials. "shmks" Also I fixed the english a little. (which have not been)
+

minor edit in line 474: (Use "it" not "they")
Therefore, the code outlined in this section is outdated and may not work out of the box. It will be revisited after our MVP release.

line 480 - need to "a" - (a good thing)
Extending your API spec with examples is a good thing on its own, since developers consuming your API will find them useful too.

pages/en/lb4/Testing-your-application.md
```diff
@@ -42,19 +43,17 @@ integrates these packages and makes them easy to use together with LoopBack.
...
 ### Set up testing infrastructure with LoopBack CLI
 
-{% include note.html content="The LoopBack CLI does not yet support LoopBack 4,
-so using the CLI is not an option with the beta release.
-" %}
-
-<!--
-If you are just getting started with LoopBack, use the LoopBack command-line tool (CLI)  `loopback-cli`. It's ready to use and installs simply by `npm install -g loopback-cli`. You don't need to do any extra steps for setup, and can head straight to the [next section](Testing-Your-Application#acceptance-testing.html).
--->
+LoopBack applications that have been generated by `lb4 app` or `lb4`
```
virkt25: LoopBack application that have been generated using the `lb4 app` command from `@loopback/cli` come with `@loopback/testlab` and `mocha` as a default, so no other testing infrastructure setup is needed.

^ I tried to make it shorter and more to the point. Don't think we need to list both cli commands, just the one we promote as a default for creating a new project in examples. And don't think we need to mention it's a prompt. They should know that because they would've already created the app by now.

pages/en/lb4/Testing-your-application.md
```diff
@@ -65,18 +64,19 @@ Your `package.json` should then look something like this:
   // ...
   "devDependencies": {
     "@loopback/testlab": "^<current-version>",
+    "@types/mocha": "^<current-version>",
     "mocha": "^<current-version>"
   },
   "scripts": {
-    "test": "mocha"
+    "test": "mocha --recursive \"DIST/test\""
```
virkt25: Use of DIST requires the use of `lb-mocha` from `@loopback/build` which replaces `DIST` with the appropriate folder name based on version ... If you want to use mocha then it should be `dist` since we no longer support Node 6 (dist6).

virkt25 (approved):

jannyHou (commented):
Still reviewing more code, just some personal opinion for your question regarding database:

The memory db created in [`example-getting-started`](https://github.com/strongloop/loopback-next/blob/master/packages/example-getting-started/src/datasources/db.datasource.ts) is easy to setup and stores your data in the specified file, which IMO is a reasonable choice for testing data persistence.

b-admike (approved):

bschrammIBM (changes requested):
Line 57: need to make application plural! 
LoopBack applications that have been generated using the `lb4 app` command from

132-135: suggest the following rewrite:
Easier to understand, since it's immediately clear what model properties are
+relevant to the tests. If the tests set the required properties,
+it is difficult to tell whether the properties are actually
+relevant to the tested scenario.

lines 138-141, suggest the following rewrite:
If the tests build model instance data
+manually, you would have to update all tests to set a new required property.
+With a shared helper, you update a single location  with the new property.

pages/en/lb4/Testing-your-application.md
```diff
@@ -270,12 +274,12 @@ describe('ProductController (unit)', () => {
     it('retrieves details of a product', async () => {
       const controller = new ProductController(repository);
       const findStub = repository.find as sinon.SinonStub;
-      findStub.resolves([{id: 1, name: 'Pen'}]);
+      findStub.resolves([{name: 'Pen', slug: 'pen'}]);
 
-      const details = await controller.getDetails(1);
+      const details = await controller.getDetails('pen');
```
jannyHou, line 414: @shimks A general question: do we have an example package contains all these code? 
I came across the thought when searched for function `getDetails`, it should be defined in some folded snippet. Just feel if we have an example testing package might be more straight-forward.(out of the scope of your PR definitely)

bschrammIBM (changes requested):
lines 254-256: control not controls
+Unit tests are considered "white-box" tests because they use an "inside-out"
+approach where the tests know about the internals and control all the variables
+of the system being tested.

shimks:
@jannyHou then the question becomes should the memory db exist in `src` folder since it'll be solely for the test? Or do we want to have it also be the production database?

pages/en/lb4/Testing-your-application.md
shimks: I've actually went ahead and created an app to make sure these codes work. These code snippets are literally code snippets from the project that I've made :)

bschrammIBM:
I am copyediting this section. I will need to finish it tomorrow as we have a translation drop on Weds for Velox. I have submitted a few copyedits. I did want to make a comment that the section is very long and reads like a research paper. I am thinking it should be trimmed to provide only the suggested steps for generating LB4 tests, rather than providing so much discussion? Or maybe the intent was to provide a discussion of the pro's and con's of different approaches. It does seem long.

shimks:
I agree that it is a bit long, but I like the intent that it's going for where the testing strategy is discussed in details. Do you think that it'd be better if it was broken down into sections maybe?

pages/en/lb4/Testing-your-application.md
```diff
@@ -514,19 +544,23 @@ You should have at least one acceptance (end-to-end) test for each of your REST
...
+import {HelloWorldApplication} from '../..';
+import {supertest, expect, createClientForHandler} from '@loopback/testlab';
 import {givenEmptyDatabase, givenProduct} from '../helpers/database.helpers';
+import {RestServer, RestBindings} from '@loopback/rest';
 
 describe('Product (acceptance)', () => {
-  let app: HelloWorldApp;
-  let request: supertest.SuperTest<supertest.Test>;
+  let app: HelloWorldApplication;
+  let client: supertest.SuperTest<supertest.Test>;
```
jannyHou: I think you can simplify it by `let client: Client` and import `Client` from `@loopback/testlab`

See https://github.com/strongloop/loopback-next/blob/master/packages/testlab/src/client.ts#L15

jannyHou (approved):
👍 great update, LGTM! just a few nitpick.

PR #642, opened 2018-03-07
docs: add extension generator on cli page

dhmlau:
Currently in the CLI docs page http://loopback.io/doc/en/lb4/Command-line-interface.html, there is missing extension generator entry in the table.

### DIFF
--- a/_includes/content/lb4-project-commands.html
+++ b/_includes/content/lb4-project-commands.html
@@ -12,6 +12,12 @@
 </td></tr>
 
 <tr>
+  <td><code>lb4 extension</code>
+  </td><td> Create a new LoopBack4 extension
+  </td><td> <a href="Extension-generator.html">Extension generator</a>
+  </td></tr>
+
+<tr>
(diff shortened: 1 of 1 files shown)


_includes/content/lb4-project-commands.html
```diff
@@ -11,6 +11,12 @@
 </td><td> <a href="Application-generator.html">Application generator</a>
 </td></tr>
 
+<tr>
+  <td><code>lb4 example</code>
```
shimks: I don't think is the correct command 😆

kjdelisle (changes requested):
>@shimks
>I don't think is the correct command 😆

_includes/content/lb4-project-commands.html
dhmlau: oops.. you're right!

kjdelisle (approved):
:+1:

shimks (approved):
cool and good

PR #643, opened 2018-03-09
[WIP] Pull LoopBack 4 doc updates from loopback-next

b-admike:
Connect to https://github.com/strongloop/loopback-next/issues/843. This script copies over the doc files for LoopBack 4 content into this repo, and reverts some linking changes made in `loopback-next` in order to render it properly when Jekyll is run. Mad credits to @rmg for essentially writing the script! The script is intended to demonstrate how to achieve the proposal in the spike, so it would probably need to be modified if we are to use this approach.

**EDIT** The script now utilizes `@loopback/docs` to pull in the latest changes from `loopback-next` for LoopBack 4 docs and `jekyll-relative-links` plugin to handle `.md` links (with one exception). It also provides a Travis config to deploy to GitHub Pages whenever there are new changes.


rmg:
Your `update-lb4-docs.sh` needs a shebang at the top (`#!/bin/bash`) to ensure that it is executed with the correct interpreter instead of relying on `/bin/sh` being the default.

update-lb4-docs.sh
```diff
@@ -0,0 +1,19 @@
...
+tmp=$target/tempdir
+
+mkdir -p $tmp
+# pull down all the markdown files (reserving dir structure)
+# given the GitHub org, repo, and branch
+curl -L https://github.com/$org/$repo/archive/$branch.tar.gz \
+   | tar --strip-components 1 -C $tmp -xzf- '*.md'
+cp -R $tmp/packages $target/readmes/$repo
+#rename relative link extensions from md to html
+sed -E -i 's/\(([^(]+)\.md/(\1.html)/g' $tmp/docs/*.md
```
rmg: You have:
```
s/\(([^(]+)\.md/(\1.html)/g
```
It should be this:
```
s/\(([^(]+)\.md\)/(\1.html)/g
```
You need to match the closing `)` literal otherwise it doesn't get replaced by the `)` literal in the replacement and you end up with double `))` on the output - and also incorrectly matching:
 `(index.md other stuff)` -> `(index.html) other stuff)`

b-admike: Good catch. Actually I wanted to include the cases where `.md` is followed by `#` and left out the literal match for `)`. I will remove it from the replacement as well.

bschrammIBM (approved):
lgtm

update-lb4-docs.sh
```diff
@@ -0,0 +1,17 @@
...
+COMMIT=893d68547766fc41d8bca82425645416e357d052
+ARCHIVE_NAME="loopback-next-$COMMIT"
+
+mkdir -p $DOCS
+# pull down all the markdown files (reserving dir structure)
+# given the GitHub ORG, REPO, and COMMIT
+curl -L https://github.com/$ORG/$REPO/archive/$COMMIT.tar.gz | tar --strip-components 1 -C "$DOCS" --wildcards \
+-xzf- "$ARCHIVE_NAME/docs/*.md" "$ARCHIVE_NAME/packages/*.md" "$ARCHIVE_NAME/docs/*.png"
+
+# cp -R $DOCS/packages $DOCS/readmes/$REPO
```
shimks: Was this meant to be commented out?

shimks (approved):
LGTM, but one comment

update-lb4-docs.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+ORG=strongloop
+REPO=loopback-next
+DOCS=./pages/en/lb4
+COMMIT=893d68547766fc41d8bca82425645416e357d052
```
virkt25: Is this meant to be hard-coded?

update-lb4-docs.sh
virkt25: +1

Jenkinsfile
```diff
@@ -9,13 +9,18 @@ def gitEnv = [
...
+    pipelineTriggers([cron('H */1 * * *')])
+])
+
 node('linux && git') {
   stage('checkout') {
     checkout scm
   }
   stage('download updates') {
     sh './update-readmes.sh'
-    sh './update-v4-readmes.sh'
```
raymondfeng: We probably should keep `sh './update-v4-readmes.sh'` here as it's used to pull in README files. Even after we move docs to the monorepo, we might continue to pull in README files.

update-lb4-docs.sh
```diff
@@ -0,0 +1,16 @@
...
+ORG=strongloop
+REPO=loopback-next
+DOCS=./pages/en/lb4
+COMMIT=HEAD
+ARCHIVE_NAME="loopback-next"
+
+mkdir -p $DOCS
+# pull down all the markdown files (reserving dir structure)
+# given the GitHub ORG, REPO, and COMMIT
+curl -L https://github.com/$ORG/$REPO/archive/$COMMIT.tar.gz | tar --strip-components 1 -C "$DOCS" --wildcards \
```
raymondfeng: At some point, it's probably better to switch to `npm install`

raymondfeng (changes requested):

update-lb4-docs.sh
b-admike: Good catch; it's supposed to be master.

shimks (approved):

update-lb4-docs.sh
```diff
@@ -0,0 +1,20 @@
...
+ARCHIVE_NAME="loopback-next-$BRANCH"
+
+mkdir -p $DOCS
+# pull down all the markdown files (reserving dir structure)
+# given the GitHub ORG, REPO, and BRANCH
+#curl -L https://github.com/$ORG/$REPO/archive/$BRANCH.tar.gz | tar --strip-components 1 -C "$DOCS" --wildcards \
+# -xzvf- "$ARCHIVE_NAME/docs/*.md" "$ARCHIVE_NAME/packages/*.md"
+# "$ARCHIVE_NAME/docs/*.png"
+npm i @loopback/docs --prefix $DOCS
+mv $DOCS/node_modules/@loopback/docs $DOCS/monorepo-docs
```
virkt25: Is `/monorepo-docs` needed? Or can we just move the docs fo `$DOCS`?

raymondfeng:
I just published `@loopback/docs@0.1.2`, which has all related files under `site` directory.

update-lb4-docs.sh
```diff
@@ -0,0 +1,20 @@
...
+BRANCH=master
+ARCHIVE_NAME="loopback-next-$BRANCH"
+
+mkdir -p $DOCS
+# pull down all the markdown files (reserving dir structure)
+# given the GitHub ORG, REPO, and BRANCH
+#curl -L https://github.com/$ORG/$REPO/archive/$BRANCH.tar.gz | tar --strip-components 1 -C "$DOCS" --wildcards \
+# -xzvf- "$ARCHIVE_NAME/docs/*.md" "$ARCHIVE_NAME/packages/*.md"
+# "$ARCHIVE_NAME/docs/*.png"
+npm i @loopback/docs --prefix $DOCS
```
raymondfeng: I think it can be further simplified if we add `@loopback/docs` as a dependency in `package.json` of `loopback.io`. This way, `npm install` at the root level will pull in `node_modules/@loopback/docs` and we can add a post-install script to copy them to the right place.
_

update-lb4-docs.sh
raymondfeng: The `site/*` should be copied into `pages/en/lb4/`.

update-lb4-docs.sh
b-admike: I talked to @rmg and if we aim to do that, then we should probably introduce Travis to this repo since the Jenkins pipeline is
>running directly on the Jenkins master and that server doesn't have any external executors 

and doesn't have `node/npm` installed. It would be a hassle to install them there. Also,
>all GitHub Pages does is run Jekyll and upload the generated `_site` directory (or whatever it is..) to S3

update-lb4-docs.sh
b-admike: I was just using it as a placeholder, before we implement changes that override the current doc contents, but yeah it should be `site/*`.

update-lb4-docs.js
```diff
@@ -0,0 +1,34 @@
...
+const path = require('path');
+
+const srcDocs = path.resolve(__dirname, 'node_modules', '@loopback', 'docs', 'site');
+const srcImages = path.resolve(srcDocs, 'imgs');
+
+const destDocs = path.resolve(__dirname, 'pages', 'en', 'lb4');
+const destImages = path.resolve(__dirname, 'images', 'lb4');
+
+function copyDocs(src, dest) {
+    try {
```
raymondfeng: IIRC, we should use two spaces for indentation.

update-lb4-docs.js
```diff
@@ -0,0 +1,34 @@
...
+        fs.readdirSync(src).forEach((val, index) => {
```
raymondfeng: I suggest that we use `fs-extra` to make the code much simpler.

package.json
```diff
@@ -16,6 +18,7 @@
     "url": "git+https://github.com/Strongloop/loopback.io.git"
   },
   "devDependencies": {
+    "@loopback/docs": "^0.1.2",
```
virkt25: Should we use `latest` or `*`?

shimks:
The `Edit this page` button in `_layouts/home.html` and `_layouts/page.html` need to be changed so that when the button is pressed, it links towards the pages in the monorepo and not the one in loopback.io

shimks:
I also wanted to know whether we want to bring over sidebar and some other miscellaneous stuff (one I could think off the top of my head are tables in app/artifact-generator.md). It makes sense to bring this over to the monorepo since these types of files need to be modified/changed as docs are being written/modified.

update-lb4-docs.js
```diff
@@ -0,0 +1,42 @@
+const fs = require('fs-extra');
+const path = require('path');
+
+const srcDocs = path.resolve(
+  __dirname,
+  'node_modules',
+  '@loopback',
+  'docs',
+  'site',
+);
```
raymondfeng: It can be simplified as  `const srcDocs = path.resolve(__dirname, 'node_modules/@loopback/docs/site');`.

update-lb4-docs.js
```diff
@@ -0,0 +1,42 @@
...
+const srcImages = path.resolve(srcDocs, 'imgs');
+const destDocs = path.resolve(__dirname, 'pages', 'en', 'lb4');
+const destImages = path.resolve(__dirname, 'images', 'lb4');
+
+async function copyDocs(src, dest) {
+  try {
+    await fs.copy(src, dest);
```
raymondfeng: I think `fs.copySync` is good enough.

update-lb4-docs.js
```diff
@@ -0,0 +1,42 @@
...
+    console.error('failed to copy docs %s', err.stack);
+  }
+}
+
+copyDocs(srcDocs, destDocs);
+copyDocs(srcImages, destImages);
+fs
+  .ensureDir(path.resolve(destDocs, 'imgs'))
+  .then(() => {
+    fs.removeSync(path.resolve(destDocs, 'imgs'));
```
raymondfeng: No need to check existence of `imgs` when we use `fs.removeSync` (rm -rf). See https://github.com/jprichardson/node-fs-extra/blob/master/docs/remove-sync.md.

update-lb4-docs.js
```diff
@@ -0,0 +1,42 @@
...
+
+const fileToUpdate = path.resolve(destDocs, 'Testing-the-API.md');
+
+try {
+  let contents = fs.readFileSync(fileToUpdate, 'utf-8');
+  contents = contents.replace('include previous.md', 'include previous.html');
+  fs.writeFileSync(fileToUpdate, contents, 'utf-8');
+} catch (err) {
+  console.error('failed to replace relative link %s', err.stack);
+}
```
raymondfeng: If there is an error, we should make sure the process exit with a non-zero code so that `npm` fails.

update-lb4-docs.js
```diff
@@ -0,0 +1,37 @@
+const fs = require('fs-extra');
+const path = require('path');
+
+const srcDocs = path.resolve(__dirname,'node_modules/@loopback/docs/site');
+const srcImages = path.resolve(srcDocs, 'imgs');
+const destDocs = path.resolve(__dirname, 'pages', 'en', 'lb4');
```
raymondfeng: Same here, `pages/en/lb4`

update-lb4-docs.js
```diff
@@ -0,0 +1,37 @@
...
+const destImages = path.resolve(__dirname, 'images', 'lb4');
```
raymondfeng: `images/lb4`

update-lb4-docs.js
```diff
@@ -0,0 +1,37 @@
...
+
+function copyDocs(src, dest) {
+  try {
+    await fs.copySync(src, dest);
```
raymondfeng: `await` should be removed. It cannot be used within a regular function.

b-admike:
@slnode test please

Jenkinsfile
```diff
@@ -9,13 +9,19 @@ def gitEnv = [
...
+])
+
 node('linux && git') {
   stage('checkout') {
     checkout scm
   }
   stage('download updates') {
     sh './update-readmes.sh'
     sh './update-v4-readmes.sh'
+    sh './update-lb4-docs.sh'
```
raymondfeng: I only see `update-lb4-docs.js`

raymondfeng (approved):
Please squash the commits before merge.

raymondfeng:
@b-admike Do we want to delete existing files from `pages/en/lb4` and add this folder to `.gitignore`?

b-admike:
@raymondfeng I was thinking we overwrite them, but that is a good idea.

.travis.yml
```diff
@@ -0,0 +1,39 @@
...
+
+after_success:
+  - npm start
+
+jobs:
+  include:
+    - stage: copy v2,v3 readmes
+      script: /bin/bash ./update-readmes.sh
+      after_success: skip
+      os:
```
virkt25: In a matrix config, if the setting is the same as the one outside it doesn't need to be re-listed. So in other words, you can remove the `os` property from under each `stage`.

.travis.yml
```diff
@@ -0,0 +1,39 @@
...
+        - linux
+
+deploy:
+  provider: pages
+  skip-cleanup: true
+  github-token: $GITHUB_TOKEN
+  keep-history: true
+  on:
+    branch: master
```
virkt25: Needs newline

virkt25 (changes requested):
I don't think the Travis config will do what you're trying to accomplish. Each Job runs in it's own VM and as such the copied artifacts aren't shared across each job or stage (deploy is considered a stage). The scripts for moving the readme's should be run in the `deploy` stage. 

Reference: 
https://docs.travis-ci.com/user/build-stages/#Data-persistence-between-stages-and-jobs

update-lb4-docs.js
```diff
@@ -0,0 +1,37 @@
+const fs = require('fs-extra');
+const path = require('path');
+
+const srcDocs = path.resolve(__dirname,'node_modules/@loopback/docs/site');
+const srcImages = path.resolve(srcDocs, 'imgs');
+const destDocs = path.resolve(__dirname, 'pages/en/lb4');
+const destImages = path.resolve(__dirname, 'images/lb4');
```
raymondfeng: Now we copy markdown and image files to two different places. Is it going to be  a problem for image links such as `./imgs/my-diagram.png`?

update-lb4-docs.js
```diff
@@ -0,0 +1,37 @@
...
+
+function copyDocs(src, dest) {
+  try {
+    fs.copySync(src, dest);
```
raymondfeng: I think we should call `fs.removeSync(dest);` first now.

update-lb4-docs.js
b-admike: That's a good point. I've removed `images/lb4` entirely, and the images will live in `pages/en/lb4/imgs`. That way we don't have to worry about image links too (I've verified it).

update-lb4-docs.js
```diff
@@ -0,0 +1,40 @@
...
+  } catch (err) {
+    console.error('failed to copy latest docs %s', err.stack);
+    process.exit(1);
+  }
+}
+
+// copy the latest docs from @loopback/docs to pages/en/lb4 directory
+copyDocs(srcDocs, destDocs);
+
+const fileToUpdate = path.resolve(destDocs, 'Testing-the-API.md');
```
shimks: Obsolete code?

b-admike: Nah, it is used down the line (see https://github.com/strongloop/loopback.io/pull/643/commits/1ba3aba52978255543499c0ceaa8f429e2f30427#diff-015a67e879f2083136bcd7796abb4f33R34)

Jenkinsfile
```diff
@@ -9,6 +9,11 @@ def gitEnv = [
 
 env.CHANGE_BRANCH = env.CHANGE_BRANCH ?: env.BRANCH_NAME
 
+properties([
+    // Run every hour
```
bajtos: Running Jenkins job every hour, isn't that excessive use of resources? Ideally, we should use GitHub's HTTP hooks to trigger the run of this job right after loopback-next's master branch was changed, or perhaps use some sort of npm mechanism to run the job after a new version of `@loopback/docs` was published.

If there isn't any easy solution for triggering the run only when there are changes to pull, then I am proposing to change this frequence to something like once a day, or at least every 4 hours or so.

cc @rmg

.travis.yml
```diff
@@ -0,0 +1,32 @@
...
+deploy:
+  provider: pages
+  skip-cleanup: true
+  github-token: $GITHUB_TOKEN
+  keep-history: true
+  on:
+    branch: pull-monorepo-docs #for testing purposes
+  include:
+    - stage: copy v2,v3 readmes
+      script: /bin/bash ./update-readmes.sh
```
bajtos: I find it a bit confusing that we have both Travis CI and Jenkins configured for this repository. What is the interaction between these two CI servers? Could you please describe this setup in more details, e.g. in [README.md](https://github.com/strongloop/loopback.io/blob/gh-pages/README.md) or perhaps in a new file?

cc @rmg

Jenkinsfile
b-admike: Yeah I had a discussion with @rmg and running every hour isn't excessive from what he told me (correct me if I'm wrong, @rmg). But I would like us to use Travis instead of Jenkins for updating the docs, so this is probably outdated (I'll probably revert/disable this once we have Travis up and running).

.travis.yml
b-admike: Good point. I will take a crack at documenting it.

_layouts/page.html
```diff
@@ -8,10 +8,14 @@
...
    <h1 class="post-title-main">{{ page.title }}</h1>
-   {% if site.github_editme_path and page.layout != 'readme' %}
+   {% if site.github_editme_path and page.layout != 'readme' and page.sidebar != 'lb4_sidebar' %}
        {% assign edit_url =  page.url | remove: "/doc" | remove: ".html" | append: ".md" | prepend: site.github_editme_path %}
        <a target="_blank" href="https://github.com/{{edit_url}}" class="btn btn-default githubEditButton" role="button"><i class="fa fa-github fa-lg"></i> Edit this page</a>
    {% endif %}
+   {% if site.lb4_editme_path and page.layout != 'readme' and page.sidebar == 'lb4_sidebar' %}
+       {% assign edit_url =  page.url | remove: "doc/en/lb4" | remove: ".html" | append: ".md" | prepend: site.lb4_editme_path %}
+       <a target="_blank" href="https://github.com/{{edit_url}}" class="btn btn-default githubEditButton" role="button"><i class="fa fa-github fa-lg"></i> Edit this page</a>
+  {% endif %}
```
rmg: wrong indentation?

Jenkinsfile
```diff
@@ -9,6 +9,7 @@ def gitEnv = [
 
 env.CHANGE_BRANCH = env.CHANGE_BRANCH ?: env.BRANCH_NAME
 
+
```
rmg: Looks like a random whitespace change was accidentally left in.

.travis.yml
```diff
@@ -0,0 +1,21 @@
...
+  keep-history: true
+  on:
+    branch: pull-monorepo-docs #for testing purposes
+  script:
+    - /bin/bash ./update-readmes.sh
+    - /bin/bash ./update-v4-readmes.sh
+    - npm test
+  email: slnode@ca.ibm.com
+  name: StrongLoop Bot
+
```
virkt25: Think there's some extra empty space here.

rmg:
Travis has been configured

virkt25 (approved):

raymondfeng (approved):

b-admike:
Unfortunately, we can't use custom commit messages for the commits made by Travis on successful builds, but the feature might come soon https://github.com/travis-ci/travis-ci/issues/9287.

PR #644, opened 2018-03-12
docs(boot): Move boot docs to top level

dhmlau:
The `Booting an application` page does not seem to fit under `Key Concept` in the official documentation. 
I'm suggesting to move it to the top level. 

connected to https://github.com/strongloop/loopback-next/issues/988

### DIFF
--- a/_data/sidebars/lb4_sidebar.yml
+++ b/_data/sidebars/lb4_sidebar.yml
@@ -41,10 +41,6 @@ children:
     url: Dependency-injection.html
     output: 'web, pdf'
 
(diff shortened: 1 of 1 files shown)


kjdelisle (approved):
I don't really see a problem with this.

Keep in mind that this will all probably change once we start pulling the docs from this repo directly.

virkt25 (approved):

dhmlau:
@kjdelisle , would we change what's shown on the sidebar? if it's all gonna change, then there's no need for this PR :)

b-admike (approved):

PR #647, opened 2018-03-20
no need to update "appliction.ts" in lb4.

mlibre:
There is no setupControllers() in lb4 anymore/anyway.
http://loopback.io/doc/en/lb4/Getting-started.html

### DIFF
--- a/pages/en/lb4/Getting-started.md
+++ b/pages/en/lb4/Getting-started.md
@@ -70,20 +70,6 @@ Let's add a simple "Hello World" controller as follows:
   }
   ```
 
-* Update `/src/application.ts` to load the controller:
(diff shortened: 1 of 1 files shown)


shimks (changes requested):
We've recently changed how docs are going to be written with LoopBack4 and onwards. From now on, please make any docs change to the monorepo: https://github.com/strongloop/loopback-next.

The file that needs to be changed is here: https://github.com/strongloop/loopback-next/blob/master/docs/site/Getting-started.md

See https://github.com/strongloop/loopback.io/pull/643 for more info.

The change itself looks good though :)

mlibre:
ok. ty

PR #658, opened 2018-04-02
chore: tweaks to scripts

virkt25:
loopback.io deploys are failing since old files aren't removed from `@loopback/docs`. This PR:

- removes old folder before copying files from `@loopback/docs`
- Tweaks travis.yml and package.json scripts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ node_js:
 os:
   - linux
 
+script:
+  skip
+
 branches:
   only:
   - gh-pages
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,6 @@ deploy:
   on:
     branch: gh-pages
   script:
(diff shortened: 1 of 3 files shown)


b-admike (approved):

shimks (approved):

.travis.yml
```diff
@@ -5,6 +5,9 @@ node_js:
 os:
   - linux
 
+script:
+  skip
```
b-admike, line 9: IIUC, this is to skip the scripts (`npm run fetch-readmes` and `npm test` I presume) until you get a green light right?

virkt25: It will skip all script when a PR is made but will run them on deploy (after a PR is merged / every 24 hours). I just didn't like seeing the `No test specified` message.

virkt25: At least that's what I'm aiming for. I'll revert this if it's not this behaviour.

PR #670, opened 2018-04-18
lb3: improve Model Discovery docs

### LINKED ISSUE #347: Need better examples of how to discover models from Relational DBs
## URL of the page 

http://loopback.io/doc/en/lb3/Discovering-models-from-relational-databases.html

## Nature of the issue 
I'm not sure what the "other repos" being referred to are, but if this is the wrong spot I'd be happy to move this. 

I've been interested in how this works and despite the really simple database I was using that had a single table and two columns, I was a little confused about how this was supposed to actually work.

The docs say, "Use fs.writeFile() to save the output in common/models/model-name.json." Output of what? I'm assuming it was referring to the `models` parameter in the callback function for `discoverAndBuildModels`, but upon printing it to my console, it looks nothing like a model definition. After some digging it seems that `discoverSchema(s)` is closer to what I was expecting that output to be.

Another issue is that the example doesn't actually show how to write the files out, using information from the discovered models. I had to find this [Stackoverflow answer](http://stackoverflow.com/a/31312012) to get me pointed in the right direction.

I see this feature being pointed to from a lot of different places, so it doesn't seem to be some obscure feature, so either that's not true or I'm missing something.

bajtos:
Enhance the documentation page [Discovering models from relational databases](http://loopback.io/doc/en/lb3/Discovering-models-from-relational-databases.html) to show both options for discovering models:
 - automatically load discovered models at runtime
 - discover only model definitions and create model JSON files accordingly

These changes should hopefully fix the confusion caused by the current content.

Fixes #347

### DIFF
--- a/pages/en/lb3/Discovering-models-from-relational-databases.md
+++ b/pages/en/lb3/Discovering-models-from-relational-databases.md
@@ -23,69 +23,199 @@ This process is called _discovery_ and is supported by the following connectors:
 
 For NoSQL databases such as MongoDB, use [instance introspection](Creating-models-from-unstructured-data.html) instead.
 
+{% include warning.html content="
+The discovery of model relations is limited to
+[BelongsTo](./BelongsTo-relations.md) relations.
+Other relation types, for example [HasMany](./HasMany-relations.md),
+are not supported yet.
+" %}
+
 Data sources connected to relational databases automatically get the asynchronous 
 [Database discovery API](http://apidocs.loopback.io/loopback-datasource-juggler/#datasource-prototype-discoverandbuildmodels).
 
-### Basic procedure
+LoopBack supports two flavours of model discovery:
 
-Follow these basic steps:
+ 1. Discover and define models at runtime during application startup.
+ 2. Discover model definitions on demand and save these definitions
+    to model JSON files.
 
-1.  Run a script to discover the model definitions for use in your application.
-See the [example below](#additional-discovery-functions) for details.
-2.  Use `fs.writeFile()` to save the output in `common/models/model-name.json`.
-3.  Add the new model entry to `server/model-config.json`.
-4.  Run the app:
+In the first flavor, models are discovered and defined in the application
+in a single step at runtime. There is no source code nor any JSON (model)
+config files versioned for these discovered models, they are all loaded by
+the discovery script. Once the discovery script has finished, the discovered
+models can be accessed via `app.models` API). The application always uses
+the data structures (models) appropriate for the environment it runs in
+(dev/test/production database). This means the same application version
+can use different data structures (models) when it's deployed to different
+environments.
 
-    ```shell
-    $ node .
-    ```
+In the second flavor, models are discovered at the time the application is
+developed, their definitions are stored in model JSON files and versioned
+by git. This allows developers to make further tweaks and changes to model
+schemas, add custom model functions, etc. The application uses the same data
+structures (models) regardless of the environment it runs in
+(dev/test/production database).
 
-5.  Use LoopBack Explorer to verify the schema is defined properly.
+In both cases, you need a data source configured to access the database
+you would like to discover, see [Defining data
+sources](http://loopback.io/doc/en/lb3/Defining-data-sources.html). The code
+examples in the following sections assume that your application
+has a data source called `db` that's configured to access the target database.
 
-## Example discovery
+See our [database tutorial](https://github.com/strongloop/loopback-example-database)
+for an example application that leverages discovery and can be run out of the
+box.
 
-For example, consider an Oracle database. First, the code sets up the Oracle data source.
-Then the call to `discoverAndBuildModels()` creates models from the database tables.
-Calling it with the `associations: true` option makes the discovery follow primary/foreign key relations.
+## Discover and define models at runtime
 
-{% include code-caption.html content="/server/bin/script.js" %}
-```javascript
-var loopback = require('loopback');
-var ds = loopback.createDataSource('oracle', {
-  "host": "oracle-demo.strongloop.com",
-  "port": 1521,
-  "database": "XE",
-  "username": "demo",
-  "password": "L00pBack"
-});
+Use [dataSource.discoverAndBuildModels()](http://apidocs.loopback.io/loopback-datasource-juggler/#datasource-prototype-discoverandbuildmodels)
+API to discover and define your model classes at runtime.
 
-// Discover and build models from INVENTORY table
-ds.discoverAndBuildModels('INVENTORY', {visited: {}, associations: true},
-function (err, models) {
-  // Now we have a list of models keyed by the model name
-  // Find the first record from the inventory
-  models.Inventory.findOne({}, function (err, inv) {
-    if(err) {
-      console.error(err);
-      return;
-    }
-    console.log("\nInventory: ", inv);
-    // Navigate to the product model
-    // Assumes inventory table has a foreign key relationship to product table
-    inv.product(function (err, prod) {
-      console.log("\nProduct: ", prod);
-      console.log("\n ------------- ");
+For example, you can add a following boot script to the `server/boot`
+directory in your LoopBack project.
+
+{% include code-caption.html content="/server/boot/discover-models.js" %}
+```js
+'use strict';
+
+module.exports = function(app, callback) {
+  // Obtain the datasource registered with the name "db"
+  const dataSource = app.dataSources.db;
+
+  // Step 1: define a model for "INVENTORY" table,
+  // including any models for related tables (e.g. "PRODUCT").
+  dataSource.discoverAndBuildModels(
+    'INVENTORY',
+    {relations: true},
+    function(err, models) {
+      if (err) return callback(err);
+
+      // Step 2: expose all new models via REST API
+      for (const modelName in models) {
+        app.model(models[modelName], {dataSource: dataSource});
+      }
+
+      callback();
     });
-  });
-});
+};
 ```
+
+Next steps:
+
+ 1. Start your application in the usual way, e.g. via `npm start` or `node .`
+ 2. Use [API Explorer](./Use-API-Explorer.md) to inspect your dynamically
+    created REST API.
+
+## Discover and save model definitions
+
+Use [dataSource.discoverSchema()](http://apidocs.loopback.io/loopback-datasource-juggler/#datasource-prototype-discoverschema)
+API to obtain a model definition for a given database table.
+
+To leverage this API, you need to write a small script to discover the models
+you are interested in and write them to model JSON files.
+
+Let's walk through an example implementation of such script.
+
+{% include note.html content="
+The code examples in this section are using features introduced in
+Node.js 8.x LTS. Users running on an older version of Node.js need
+to rework the code to use promises (or callbacks) instead of async/await
+for flow control.
+" %}
+
+First of all, we need to import APIs we will be using later:
+
+{% include code-caption.html content="/discover-models.js" %}
+```js
+'use strict';
+
+const loopback = require('loopback');
+const promisify = require('util').promisify;
+const fs = require('fs');
+const writeFile = promisify(fs.writeFile);
+const readFile = promisify(fs.readFile);
+const mkdirp = promisify(require('mkdirp'));
+```
+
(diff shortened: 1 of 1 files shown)


pages/en/lb3/Discovering-models-from-relational-databases.md
```diff
@@ -23,69 +23,182 @@ This process is called _discovery_ and is supported by the following connectors:
...
 
+Node.js does not allow the main module function to be async,
+the top-level code cannot use `await` keyword for flow control.
+As a workaround, we need to wrap the discovery code in a helper function
+that's executed using Promise API.
+
+```js
+discover().then(
+  success => process.exit(),
+  error => { console.error('UNHANDLED ERROR:\n', error); process.exit(1); },
```
shimks, line 160: I'm not an expert on promises, but I thought errors were usually handled by `.catch()` chain. Is there a reason we prefer this style over the other?

pages/en/lb3/Discovering-models-from-relational-databases.md
```diff
@@ -23,69 +23,182 @@ This process is called _discovery_ and is supported by the following connectors:
...
+  await writeFile(
+    'server/model-config.json',
+    JSON.stringify(config, null, 2)
+  );
+}
+```
+
+Next steps:
+
+ 1. Run the discovery script to add the discovered models to your project
```
shimks: I'm really not sure what the context to this step here is. We've managed to create a config file from the schemas saved in the database, but what purpose does that serve? What have we done so far that allows us to `discover and save model definitions`? Is the `discovery script` the code snippet from the previous section?

shimks (commented):
This just might be due to the gap in my knowledge in lb3, but I don't really know what it means to `discover and define models at runtime`. Since there is a separate section for `saving` the model defs, I don't really know what's happening with the datasource in that section.

From guessing, I'm inclined to believe that the first section is on 'discovering' models from databases and then turning them into models that loopback can use, and the second section is on taking models described in the form of JSON files and then putting them into the databases. Are these inferrences correct?

raymondfeng (approved):

pages/en/lb3/Discovering-models-from-relational-databases.md
al-k21: The discussion about `unhandledRejection` has come up in the nodejs community many times. The last outcome I know about is:

- Catch rejected promises and throw straight away; then remove once this is standard in Node, which it would be on Node 10:

```
// Catch rejected promises and throw straight away (remove once this is standard in node).
process.on('unhandledRejection', (reason, p) => {
  console.error(chalk.red('Rejected promise:'), p);
  throw reason;
});
```

pages/en/lb3/Discovering-models-from-relational-databases.md
```diff
@@ -23,69 +23,182 @@ This process is called _discovery_ and is supported by the following connectors:
...
+  // of dataSource.discoverSchemas(), it allows the method to cache
+  // discovered related models
+  const options = {relations: true};
+
+  // Discover models and relations
+  const salariesSchemas = await db.discoverSchemas('salaries', options);
+  const employeesSchemas = await db.discoverSchemas('employees', options);
+
+  // Create model definition files
+  await mkdirp('common/models');
```
al-k21: I'd personally find using `mkdirSync` more readable. https://nodejs.org/api/fs.html#fs_fs_mkdirsync_path_mode

pages/en/lb3/Discovering-models-from-relational-databases.md
```diff
@@ -23,69 +23,182 @@ This process is called _discovery_ and is supported by the following connectors:
...
+  await writeFile(
```
al-k21: I'd personally find using `writeFileSync` more readable. https://nodejs.org/api/fs.html#fs_fs_writefilesync_file_data_options

pages/en/lb3/Discovering-models-from-relational-databases.md
```diff
@@ -23,69 +23,182 @@ This process is called _discovery_ and is supported by the following connectors:
...
+    'common/models/salaries.json',
+    JSON.stringify(salariesSchemas['employees.salaries'], null, 2)
+  );
+  await writeFile(
```
al-k21: `writeFile` => `writeFileSync`

pages/en/lb3/Discovering-models-from-relational-databases.md
```diff
@@ -23,69 +23,182 @@ This process is called _discovery_ and is supported by the following connectors:
...
+  );
+
+  // Expose models via REST API
+  const configJson = await readFile('server/model-config.json', 'utf-8');
+  console.log('MODEL CONFIG', configJson);
+  const config = JSON.parse(configJson);
+  config.Salaries = {dataSource: DATASOURCE_NAME, public: true};
+  config.Employees = {dataSource: DATASOURCE_NAME, public: true};
+  await writeFile(
+    'server/model-config.json',
```
al-k21: `writeFile` => `writeFileSync`

pages/en/lb3/Discovering-models-from-relational-databases.md
```diff
@@ -23,69 +23,182 @@ This process is called _discovery_ and is supported by the following connectors:
...
 ## Additional discovery functions
 
 Some connectors provide discovery capability so that we can use DataSource to discover model definitions from existing database schema.
 The following APIs enable UI or code to discover database schema definitions that can be used to build LoopBack models.
 
-`discoverModelDefinitions()` by default discovers database schema definitions owned by the userid used to connect to the database.  
+`discoverModelDefinitions()` by default discovers database schema definitions owned by the userid used to connect to the database.
 Calling it with the `all: true` option makes the discovery include the database schema definitions of all owners.
 
 {% include code-caption.html content="/server/bin/script.js" %}
```
al-k21, line 221: Maybe (for consistency with the other examples):
`PRODUCT` => `employees`
`INVENTORY` => `salaries`

al-k21 (commented):
Thanks, it is much clearer now!
LGTM, just couple nits.

pages/en/lb3/Discovering-models-from-relational-databases.md
bajtos: The difference between a two-args `then` and `catch` is in the case when a promise is resolved (successfully) and the first callback in `then` throws an error (e.g. `TypeError: undefined is not a function`). 

 - With a two-arg `.then()`, this error correctly triggers `unhandledRejection`
 - With `.catch()`, this error is passed to the `catch` callback, which has the effect that both "then" and "catch" callbacks are called.

Here is the situation where such behavior is not intended:

```ts
function legacyApiWithCallback(callback) {
  somePromiseApi()
    .then(() => callback())
    .catch(callback);
}
```

If `callback` throws an error, it will be called again to handle the error it has thrown.

pages/en/lb3/Discovering-models-from-relational-databases.md
bajtos: Good catch!

In the first part of this section, I have described a discovery script - how to write the code.

In this step, I am telling our users to run the script they have written. (Compare this to the previous section, where the boot script is automatically run by `node .`.)

How can I make this intent more clear? Should I add a "step 0" saying something like "save the source code shown above to a file in your project's root directory"?

> We've managed to create a config file from the schemas saved in the database, but what purpose does that serve? 

We created model config files in such way that application is going to automatically pick these models during app boot (via loopback-boot). 

> What have we done so far that allows us to discover and save model definitions? Is the discovery script the code snippet from the previous section?

So far, the documentation has described how to write a script for discovering models and updating app config to add these models and expose them via app's REST API.

The discovery script is a concatenation of all code snippets from this section.

@shimks could you please help me with coming up with a better way how to express this information in the doc page? Feel free to make the edits directly in the file and push them as a new commit to this feature branch.

pages/en/lb3/Discovering-models-from-relational-databases.md
bajtos: I need `mkdirp` because the directory `common` may not exist yet. It has a sync API too ([docs](https://www.npmjs.com/package/mkdirp#mkdirpsyncdir-opts)). I prefer `await` over sync API, I find it more idiomatic.

Or are you suggesting to get rid of `mkdirp` and call `mkdir`/`mkdirSync` twice, once for each directory? I find that cumbersome.

pages/en/lb3/Discovering-models-from-relational-databases.md
bajtos: Ah, good catch. I'll do the opposite change and update my new examples to use `PRODUCT` and `INVENTORY`.

pages/en/lb3/Discovering-models-from-relational-databases.md
al-k21: > I find it more idiomatic.

Sure, it was just a suggestion.

pages/en/lb3/Discovering-models-from-relational-databases.md
al-k21: > Should I add a "step 0" saying something like "save the source code shown above to a file in your project's root directory"?

al-k21: > Should I add a "step 0" saying something like "save the source code shown above to a file in your project's root directory"?

Sounds reasonable to me (that's what I have done).

al-k21 (approved):

bajtos:
Thank you @al-k21 for joining the discussion, I'll answered your comments in-line (see above).

@shimks 
>This just might be due to the gap in my knowledge in lb3, but I don't really know what it means to discover and define models at runtime. Since there is a separate section for saving the model defs, I don't really know what's happening with the datasource in that section.
>
> From guessing, I'm inclined to believe that the first section is on 'discovering' models from databases and then turning them into models that loopback can use, and the second section is on taking models described in the form of JSON files and then putting them into the databases. Are these inferrences correct?

I see how this part can be confusing, let's work together to find a way how to clarify it.

The first section describes the approach where models are discovered and defined in the application in a single step. All of this is happening at runtime and there are no model JSON files involved at all. The source code & JSON config files versioned by `git` do not have anything for these discovered models, they are all loaded by the discovery script. (Of course, once your boot script is done, you can access the discovered models via `app.models.Product` and `app.models.Inventory`).

I personally find this approach worrying, because the application can potentially use different data structures (models) when it's run against development, test and production databases.

The second approach describes my preferred approach, which is BTW implemented in our GUI tools like APIC Editor too. We discover the models at dev time (the developer has to explicitly execute model discovery), store the discovered schemas to model JSON files and commit them to git. The developer can make any further tweaks and changes to model schemas, add custom model functions, etc. and version everything in git. The application uses the same data structures (models) regardless of the environment it runs in (dev/test/prod database).

Now my question to you - how can I capture this information in the doc page?

pages/en/lb3/Discovering-models-from-relational-databases.md
shimks: I think it's sufficient enough to add in some words to the first step:
`Run the discovery script we've just written to add the discovered models ...`. In hindsight though, it feels like what the discover script was is really obvious now, so feel free to just leave it as it is.

shimks:
@bajtos Thanks for the clarification, the contents of the docs make much more sense to me now! I think in order to help out users like myself, it's sufficient enough to add in short descriptions of what the two approaches seek to accomplish and maybe the pros and cons of each approaches of model discovery. It's with this context that I was able to realize what each approaches were for.

I think the appropriate place to give this context is under `LoopBack supports two flavors of model discovery: ...`.

bajtos:
@shimks I have updated the content based on your comments (see 406ebdb). LGTY now?

shimks (approved):
LGTM. Thanks for helping me understand!

arjamizo:
Regarding [1], has anyone tried doing `hasMany` discovery (some fork modifyng [2])?

Thanks!

[1] https://github.com/strongloop/loopback.io/blame/c7f9ca9/pages/en/lb3/Discovering-models-from-relational-databases.md#L30 (or live https://github.com/strongloop/loopback.io/blame/gh-pages/pages/en/lb3/Discovering-models-from-relational-databases.md)
[2] https://github.com/strongloop/loopback-datasource-juggler/blob/bca6315/lib/datasource.js#L1840-L1877

PR #669, opened 2018-04-18
Add recommendations for async + promises to styleguide

joeytwiddle:
I tried to provide some justification for each recommendation.

On the one hand that may be considered clutter, on the other it may be useful to the reader!

Let me know if you need any changes, or feel free to make them directly yourself!

PR #671, opened 2018-04-18
Add recommendation for final await in an async functions

joeytwiddle:
Additional to #669 

I realised just too late that there is another style we can document.

I have seen this pattern in your code and I do think it helps to reduce ambiguity.

(It might have a small performance overhead, consuming a promise only to immediately create another when returning, but I guess that's only a concern for inner loops.)

### DIFF
--- a/pages/en/contrib/style-guide.md
+++ b/pages/en/contrib/style-guide.md
@@ -449,6 +449,46 @@ async function getDodo() {
 }
 ```
 
+### Always await the last promise in a function
+
+If the last line of a function handles a promise, always `await` that promise, regardless whether the result will be returned or not.
+
+(This makes it clear whether we intend to return the value or not, and dismisses any idea that we are still linking up the promise chain in pre-await style.  Consistently using await also ensures that a try-catch around the function body will catch errors even from the last call.)
+
+**Good**:
+
+```js
+async function openCircus() {
+  await openCarPark();
+  await openDoors();
+}
+```
+
+**Good**:
+
+```js
+async function getAnimal() {
+  return await getElephant();
+}
+```
+
+**Bad**:
+
+```js
+async function openCircus() {
+  await openCarPark();
+  return openDoors();
+}
+```
+
+**Bad**:
+
+```js
+async function getAnimal() {
+  return getElephant();
+}
+```
+
 ## Style guidelines for tests
 
 ### Sandbox directories


shimks (approved):

b-admike (approved):

pages/en/contrib/style-guide.md
```diff
@@ -449,6 +449,44 @@ async function getDodo() {
 }
 ```
 
+### Always await the last promise in a function
+
+If the last line of a function handles a promise, always await that promise, regardless whether the result will be returned or not. (This makes it clear whether we did or did not intend to return the value, and dismisses any idea that we are still linking up the promise chain in pre-await style.)
+
```
raymondfeng: Is it better to rephrase it as `This makes it clear whether we intend to return the value or not, and dismisses any idea that we are still linking up the promise chain in pre-await style.`?

pages/en/contrib/style-guide.md
```diff
@@ -449,6 +449,44 @@ async function getDodo() {
...
+  await openCarPark();
+  return openDoors();
+}
+```
+
+**Bad**:
+
+```js
+async function getAnimal() {
+  return getElephant();
```
raymondfeng, line 488: In this case, `async` keyword is not needed as no `await` is used in the method body.

pages/en/contrib/style-guide.md
```diff
@@ -449,6 +449,44 @@ async function getDodo() {
...
+  return await getElephant();
+}
+```
+
+**Bad**:
+
+```js
+async function openCircus() {
+  await openCarPark();
+  return openDoors();
```
raymondfeng, line 480: Should we mention this style is not friendly to `try...catch`. For example:
```
async function openCircus() {
  try {
    await openCarPark();
    return await openDoors();
  } catch(err) {
    // rejected promise from openDoors will be trapped here
    throw err;
  }
```

```
async function openCircus() {
  try {
    await openCarPark();
    return openDoors();
  } catch(err) {
    // rejected promise from openDoors will NOT be trapped here
    throw err;
  }
```

pages/en/contrib/style-guide.md
shimks: I think the point @joeytwiddle was trying to give a bad example of using promises and async/await. Users might be tempted to return a promise for the purpose of promise chaining, so this example is showing them to not do that

raymondfeng: I understand. But I would like to help users understand why it's bad.

joeytwiddle: In this example I want to emphasize that the problem is the lack of `await`.

If we have an example without `async` I am worried the reader may think that the missing `async` is the problem.

It's worth noting that two sections above we have said "Always use the `async` keyword" so not having `async` is already considered undesirable.

pages/en/contrib/style-guide.md
joeytwiddle: Thanks, that is another good reason. I have added a note about that.

Although I didn't include specific examples, concerned they would be a lot for the reader to read.

pages/en/contrib/style-guide.md
joeytwiddle: Thanks for your clearer rewording. Updated.

pages/en/contrib/style-guide.md
raymondfeng: That's fine.

dhmlau:
@raymondfeng , are you good with the latest changes in the PR?  
@joeytwiddle, you'd need to rebase the PR.  Thanks.

joeytwiddle:
Rebased it!

raymondfeng (approved):

PR #674, opened 2018-04-30
Bugfix in doc, polymorphic foreign key

Menziess:
Definition should be contained inside relation. 

This fixes the error: 
`/.../node_modules/loopback-datasource-juggler/lib/datasource.js:2050
    pkType = foreignModel.properties[pkName].type;
                                             ^
TypeError: Cannot read property 'type' of undefined`

### DIFF
--- a/pages/en/lb3/Polymorphic-relations.md
+++ b/pages/en/lb3/Polymorphic-relations.md
@@ -248,10 +248,10 @@ First define relations in through model ImageLink.
       "type": "belongsTo",
       "model": "Picture",
       "foreignKey": ""
-    },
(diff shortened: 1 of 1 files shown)


virkt25 (approved):

shimks (approved):

b-admike (approved):

jannyHou (approved):
@Menziess thanks! The fix is correct, corresponding code example could be found in https://github.com/strongloop/loopback-example-relations/blob/example/polymorphic-relations/common/models/polymorphic/hasManyThrough/image-link.json#L14-L24

PR #676, opened 2018-05-04
chore: update site analytics

virkt25:
GDPR Compliance. Removes Google Analytics and uses IBM Coremetrics via Tealium. Users can opt-out of cookie tracking.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -41,7 +41,5 @@
 <!-- /.container -->
     </div>
 </body>
-{% if site.google_analytics %}
-{% include google_analytics.html %}
-{% endif %}
+
 </html>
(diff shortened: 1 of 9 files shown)


b-admike (approved):

_layouts/default.html
```diff
@@ -41,7 +41,5 @@
 <!-- /.container -->
     </div>
 </body>
-{% if site.google_analytics %}
-{% include google_analytics.html %}
-{% endif %}
+
```
shimks, line 44: Not exactly sure how it works, but why do we not replace `google_analytics` with the new `analytics` file?

virkt25: Because this layout loads the `head.html` ... which in turn loads `analytics.html`

shimks (approved):

dhmlau (approved):

PR #682, opened 2018-05-16
Update Making-authenticated-requests.md

wolrajhti:
`PersistedModel.destroyAll` takes an optional `where filter` instead of a `Filter JSON object`

[link to the docs](https://apidocs.strongloop.com/loopback/#persistedmodel-destroyall)


dhmlau:
@jannyHou , could you PTAL? Thanks.

jannyHou (approved):
👍 @wolrajhti Thank you for correcting the input of `destroyAll()` method.

You are right, here we should provide a where clause not the whole filter. See code https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L2191

PR #695, opened 2018-05-29
Update Making-authenticated-requests.md

wolrajhti:
### Description

Since `middleware.json` exists, it's better to use it instead of the old raw `app.use()`.

### Related

[https://github.com/strongloop/loopback/pull/3891](https://github.com/strongloop/loopback/pull/3891)

### DIFF
--- a/pages/en/lb3/Making-authenticated-requests.md
+++ b/pages/en/lb3/Making-authenticated-requests.md
@@ -39,7 +39,7 @@ curl -X GET http://localhost:3000/api/widgets?access_token=$ACCESS_TOKEN
 To use cookies for authentication, add the following to middleware.json :
 
 {% include code-caption.html content="/server/middleware.json" %}
-```javascript
+```json
 {
     "auth": {
         "loopback#token": {}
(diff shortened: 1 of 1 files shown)


pages/en/lb3/Making-authenticated-requests.md
```diff
@@ -36,13 +36,15 @@ http://localhost:3000/api/widgets
...
 ```
 
-To use cookies for authentication, add the following to server.js (before boot):
-
-{% include code-caption.html content="/server/server.js" %}
-```javascript
-app.use(loopback.token({
-    model: app.models.accessToken
-}));
+To use cookies for authentication, add the following to middleware.json :
```
b-admike: Can we preserve both the programatic and declarative way of doing this?

wolrajhti: you're right I add the 2 options (and replace app.use by app.middleware for phases)

bajtos (approved):
This is awesome, thank you @wolrajhti for the contribution ❤️

PR #717, opened 2018-07-27
Update HasOne-relations.md

juliandanciu:
This documentation chooses to explain the HasOne relationship with the Account and Supplier models as examples. Since we are documenting the HasOne relationship, it is my understanding that the DECLARING model is the Supplier model and the TARGET model is the Account model. The foreign key therefore would be on the declaring model as account.supplierId. I corrected the documentation to reflect that. Also, the type of the foreign key property (defined on the target model) should be the same as the type of the id key property on the declaring model (and not the target model as the docs state). I believe that the use of declaring and target has been flipped mistakenly. Of course, I'm just learning this stuff so I could be very wrong. If so, please disregard this proposed file change :)

### DIFF
--- a/pages/en/lb3/HasOne-relations.md
+++ b/pages/en/lb3/HasOne-relations.md
@@ -48,7 +48,7 @@ For example, consider two models: supplier and account.
 }
 ```
 
-A supplier has one account, where the foreign  key is on the declaring model: account.supplierId -> supplier.id.
+A supplier has one account, where the foreign  key is on the target model: account.supplierId -> supplier.id.
 
 {% include code-caption.html content="common/models/account.json" %}
 ```javascript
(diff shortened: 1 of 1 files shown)


dhmlau (approved):

dhmlau:
@juliandanciu , thanks for your contribution.  Your PR looks good to me.  
Before we can land, could you please sign the CLA? https://cla.strongloop.com/agreements/strongloop/loopback.io

@b-admike @jannyHou , would like another pair of eyes just in case I got the declaring & target model mixed up. :)  Thanks!

juliandanciu:
Hello Diana,

Thanks. I just signed the CLA.

Have a wonderful day.

jannyHou (approved):
🚢

dhmlau:
Thanks @juliandanciu for your contribution.  Your PR has landed! 🎉

PR #719, opened 2018-07-31
Updates aug1

bschrammIBM:
edits for adding v5 as the developer toolkit version

### DIFF
--- a/pages/en/lb4/Testing-your-application.md
+++ b/pages/en/lb4/Testing-your-application.md
@@ -163,6 +163,7 @@ sufficient.
 {% include code-caption.html content="test/helpers/database.helpers.ts" %}
 
 {% include code-caption.html content="test/helpers/database.helpers.ts" %}
+
 ```ts
 // ...
 export function givenProductData(data?: Partial<Product>) {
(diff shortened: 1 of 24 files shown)


pages/en/lb4/Testing-your-application.md
```diff
@@ -1,4 +1,4 @@
----
+Ok will push to your branch shortly.---
```
virkt25, line 1: LoopBack 4 documentation **should not** be updated in this repo as any changes here will be overwritten on deploy by the `@loopback/docs` package. 

It should be modified in the strongloop/loopback-next repository in the `docs/site` folder.

bschrammIBM:
adding my updates to accommodate that the apic commands only apply to v5

pages/en/lb4/Testing-your-application.md
bschrammIBM: I had many merge conflicts with this file: pages/en/lb4/Testing-your-application.md but I did not make any edits there so it doesn't matter if it gets overwritten. The edits that matter are in the 901c53a commit.

PR #715, opened 2018-07-26
Update LTS policy to adhere to Module LTS policy

bajtos:
cc @seabaylea

PR #720, opened 2018-08-01
LTS: rename LTS to Active LTS to avoid confusion

bajtos:
Rename "LTS" and "Maintenace" modes to "Active LTS" and "Maintenance LTS".

This is to avoid confusion, because "Module LTS" uses LTS for what we call Maintenance mode.

Also drop the minimum of 6 months in Maintenance mode and use Module LTS rules to control EOL of our maintenance.

See also https://github.com/strongloop/loopback.io/pull/715

### DIFF
--- a/pages/en/contrib/LTS.md
+++ b/pages/en/contrib/LTS.md
@@ -65,7 +65,7 @@ In August 2018, there were two major Node.js versions covered by the LTS plan:
 6.x and 8.x; and two supported LoopBack versions: 2.x and 3.x.
 
 Framework | Status | Published | EOL | Runtime | GA | EOL
--- | -- | -- | -- | -- | --
+-- | -- | -- | -- | -- | -- | --
 Loopback 3 | Current | Dec 2016 | Dec 2019 |  Node 8 | Oct 2017 | Dec 2019
 Loopback 2 | Active LTS | Jul 2014 | Apr 2019 | Node 6 | Oct 2016 | Apr 2019
  |  |  | Apr 2018 | Node 4 | Oct 2015 | Apr 2018
@@ -74,7 +74,7 @@ Assuming LoopBack 4.0 is released in October 2018 to align with Node.js 10.x
 entering LTS:
 
 Framework | Status | Published | EOL | Runtime | GA | EOL
--- | -- | -- | -- | -- | --
+-- | -- | -- | -- | -- | -- | --
 LoopBack 4 | Current | Oct 2018 | Dec 2021 | Node 10 | Oct 2018 | Apr 2021
 Loopback 3 | Active LTS | Dec 2016 | Dec 2019 |  Node 8 | Oct 2017 | Dec 2019
 Loopback 2 | Maintenance LTS | Jul 2014 | Apr 2019 | Node 6 | Oct 2016 | Apr 2019
--- a/pages/en/contrib/LTS.md
+++ b/pages/en/contrib/LTS.md
@@ -5,15 +5,15 @@ tags: [contributing, community]
 keywords: LoopBack, versioning, long-term support
 sidebar: home_sidebar
 permalink: /doc/en/contrib/Long-term-support.html
-summary: LoopBack maintains a current version, a long-term support (LTS) version, and a maintenance version.
+summary: LoopBack maintains a Current version, an Active long-term support (LTS) version, and one or more Maintenance LTS versions.
 ---
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


pages/en/contrib/LTS.md
```diff
@@ -32,8 +33,6 @@ Changes are limited to:
 
 2. Security updates;
 
-3. Non-semver-major npm updates;
-
 4. Relevant documentation updates;
```
virkt25: List should be updated here to be `3.` and 5. should be `4.`

pages/en/contrib/LTS.md
```diff
@@ -45,9 +44,12 @@ Changes are limited to:
...
  changes are only permitted if required for critical security and bug fixes.
 
-## Maintenance versions
+## Maintenance LTS
 
-When a new major version (for example, 4.0) is released, the oldest LTS version (for example, 2.x) enters maintenance mode, where it will stay at least for another six months and so long as the Node.js LTS versions available at release time are maintained by the Node.js project.
+When a new major version (for example, 4.0) is released, the oldest Active LTS
+version (for example, 2.x) enters Maintenance LTS mode, where it will stay for
+so long as the Node.js LTS versions available at release time are maintained by
+the Node.js project, but at least for six months.
```
virkt25: but at least for 6 months may possibly end up conflicting with Module LTS Policy :/ -- Which is what I think we should adhere for Maintenance LTS.

bajtos: > but at least for 6 months may possibly end up conflicting with Module LTS Policy

I don't understand. "at least for 6 months" means that we offer even longer support period than Module LTS Policy does. What's wrong with that?

Here is what I am trying to say here, I very welcome suggestion on how to describe this in a better way.

1. If the Node.js LTS version goes end of life after more than 6 months since a LoopBack release moved to Maintenance LTS mode, then the LoopBack version goes end of life at the same time as the Node.js LTS version.

2. If the Node.js LTS version goes end of life sooner than 6 months since a LoopBack release moved to Maintenance LTS mode, then the LoopBack version will stay in Maintenance LTS for 6 months. This typically happens when we don't release a new major version every 6 months, as is the case with LB 4.0.

Here is the reason why I think it's important to keep those 6 months at minimum: Right now, LoopBack 3.x is Current and 2.x is Active LTS. People preferring stability should be using 2.x. 

Now imagine we were releasing 4.0 in October next year (when Node.js 12 goes to LTS), so that 2.x was aligned with Node.js 8.x LTS which reaches end of life in December 2019 (a shorter LTS period than usual). If we were strictly following Module LTS policy, then 2.x would be in Maintenance LTS mode only for two months (between October and December), meaning that people running on 2.x would have only two months to migrate their existing applications from 2.x to 3.x. In my opinion, two months is not enough time for everybody to migrate all their existing code and that's why we should offer 6 months instead.

Thoughts?

virkt25: I guess the concern / my point of confusion comes from the following: 

LB 2.x is supported till Node 8.x (December 2019) **OR** 6 months after 4.0 release. If we release in October -- so in January do we support LB 2.x on Node 10.x? Do we even know if everything will work fine on Node 10.x with LB 2.x (unlikely but possible)? (This is the source of my concern ...). 

I think a way around my confusion will be to have a note that no version will ever support a version of Node that is no longer supported by Node.js.

bajtos:
Based on our discussion here and in the blog post review, and the fact that my example in https://github.com/strongloop/loopback.io/pull/720#discussion_r206899413 is purely hypothetical, I am going to drop the requirement to of a minimum period of 6 months in Maintenance LTS.

In practice, if we release at least one major version every year, then we should be good, because Node.js LTS offers 18 + 12 months of support:

See https://github.com/nodejs/Release#release-plan
> Every major version covered by the LTS plan will be actively maintained for a period of 18 months from the date it enters LTS coverage. Following those 18 months of active support, the major version will transition into "maintenance" mode for 12 additional months.

If we ever run into the situation when our users would have less than 6 months to upgrade from one LTS to the next one then we can deal with that issue then. No need to worry about it upfront.

bajtos:
@virkt25 @dhmlau I have removed the requirement to keep a version in Maintenance LTS for at least 6 months based on the comments above and in the blog post review - see a78f742.

Please review the new version and let me know if it's good to be finally landed.

dhmlau (approved):

virkt25 (approved):

PR #721, opened 2018-08-01
add a new line before end div tag

shimks:
Jekyll templating used to break when `note.html` or any other similar assets were used when not in this specific format:
```
{% include note.html content="
Binding configuration such as component binding,
provider binding, or binding scopes are not possible with the constructor-based
configuration approach.
" %}
```
The said note.html file (and other similar assets) had a `</div>` tag right after where stuff specified in `content` would go. If we were to introduce a line break here, the problem goes away for the oddest reason.

The PR introduces line breaks to all of the </div> tags that didn't have one before

### DIFF
--- a/_includes/callout.html
+++ b/_includes/callout.html
@@ -1,1 +1,3 @@
-<div markdown="span" class="bs-callout bs-callout-{{include.type}}">{{include.content}}</div>
+<div markdown="span" class="bs-callout bs-callout-{{include.type}}">
+  {{include.content}}
+</div>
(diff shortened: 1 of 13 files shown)


_includes/callout.html
```diff
@@ -1 +1,2 @@
-<div markdown="span" class="bs-callout bs-callout-{{include.type}}">{{include.content}}</div>
+<div markdown="span" class="bs-callout bs-callout-{{include.type}}">{{include.content}}
```
virkt25: Can you move `{{include.content}}` to it's own line.

shimks: done

virkt25 (approved):

dhmlau (approved):

PR #728, opened 2018-08-15
fix: advisory link

virkt25:
This is why one shouldn't copy & paste.


shimks (approved):
`This is why we shouldn't copy-paste with reckless abandon`
FTFY

PR #733, opened 2018-08-31
Update Extend-your-API.md  : access remote method through path as parameter source

ghost:
It is important to explain how to access the endpoint with parameter source as path because most likely people would want to use it that ways rather than a query string . 
":parameter_name" is the syntax and its important because otherwise it throws exception   which says get Method handling is not defined and developers may try to find the solution in the ACL which is not where the solution lies.

### DIFF
--- a/pages/en/lb3/Extend-your-API.md
+++ b/pages/en/lb3/Extend-your-API.md
@@ -112,6 +112,9 @@ For more information, see [Remote methods](Remote-methods).
 
 The `status` remote method is trivial, but a remote method can also access any of the standard model create, retrieve, update, and delete methods to perform data processes and validation.  Here is a simple example (this is not in the `loopback-getting-started` repository):
 
+
+## accessing the remote method through query as source parameter 
+
 ```js
 module.exports = function(CoffeeShop) {
 ...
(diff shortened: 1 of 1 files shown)


ghost:
Hi , made a pull request #733 to the repository loopback.io which was rejected due to the failed check
Commit Linter — commits in this PR have formatting problems . when i saw the console the problems found says :
2 problems found:
** 0207336 - Update Extend-your-API.md: line 3 longer than 72 characters.
** 0207336 - Update Extend-your-API.md: line 4 longer than 72 characters.

i havent made any changes to line 3 and 4 , so there is no way i can solve this ,
help me understand the reason for rejection as the contribution i am trying to make is really important , it took me 2 days to figure out a simple syntax error which was actually misguided by the compiler

pages/en/lb3/Extend-your-API.md
```diff
@@ -144,7 +147,34 @@ You'll get the response:
...
+
+```
+## accessing the remote method through path as source parameter 
+```
+
+ CoffeeShop.remoteMethod (
+        'getName',
+        {
+          http: {path: '/:id/getname', verb: 'get'},
+          accepts: {arg: 'id', type: 'number', http: { source: 'path' } },
```
bajtos, line 159: I find this rather unusual. If you want to write a method that's exposed at `/api/CoffeeShops/1/getname`, then the usual solution is to put that method on the model prototype.

```js
CoffeeShop.prototype.getName = function(cb) {
  // (you can access the id via "this.id")
  const response = "Name of coffee shop is " + instance.name;
  cb(null, response);
  console.log(response);
});

CofeeShop.remoteMethod('prototype.getName', {
  http: {path: '/getname', verb: 'get'},
  returns: {arg: 'name', type: 'string'},
});
```

This way LoopBack calls `findById` for you and takes care of edge cases like when there is no model instance with the given `id`.

bajtos (changes requested):
Hello @sifavahora, thank you for the pull request!

Your example is showing an approach that I am reluctant to advise to LB users.

At the same time, I do see how an example showing path parameters in practice can be useful. I see two use cases worth documenting:

- How to write a `prototype` method exposed at `/api/model-name/:id/method-name` endpoint. See my comment below for that example.
- How to write a static method accepting a parameter from the URL path. The parameter should be different from `id` and preferably go _after_ the method name in the URL path, for example `/api/my-model/findByName/:name`.

Could you please rework your pull request to document one (or both) of these approaches?

PR #734, opened 2018-09-06
Clarify relations change

jubstuff:
Since the doc is trying to explain how to deal with a custom user model, I think it would make sense to not use `user` with lowercase `u` to identify the custom user model

### DIFF
--- a/pages/en/lb3/Authentication-authorization-and-permissions.md
+++ b/pages/en/lb3/Authentication-authorization-and-permissions.md
@@ -99,7 +99,7 @@ and uses the built-in `AccessToken` model.  In this case, you need to change the
     "relations": {
       "user": {
         "type": "belongsTo",
-        "model": "user",
+        "model": "CustomUserModel",
         "foreignKey": "userId"
       }
     }
--- a/pages/en/lb3/Authentication-authorization-and-permissions.md
+++ b/pages/en/lb3/Authentication-authorization-and-permissions.md
@@ -99,7 +99,7 @@ and uses the built-in `AccessToken` model.  In this case, you need to change the
     "relations": {
       "user": {
         "type": "belongsTo",
-        "model": "CustomUserModel",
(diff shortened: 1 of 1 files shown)


pages/en/lb3/Authentication-authorization-and-permissions.md
```diff
@@ -99,7 +99,7 @@ and uses the built-in `AccessToken` model.  In this case, you need to change the
     "relations": {
       "user": {
         "type": "belongsTo",
-        "model": "user",
+        "model": "CustomUserModel",
```
virkt25: I think a better name might be `CustomUser` as that would make this consistent with the next code snippet.

dhmlau:
@jubstuff , could you please address the comment from @virkt25? 
There are character limits on the commit message, and that's why there's a commit linter error.  

> Summaries should be short (50 chars or less) and use normal sentence capitalization rules.
> The full commit message can contain more detailed explanatory text, if necessary.
Wrap it to about 72 characters or so. In some contexts, the first line is treated as the subject of an email and the rest of the text as the body. The blank line separating the summary from the body is critical (unless you omit the body entirely); tools like rebase can get confused if you run the two together.


Please refer to this commit message guideline for more details: https://loopback.io/doc/en/contrib/git-commit-messages.html. Thanks.

pages/en/lb3/Authentication-authorization-and-permissions.md
jubstuff: You're right @virkt25, I'll change that.

Sorry for being so late, I totally missed your comment :)

jubstuff:
Thanks for the reminder @dhmlau :)

I've applied that change and updated the PR.

dhmlau (approved):

dhmlau:
@jubstuff, thanks for your contribution.  Your PR has landed! 🎉

PR #741, opened 2018-09-18
Add DAO and relation promise support to documentation

charlie-s:
Pinging @bajtos and https://github.com/strongloop/loopback-datasource-juggler/issues/1127.

I'm still getting used to Loopback source, so bear with me. It looks like `HasManyThrough` extends `HasMany`, and itself isn't tested in `./test/relation.test.js`. So, I imagine it's safe to say `HasManyThrough` would use the same wording as `HasMany`. Can you confirm? 

Also, I know it's a separate issue, but why is there no mention of `HasManyThrough` in `relation.test.js`?

### DIFF
--- a/pages/en/lb3/BelongsTo-relations.md
+++ b/pages/en/lb3/BelongsTo-relations.md
@@ -66,7 +66,7 @@ For example, here is the model JSON file for the order model in
   ...
 ```
 
-Alternatively, you can define a "belongsTo" relation in code, though in general this is not recommended:
+Alternatively, you can define a `belongsTo` relation in code, though in general this is not recommended:
 
 {% include code-caption.html content="common/models/order.js" %}
 ```javascript
@@ -83,11 +83,13 @@ If you don't specify them, then LoopBack derives the relation name and foreign k
 
 ## Methods added to the model
 
-Once you define the belongsTo relation, LoopBack automatically adds a method with the relation name to the declaring model class's prototype,
+Once you define the `belongsTo` relation, LoopBack automatically adds a method with the relation name to the declaring model class's prototype,
 for example: `Order.prototype.customer(...)`.
 
 Depending on the arguments, the method can be used to get or set the owning model instance.
 The results of method calls are cached internally and available via later synchronous calls to the method. 
+
+The relation can return a promise by calling the `get()` method on the relation property.
 
 <table>
   <tbody>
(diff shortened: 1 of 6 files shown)


bajtos:
Thank you @charlie-s for the pull request ❤️  I am a bit overloaded now, I'll try to take a look on Friday or next week. Sorry for the delay!

charlie-s:
Thanks @bajtos, no rush.

bajtos:
> Also, I know it's a separate issue, but why is there no mention of `HasManyThrough` in `relation.test.js`?

I believe we do have a test suite for "HasManyThrough", it starts here:

https://github.com/strongloop/loopback-datasource-juggler/blob/fcac276955a58c4aabcd2b602af046f45cfc73fe/test/relations.test.js#L586

> I'm still getting used to Loopback source, so bear with me. It looks like `HasManyThrough` extends `HasMany`, and itself isn't tested in `./test/relation.test.js`. So, I imagine it's safe to say `HasManyThrough` would use the same wording as `HasMany`. Can you confirm? 

I believe this test is verifying that `find` works for HasManyThrough: https://github.com/strongloop/loopback-datasource-juggler/blob/fcac276955a58c4aabcd2b602af046f45cfc73fe/test/relations.test.js#L715-L740

So yes, it's safe to use the same wording for HasMany and HasManyThrough.

charlie-s:
My bad, I was searching for `hasManyThrough`. Thanks @bajtos.

bajtos (approved):
The content is technically correct.

I think the grammar & style could be improved a bit. @virkt25 @bschrammIBM could you please review and perhaps suggest better phrasing?

bajtos:
Sorry for the delay, @charlie-s. I have rebased your changes on top of the latest master and squashed all commits into a single one with a message following our conventions - see https://loopback.io/doc/en/contrib/git-commit-messages.html.

Let's wait for CI results before landing.

Once this patch is landed (but not before that!), you will need to hard-reset your local clone of the repository to make your `gh-pages` branch up to date with our `gh-pages` branch (the upstream).

1. If you don't have strongloop's repository configured as an upstream `remote` yet:
    ```
    $ git remote add upstream https://github.com/strongloop/loopback
    ```

2. Fetch the latest updates from `upstream`:
    ```
    $ git fetch upstream
    ```

3. Hard-reset your gh-pages to `upstream`:
    ```
    $ git checkout gh-pages
    $ git reset --hard upstream/gh-pages
    ```

4. Force-push your changes back to GitHub
    ```
    $ git push -f
    ```

In the future, please create a new feature branch for each pull request - see https://loopback.io/doc/en/contrib/code-contrib.html#submitting-a-pull-request

bajtos:
Landed 🎉  Thank you @charlie-s for your contribution! ❤️

PR #742, opened 2018-09-19
chore: remove api-explorer

virkt25:
- Removes api-explorer from loopback.io in favour for explorer.loopback.io. 
- Discussion: Keep this or remove it? The best solution I found to a redirect that would preserve the query param is https://maxchadwick.xyz/blog/preserving-the-hash-and-query-string-with-jekyll-redirects ... but it just becomes another thing to maintain. Thoughts?


raymondfeng:
We need to keep `https://loopback.io/api-explorer` up and running for some time so that existing apps won't be broken.

The JavaScript based redirect is fine with me too.

bajtos:
+1 to preserve `https://loopback.io/api-explorer` for a while and redirect it to the new URL.

I'd rather not rely on jekyll-redirect-from and use a static `index.html` file instead, one using the javascript-based redirect as described in the linked blog post.

dhmlau:
@raymondfeng @bajtos, so do you mean we're abandoning this PR?

bajtos:
I think this pull request is still relevant. We need to rework `https://loopback.io/api-explorer ` to redirect to `https://explorer.loopback.io` and preserve optional query string.

PR #747, opened 2018-10-05
Update ACL End-point

yashmurty:
ACL docs have wrong end-point for **replaceById**

It is mentioned as `/model-name-plural/{id}`.

But it should be `/model-name-plural/{id}/replace`.

Source code :
https://github.com/strongloop/loopback/blob/master/lib/persisted-model.js#L785


dhmlau:
@yashmurty, your PR has landed.  Thanks for your contribution!

yashmurty:
@dhmlau Always a pleasure! :smiley:

PR #748, opened 2018-10-06
Specify the rule for eternal access tokens

felladrin:
Currently, there's no documentation about the need of setting the access token with `ttl = -1` for it to be considered eternal, after setting `allowEternalTokens = true` on the User model.

Code reference on `access-token.js`: <https://github.com/strongloop/loopback/blob/master/common/models/access-token.js#L255> 

Maybe it could also help if the description says it should be set on the model the AccessToken model points to (which is usually an extension of the built-in User model).


felladrin:
I'm closing this PR, as it failed the checks.
Opening a new one in a moment.

PR #753, opened 2018-10-20
Update ACL property for create

yashmurty:
ACL docs have wrong property value for **create** endpoint `POST` `/model-name-plural`

It is mentioned as `createChangeStream`.

But it should be `create`.

Source code : 
https://github.com/strongloop/loopback/blob/master/lib/persisted-model.js#L662


dhmlau (approved):

dhmlau:
Thanks @yashmurty for your contribution.  Your PR has landed! 🎉

yashmurty:
@dhmlau Thanks. Always a pleasure! :)

PR #760, opened 2018-10-31
Add docs to cover filter object santization

raymondfeng:
See https://github.com/strongloop/loopback-datasource-juggler/pull/1652


b-admike (approved):
Nice 💯

PR #766, opened 2018-11-10
update governance page

dhmlau:
Related to https://github.com/strongloop/loopback-next/issues/1673

- Update maintainer list
- Remove functional owner page and its references, inform users to use CODEOWNERS file to determine the maintainers for the repo


pages/en/contrib/Governance.md
```diff
@@ -20,18 +20,22 @@ Each repository has one or more lead maintainers responsible for:
 *   Seeking consensus on technical decisions.
 *   Making the final decisions when consensus cannot be achieved.
 
-Maintainers have npm publishing rights for modules and the final say on releasing new versions.
+Maintainers have npm publishing rights for modules and the final say on releasing new versions.
```
bajtos: Hmm, I think this is not true. AFAIK, we don't give npm publishing rights to maintainers from the community. I am proposing to change "Maintainers" to "Lead maintainers" in this sentence to avoid possible confusion.

pages/en/contrib/Governance.md
```diff
@@ -20,18 +20,22 @@ Each repository has one or more lead maintainers responsible for:
 *   Seeking consensus on technical decisions.
 *   Making the final decisions when consensus cannot be achieved.
 
-Maintainers have npm publishing rights for modules and the final say on releasing new versions.
+Maintainers have npm publishing rights for modules and the final say on releasing new versions.  
 
-## Lead maintainers
+Refer to `CODEOWNERS` file in each repository for the list of lead maintainers.
```
bajtos: Let's use the term "active maintainers" instead of "lead maintainers". We list community members in CODEOWNERS too.

pages/en/contrib/Governance.md
```diff
@@ -20,18 +20,22 @@ Each repository has one or more lead maintainers responsible for:
...
+Refer to `CODEOWNERS` file in each repository for the list of lead maintainers. 
 
-LoopBack lead maintainers include:
+### Full-time maintainers
 
+Project architects:
 *   [Miroslav Bajtoš](https://github.com/bajtos)
 *   [Raymond Feng](https://github.com/raymondfeng)
-*   [Ritchie Martori](https://github.com/ritch)
-*   [Simon Ho](https://github.com/superkhau)
```
bajtos: Let's add a new section "Alumni" towards the end and list people that are no longer actively working on the project there.

pages/en/contrib/Governance.md
```diff
@@ -20,18 +20,22 @@ Each repository has one or more lead maintainers responsible for:
...
 
-For more details, see [Functional area owners](functional-area-owners.html).
+Full-time maintainers include:
```
bajtos: I think we should either:
- be explicit about the fact that this list contains people on IBM payroll only
- or include all current maintainers 

I feel the current proposal is making community contributors a kind of second-class maintainers.

It would be great if we could fetch an up-to-date list of people from somewhere else (or point readers to such source), e.g. GitHub's list of people in `@strongloop/loopback-maintainers` teams. Unfortunately, AFAICT, there is no such public view :(

pages/en/contrib/Governance.md
dhmlau: You're right.  
How to define/find out who are the "lead maintainers"? I'm thinking, as @jjtang1 suggested, whether it's better to have "core maintainers" (i.e. the core LB team) and "community maintainers"?

pages/en/contrib/Governance.md
dhmlau: One concern of adding the alumni section is that there is probably a much bigger list (it's likely we'll miss out a few?) that we try to keep track of it through the CODEOWNERS file.  
If we want to keep Ritchie and Simon on this page because they have significant contributions to LB, we can put it as "previous project architects"?

pages/en/contrib/Governance.md
dhmlau: thanks for pointing it out.  I didn't realize it will cause this impression!

dhmlau:
@bajtos, thanks for your feedback.  I've made some changes based on your feedback.
I'd like to propose to have 2 categories of maintainers:
- core maintainers, i.e. LoopBack core team (the 7 of us)
- community maintainers, i.e. maintainers, regardless IBMers or not, who have made substantial amount of contributions in LB

I am a bit reluctant to create the alumni list because:
1. the alumni list is a much bigger list that we are trying to keep track in the CODEOWNERS file already. 
2. while we value the contributions of the alumni maintainers had made, it may not be too relevant to community users who come to this page to find out the current active maintainers.

pages/en/contrib/Governance.md
```diff
@@ -14,25 +14,33 @@ LoopBack is an open, inclusive, and tolerant community of people working togethe
...
 
 *   Daily operations: approving pull requests, responding to new issues, guiding discussions, and so on.
 *   Seeking consensus on technical decisions.
 *   Making the final decisions when consensus cannot be achieved.
 
-Maintainers have npm publishing rights for modules and the final say on releasing new versions.
+Core maintainers have npm publishing rights for modules and the final say on releasing new versions.  
 
-## Lead maintainers
+Refer to `CODEOWNERS` file in each repository for the list of active core and community maintainers.
```
bajtos:
```suggestion
Besides core maintaners working on LoopBack full-time, there is also a growing group of voluntary maintainers from our community.

Refer to `CODEOWNERS` file in each repository for the list of all active maintainers. 
```

bajtos (approved):
LGTM.

PR #770, opened 2018-11-15
Remove update-v4-readmes.sh

bajtos:
loopback-next was recently changed to bundle all READMEs inside `@loopback/docs` package published to npm, see strongloop/loopback-next#2025. There is no need to copy v4 READMEs.


b-admike (approved):
LGTM 🎉

Issue #777, opened 2018-11-20
Scrollbar position on new theme


Tyler-RCSD:
## URL of the page 

https://loopback.io/doc/en/lb3/Defining-models.html

## Nature of the issue 

On screen widths larger than 1600 pixels the scrollbar overlaps the main content text.
![lb](https://user-images.githubusercontent.com/42674425/48778102-f5099f00-ec99-11e8-87d4-4a78921f3287.PNG)

## Suggested resolution

Revert theme or fix css.

dhmlau:
@nabdelgadir, could you please take a look? thanks.

dhmlau:
Should be fixed via #778. Could you please verify? Thanks.

Tyler-RCSD:
Much better. Thank you!

PR #778, opened 2018-11-21
fix scrollbar

nabdelgadir:
Connect to https://github.com/strongloop/loopback.io/issues/777.

Fixed scrollbar position for screen widths higher than 1600px, so it doesn't overlap with main content. Also, changed margins of the nav bar for some screen widths, so that the banner doesn't increase in height for smaller screen widths.


css/theme-v4.css
```diff
@@ -183,3 +195,21 @@ div.sidebarTitle, div.sidebarTitle a {
...
+    .col-md-3 {
+        width: 17%;
+    }
+}
+
+@media (min-width: 1880px) {
+    .col-md-3 {
+        width: 14%;
+    }
+}
```
dhmlau: add a new line please

_includes/topnav.html
```diff
@@ -12,7 +12,7 @@
                 <span class="icon-bar"></span>
             </button>
             <a class="fa fa-lg navbar-brand no_icon" href="{{base}}/doc/index.html">
-              <img class="no_icon" style="width:150px;" src="{{base}}/images/branding/logo/blue/loopback.svg">&nbsp;<span class="projectTitle"></span></a>
+              <img class="no_icon" style="width:144px;" src="{{base}}/images/branding/logo/blue/loopback.svg">&nbsp;<span class="projectTitle"></span></a>
```
dhmlau: is it something we've changed when changing to the new L&F? 
My concern is that changing the hard coded value may not solve the problem for different screen side and resolution.

nabdelgadir: I did add the hard coded value with the new L&F so that it would match v4.loopback.io, but I was running into an issue where the banner would increase in height (the picture attached) when the screen was too small. However, I changed it back to 150px to match again and changed the margins instead. 
screen shot 2018-11-21 at 12 49 37 pm

dhmlau (approved):

PR #779, opened 2018-11-22
Update dependencies

bajtos:
Run `bundle update` to update the Gem lock file and get rid of potential security vulnerabilities in outdated dependencies.


dhmlau (approved):
Not quite familiar with this file, but the changes look reasonable to me.  Thanks.

PR #785, opened 2018-12-10
contrib: how to report LB4 issues + cleanup

bajtos:
See https://github.com/strongloop/loopback-next/issues/1854

In this pull request, I am cleaning up the structure of "Reporting issues" and adding a new section with instructions for creating "steps to reproduce" for LoopBack 4.x applications.

### DIFF
--- a/pages/en/contrib/Reporting-issues.md
+++ b/pages/en/contrib/Reporting-issues.md
@@ -14,9 +14,10 @@ If you have a question about how to do something in LoopBack, follow these steps
 
 1. [Consult the documentation](../../) and the [API documentation](http://apidocs.strongloop.com).  **NOTE**: The search box above searches _both_ of these sites.
 2. If you don't find an answer to your question, then do one of the following:
-  - Search and then ask on the [developer forum / Google group](https://groups.google.com/forum/#!forum/loopbackjs).
-  - Search and then ask on  [StackOverflow](http://stackoverflow.com/questions/tagged/loopbackjs+or+strongloop?sort=newest&pageSize=50).
-  - Ask on [Gitter](https://gitter.im/strongloop/loopback).
+
+    - Search and then ask on the [developer forum / Google group](https://groups.google.com/forum/#!forum/loopbackjs).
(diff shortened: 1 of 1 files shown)


pages/en/contrib/Reporting-issues.md
```diff
@@ -85,18 +98,48 @@ When reporting bugs, please include at least all the following:
...
+  [Todo](https://github.com/strongloop/loopback-next/tree/master/examples/todo)
+  example should work well for most users.
+4. Modify the selected example project to reproduce the issue you are
+   experiencing. Add the required code and configuration. When adding new
+   dependencies, don't forget to save them to `package.json` too.
+5. If possible, encode steps to reproduce in an automated test we can run
+   via `npm test`.
+6. Commit the changes (`git commit`) and push the feature branch to GitHub
+   (`git push -u origin my-branch-name`)
+6. Create an issue in the appropriate GitHub repository with steps to reproduce
```
b-admike: Should we mention that they create it in either `loopback` or `loopback-next` if they don't know the appropriate repository?

b-admike (approved):

pages/en/contrib/Reporting-issues.md
bajtos: The repository names are already mentioned in "Report in appropriate channel". Do you think it's worth repeating that information again? (I am genuinely asking.) Maybe I can add a link to "Report in appropriate channel" here, so that people know where to find the repo name, but the information is not duplicated?

Also note that this list of steps is for LB 4.x and newer only, so it would be `loopback-next` by default (`loopback` is not an option here).

b-admike: Ah good point. I don't think it's worth repeating if that's the case. Adding a link to `Report in appropriate channel` sounds like a good middle ground.

jannyHou: a dup 6. here.
Should be 7.

jannyHou (approved):
The step number has a dup 6, other than that minor fix LGTM 😎 .

Issue #808, opened 2019-02-11
Search "sidebar" replacing with lb3


rahulrkr08:
## URL of the page 

https://loopback.io/search/?q=rest&sidebar=lb4_sidebar

## Nature of the issue 

Re-searching from search page, replacing sidebar option to lb3 instead of lb4 and results are coming again from lb3.

## Expected behavior

Keep the state of sidebar as "lb4_sidebar"if it is in query.

nabdelgadir:
Fixed via #815, thanks for the contribution!

PR #809, opened 2019-02-11
fix: set sidebar input field as per query value

rahulrkr08:
Updating sidebar input field setting from query value.

#808

### DIFF
--- a/_layouts/search.html
+++ b/_layouts/search.html
@@ -77,7 +77,10 @@
     }
 
     $('.search-bar').val(queryObj.q.replace('+', ' '));
-
+    if(queryObj.sidebar && queryObj.sidebar != "") {
+        $('input[name="search-bar"]').val(queryObj.sidebar);
+    }
+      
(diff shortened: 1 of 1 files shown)


dhmlau (approved):
I'd like to get at least one more reviews before landing.
@nabdelgadir, since you've worked with loopback.io recently, PTAL? Thanks.

nabdelgadir (commented):
@rahulrkr08 Thanks for the contribution! It's still redirecting to `lb3_sidebar` with this change for me. I see the problem is the query (https://github.com/strongloop/loopback.io/blob/335e83c7caf0b2f672c1e98190d0579af971cfa2/_layouts/search.html#L69) is changing to `sidebar` to `lb3_sidebar` with every new search.

rahulrkr08:
@nabdelgadir Sorry, my bad type error. Wrongly given input field name as 'search-bar' instead of 'sidebar'. I re-committed, please check.

nabdelgadir (approved):
Works now, thank you! 💯 

Can you please squash the commits and rebase your PR? And then I'll merge.

rahulrkr08:
@nabdelgadir I am facing some issue on merging commits, will raise new PR.

PR #815, opened 2019-02-18
fix: set sidebar input field as per query value

rahulrkr08:
Updating sidebar input field setting from query value.

#808


nabdelgadir (approved):
👍

PR #823, opened 2019-03-19
Add french language for LB4

Yaty:
Hello,

I finally got some time for LB4 :)
I'll start working & maintaining the French documentation if you're ok !

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -118,6 +118,7 @@ sidebars:
 - zh_lb2_sidebar
 - pt-br_lb2_sidebar
 - pt-br_lb4_sidebar
+- fr_lb4_sidebar
 
 # description: "Documentation for the LoopBack framework"
 # the description is used in the feed.xml file
--- a/_data/sidebars/fr_lb4_sidebar.yml
+++ b/_data/sidebars/fr_lb4_sidebar.yml
@@ -1,0 +1,450 @@
+## definition of `menuitem`===========
+# title: string
+# url: string
+# output: [ enum{web, pdf} ] (expressed as comma separated string in YAML)
+# children?: [ menuitem ]
+
+## top level property=================
+# entries: [ menuitem ]
+
+title: LoopBack 4
+url: /doc/fr/lb4/index.html
+translated: true
+children:
+
+  - title: 'Getting started'
+    url: /doc/fr/lb4/Getting-started.html
+    output: 'web, pdf'
+    translated: false
+
+  - title: 'Tutorials'
+    url: /doc/fr/lb4/Tutorials.html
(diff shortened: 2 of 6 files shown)


dhmlau:
Thanks @Yaty! :) 
@bschrammIBM, probably need your help to review.  
We usually have some text at the top of the community translated documentation that `This is a community translation into French. For the latest information, see the English version.`  @bschrammIBM can probably point to where/how it can be added. 

See https://loopback.io/doc/fr/lb2/ as an example.

Yaty:
This is already done here when the translation is marked as done : https://github.com/strongloop/loopback.io/blob/062c44542da4d29bfbd877f26e89b560b253bbfb/_layouts/translation.html#L16

bschrammIBM (approved):
approved, the disclaimer is inserted
However, I think you are in the wrong repo for updating LB4 docs.
@biniam Can you confirm where translations should be checked in for LB4 docs?

dhmlau:
@b-admike is the right github handle. :)     
Could you please help?  Thanks.

b-admike:
@bschrammIBM The docs content in loopback-next at https://github.com/strongloop/loopback-next/tree/master/docs/site are copied over to the `/pages/en/lb4` directory (see https://github.com/strongloop/loopback.io/blob/gh-pages/update-lb4-docs.js#L4-L5). So, if we want this patch to be contributed there, we would have to create a new directory structure there (perhaps under site we have `en` and `fr` in loopback-next) and modify the scripts to copy them over properly. @strongloop/sq-lb-apex @bajtos @raymondfeng thoughts?

bajtos:
@Yaty This is awesome ❤️ ❤️ ❤️ 

> So, if we want this patch to be contributed there, we would have to create a new directory structure there (perhaps under site we have en and fr in loopback-next) and modify the scripts to copy them over properly. @strongloop/sq-lb-apex @bajtos @raymondfeng thoughts?

We are keeping LB4 docs in loopback-next for two reasons:
 - To allow pull requests to update both the code and the relevant docs
 - To delay doc updates until the code changes are published to npm

IMO, the translations should be versioned independently of the main English content, it's unrealistic to expect loopback-next pull requests to update translated versions too. Therefore I think this repository (`loopback.io`) is the right place for LB4 translations.

b-admike:
@bajtos Thanks for the clarification and yea I agree we should put translations in this repo 👍. Thank you for the contribution @Yaty LGTM as well.

b-admike (approved):

dhmlau:
Merging this PR since we got approval from @b-admike and @bschrammIBM.
@Yaty, thanks for your contribution! 🎉

PR #844, opened 2019-06-04
update lb3 page version buttons to match lb4 page

nabdelgadir:
Before:
Screen Shot 2019-06-04 at 9 08 00 AM

After:
Screen Shot 2019-06-04 at 9 09 07 AM

### DIFF
--- a/_includes/overview/install.html
+++ b/_includes/overview/install.html
@@ -3,17 +3,17 @@
     <h2 class="c-heading -h1">The Node.js API Framework</h2>
     <div class="x-center">
       <div class="v-block">
-        <a class="v-button lb3">LoopBack 3<br>(Active LTS)</a>
+        <a href="/" class="v-button">LoopBack 4<br>(Active)</a>
         <ul class="v-descriptions">
-          <li><a href="getting-started">Get started</a></li>
-          <li><a href="doc/en/lb3/">Docs</a></li>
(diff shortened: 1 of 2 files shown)


emonddr (approved):

_includes/overview/install.html
```diff
@@ -3,17 +3,17 @@
     <h2 class="c-heading -h1">The Node.js API Framework</h2>
     <div class="x-center">
       <div class="v-block">
-        <a class="v-button lb3">LoopBack 3<br>(Active LTS)</a>
+        <a href="/" class="v-button">LoopBack 4<br>(Active)</a>
```
jannyHou, line 6: Just curious: why some link uses relative path like `/` and some uses the public url like `https://loopback.io/doc/en/lb4/Getting-started.html` =p

jannyHou (approved):
LGTM 👍 + a question.

_includes/overview/install.html
nabdelgadir: I don't know why, but for some it prepends the link with `/lb3` and for others it doesn't. 
Edit: e.g. ~it'll go to `/lb3/doc/en/lb4/Getting-started.html` instead of `/doc/en/lb4/Getting-started.html` when using the relative link.~ nvm figured it out, fixed!

jannyHou: @nabdelgadir Great thank you 👍

PR #857, opened 2019-07-08
Add user references and logos

dhmlau:
Updating the LB web site with our users (logos + testimonials). See details in https://github.com/strongloop/loopback-next/issues/3047.  Got the design from the designer (thanks Mihir!)

- The order of the logos is random. 
- I'm trying to make the logos more or less the same size, so the `width` value is not the same. If anyone has better ways, please let me know! 

There are 2 sections added in this PR.  
1. "who's using LoopBack"

screencap1

2. "what our users say"
screencap2

second half of the "what our users say" 
screencap3


dhmlau:
cc @marioestradarosa @vvdwivedi @samarpanB @kowsheek

jannyHou (approved):

raymondfeng (approved):
LGTM with my local preview. 

One minor thing - do we want to make the logos clickable to their site?

dhmlau:
> One minor thing - do we want to make the logos clickable to their site?

That's good point. Let me do it in a separate PR. Thanks!

Issue #889, opened 2019-09-27
Links targeting to MD files broken


lokeshmohanty:
## URL of the page 
https://loopback.io/doc/en/lb4/todo-tutorial-putting-it-together.html

## Nature of the issue 
Broken links

## Expected behavior
Links should target to html files

## Actual behavior
Links target to md files

dhmlau:
@lokesh1197, thanks for catching this. I've commented on your PR, please see https://github.com/strongloop/loopback.io/pull/890#issuecomment-536107281.  Thanks!

lokeshmohanty:
I have submitted a pull request here: https://github.com/strongloop/loopback-next/pull/3841
I will close the issue when it gets accepted

PR #890, opened 2019-09-27
Update expired .md links to .html links

lokeshmohanty:
Solves Issue: #889

### DIFF
--- a/pages/en/lb4/tutorials/todo/todo-tutorial-controller.md
+++ b/pages/en/lb4/tutorials/todo/todo-tutorial-controller.md
@@ -53,7 +53,7 @@ that a new controller instance is created for each request. As a result, we want
 to inject our `TodoRepository` since the creation of these instances is more
 complex and expensive than making new controller instances.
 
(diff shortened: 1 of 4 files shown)


lokesh1197:
Solves Issue: #889

dhmlau:
@lokesh1197, thanks for your PR.  I'm afraid that this is not the right place to put in the fix.  A few things to note:
- For the LoopBack 4 documentation, the pages are coming from https://github.com/strongloop/loopback-next/tree/master/docs.  They are all in .md files and will get generated and copied to loopback.io. 
- The documentation pages show broken links because the path to the .md file is not correct.
For example for https://github.com/strongloop/loopback-next/blob/master/docs/site/tutorials/todo/todo-tutorial-datasource.md, 

> In LoopBack 4, datasources can be represented as strongly-typed objects and
freely made available for [injection](Dependency-injection.md) throughout the
application. 

The link should be `../../Dependency-injection.md`. 

- As a reference, please see https://loopback.io/doc/en/contrib/doc-contrib.html.

With that, I'd like to suggest you to submit a PR in the `loopback-next` repo.  

Thanks again for your contribution!

lokeshmohanty:
@dhmlau Thanks. I will do that then.

PR #896, opened 2019-10-23
Change site analytics to GA

dhmlau:
We have an approved tracking ID for Google Analytics for the site analytics, so can switch back to it now! 
Thanks @DaveYVR!

Reviewers: We have done something similar before, see this PR https://github.com/strongloop/v4.loopback.io/pull/36 as a reference.

### DIFF
--- a/_includes/analytics.html
+++ b/_includes/analytics.html
@@ -1,16 +1,8 @@
-<script type="text/javascript">
-  /* Define digital data object based on _appInfo object */
-  window.digitalData = {
-    page: {
-      category: {
-        primaryCategory: 'IBM_Cloud_GlobalMarketing',
-      },
-      pageInfo: {
-        ibm: {
-          siteID: 'Cloud',
-        }
-      },
-    }
-  };
+<!-- Global site tag (gtag.js) - Google Analytics -->
+<script async src=“https://www.googletagmanager.com/gtag/js?id=UA-150726441-2“></script>
+<script>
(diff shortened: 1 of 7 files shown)


agnes512 (approved):

_includes/analytics.html
```diff
@@ -1,16 +1,8 @@
...
-        }
-      },
-    }
-  };
+<!-- Global site tag (gtag.js) - Google Analytics -->
+<script async src=“https://www.googletagmanager.com/gtag/js?id=UA-150726441-2“></script>
+<script>
+  window.dataLayer = window.dataLayer || [];
+  function gtag(){dataLayer.push(arguments);}
+  gtag('js', new Date());
```
nabdelgadir, line 6: I noticed in https://github.com/strongloop/v4.loopback.io/pull/36, there's `gtag('set', 'anonymizeIp', true);`. Is there a reason we're not using it here?

dhmlau: from the snippet I got from @DaveYVR, it doesn't have it now.

dhmlau: @nabdelgadir, I can confirm that we don't need the `anonymizeIp` line, because this is all applied via Tealium scripts.

nabdelgadir (approved):

agnes512:
also need to add `google_analytics id` to `config.yml`. see https://github.com/strongloop/loopback.io/pull/898

PR #903, opened 2019-11-05
Fixes inconsistent naming.

aleksspeaker:
After following all steps of the tutorial user ends up with `create` method name, not with the  `createTodo` one.
But there are no references of `createTodo` on previous pages/steps.
That's why it looks confusing.

### DIFF
--- a/pages/en/lb4/tutorials/todo/todo-tutorial-geocoding-service.md
+++ b/pages/en/lb4/tutorials/todo/todo-tutorial-geocoding-service.md
@@ -187,7 +187,7 @@ export class TodoController {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


dhmlau:
@aleksspeaker, thanks for your PR.  Our LB4 docs content is actually coming from the `loopback-next` repo.  For the page that you're changing, it should be https://github.com/strongloop/loopback-next/blob/master/docs/site/tutorials/todo/todo-tutorial-geocoding-service.md instead.  

Could you please submit a PR in the `loopback-next` repo instead? Thanks!! 

With that, I'd like to close this PR and looking forward to seeing your PR in the loopback-next repo!

aleksspeaker:
Woops. Sorry.
Re-submitting to the right repo...

PR #905, opened 2019-11-14
Add doc for fields with include of belongsTo

pdesgarets:
After having struggled for hours trying to understand why the relation was not being included...


dhmlau:
@jannyHou, I'm not familiar with the relations in LB3, could you please help reviewing this PR? Thanks.

jannyHou (approved):
@pdesgarets Thank you for the clarification! True the fields filter is applied first, then the inclusion.
LGTM

PR #936, opened 2020-03-02
fix: remove line break from "important" alerts

bajtos:
Before (notice the line break after the exclamation-mark-icon):

Screen Shot 2020-03-02 at 14 33 43

After:

Screen Shot 2020-03-02 at 14 34 03

The trick is to format templating directives `{%` and `%}` in such way that we do not emit an empty newline after `<i>` element. The original version was producing a new line, which was interpreted by the markdown parser as a paragraph delimiter.

### DIFF
--- a/_includes/important.html
+++ b/_includes/important.html
@@ -1,7 +1,6 @@
 <div markdown="1" class="alert alert-warning" role="alert">
   <i class="fa fa-warning"></i>
-  {% if include.title %}
-  <b>{{include.title}}</b>
-  {% else %}
-  <b>Important:</b>{% endif %} {{include.content}}
+  {% if include.title %}<b>{{include.title}}</b>
+  {% else %}<b>Important:</b>{% endif %}
+  {{include.content}}
 </div>


_includes/important.html
```diff
@@ -1,7 +1,6 @@
 <div markdown="1" class="alert alert-warning" role="alert">
   <i class="fa fa-warning"></i>
-  {% if include.title %}
-  <b>{{include.title}}</b>
-  {% else %}
-  <b>Important:</b>{% endif %} {{include.content}}
+  {% if include.title %}<b>{{include.title}}</b>
+  {% else %}<b>Important:</b>{% endif %}
+  {{include.content}}
 </div>
```
bajtos, line 6: I am using the same coding style as we have already in place for `note.html`, see here:

https://github.com/strongloop/loopback.io/blob/bfe2be34ed3148d08bd1b266eaaca75142a3b9d5/_includes/note.html#L1-L5

emonddr (approved):

dhmlau (approved):

jannyHou (approved):

PR #937, opened 2020-03-09
Design update

hacksparrow:
Updated design to make the UI neater.

The following are three screenshots for preview. To see the overall changes checkout branch `design-update-a`, start the website locally and browse.

---
**Old**
image

**New**
image

---
**Old**
image

**New**
image

---
**Old**
image

**New**
image


bajtos:
Nice :+1: I like how the top-bar is highlighted and set apart from the rest of the content.

I find it a bit difficult to do a meaningful review, it feels like a "find five differences" exercise from a kids magazine. Can you please provide a short list of changes you are proposing?

In the first two examples, the old version is highlighting the top-bar entry which triggered a drop-down selector ("Translations"), but your new version does not show the highlight. Can you please fix that?

Still in those first two examples, I noticed that you changed the style of the note "LoopBack is a highly-extensible, (etc.)" and removed the frame & shadow cast. Is that an intentional change? Personally, I like the old (current) style more.

Screen Shot 2020-03-09 at 09 18 41

hacksparrow:
> Can you please provide a short list of changes you are proposing?

1. Navigation bar is highlighted and set apart from the rest of the content.
2. Bottom border removed from Navigation bar.
3. Top-level navigation links (nav bar, side menu) are uppercased to hint hierarchy.
4. Drop down menu pop under the navigation bar rather than on it.
5. Clicked drop down menu highlighted more subtly.
6. Search box made more subtle and seamless.
7. Side menu mouseover, active link highlighter colors made softer.
8. Summary box ("LoopBack is a platform for building APIs and microservices in Node.js") toned down. It was competing with the page contents for attention with its "3d effects" and looked out of place. I am even in favor of getting rid of it.
9. Made the notice box ("LoopBack won API World's 2019 Best in API Middleware category, see the announcement post.") look like a notice box, so it does not look like documentation content. We should remove these notifications after some time. Let them be on the main site, documentation section should focus on documentation.
10. Did some theming on the code snippet and command line code sections. Too much gray makes things look dull.
11. Did some theming on tables. Now have a dark blue header (older design had darl grey). Center aligned them and removed border to make them fit more seamlessly.
12. Resized and recolored (to black from grey (lower size headers)) header (H1, H2 etc.) tags to make them fit more aesthetically with the documentation text content.
13. Made "Page content" section to stretch out fully, and themed to make it fit more naturally on the page.
14. Resized and repositioned the "external link indicator". It was even bigger then regular text. It should just provide a subtle hint, not become the center of attention.
15. "Edit this page" and tags turned into text links. We don't need them to be buttons.
16. "LoopBack 4", "LoopBack 3.x" on the side menu made black and no underline on mouseover. 
17. "seealso" box made to stretch out. Color themed to be more subtle and pleasing.
18. "Tip" box themed to look like it may contain some tip. 
19. Horizontal line at the bottom made more subtle.
20. "StrongLoop by IBM" image resized not to look blurred and out of place.
21. Removed highlights from inline `code` elements, made them still stand apart by use of color and padding.
22. Black text (was light grey) on drop down menu.

> but your new version does not show the highlight. Can you please fix that?

Fixed.
image

> Is that an intentional change?

Intentional. Let's not use shadows anywhere, I explained above why I made this change. Remove the italic styling based on your feedback.

image

Aesthetically simple is the design mantra. There is much more to do, but one step at a time.

css/theme-v4.css
```diff
@@ -1,13 +1,14 @@
 body {
-    --color-blue: #009cff;
+    --color-blue: #c3ccff;
```
dhmlau: Any reasons we want to change `color-blue` and `color-blue-xd`? These are coming from the designer and are used in other blueprint for the web pages.

dhmlau (commented):
@hacksparrow, I like the new design!

Here are a few more comments. I might be easier for me to point them out from the screenshot than in code :) 
Screen Shot 2020-03-09 at 12 51 13 PM

1. It would be good to have the navbar item color change when mouse hover or click. 

2. Same for @bajtos, I'd prefer to have #2 box highlighted. Having said that, looking through the pages, there's not much value to have that description, because the title should be enough describing the purpose of the page. Can be out of scope for this PR. 

3. `Page content`: ideally it might be good to put it on the right as a side bar, like https://angular.io/start. But again, out of scope for this PR.

4. `Tip`: the green color doesn't quite match the theme. But I don't have a good suggestion.  Going to https://www.color-hex.com/color/3f5dff, I'm not sure the yellow `#ffe13f` is any better. :). 

5. Code snippet: is it possible to make the text white instead of blue? It will have a bigger contrast and easier to read.

dhmlau:
cc @achrinza. You might be interested in this PR

hacksparrow:
2. Let's get rid of that. Its presence is inconsistent; present on some pages and absent on most, and doesn't serve much purpose. And, design wise we should definitely not have anything popping out like that.
3. Good idea. A separate PR.

Rest I will implement.

css/theme-v4.css
hacksparrow: Undoing.

hacksparrow:
1: Updated.

4 and 5:

image

achrinza:
All the changes are definitely a significant step-up from the previous design 👍 

I haven't been able to setup the docs locally yet, but would the yellow `tip` collide with the `warning` banner?

hacksparrow:
Yes, it looks "warning-ish". That's why I had it green. @dhmlau.

hacksparrow:
A slightly different green box:

image

@achrinza @dhmlau thoughts? This one's not commit and pushed, btw.

dhmlau:
@achrinza, thanks for bringing up the different use of the color (tips, warning, note).

@hacksparrow, the new green seems to be better. Like I mentioned, I don't have a good suggestion. :)
Another suggestion is (I saw it somewhere before but can't remember where):
Screen Shot 2020-03-10 at 1 33 27 PM

I don't have a strong preference. Might want to get @raymondfeng's review as well. Thanks.

dhmlau:
@achrinza, I was having problems in viewing the changes locally as well.  @hacksparrow might have a better way but here is my (hacky) way:
- run the preview in loopback-next (so that you can see the LB4 docs)
- under `docs`, `_preview` folder is created. I copy the changes in this PR over there.

hacksparrow:
Mandatory commands, in order:

```sh
$ bundle install
$ npm i
$ npm run fetch-readmes
$ npm start
```

hacksparrow:
> Screen Shot 2020-03-10 at 1 33 27 PM

There are some more inconsistent formatting, let's handle them in another PR. It would be great if you can find where you saw it, though.

hacksparrow:
Updated PR to use the new green.

hacksparrow:
Also proposing to get rid of summary boxes. They are redundant and add noise to the page. And majority of the pages don't have them.

**As seen in the old design**
Screenshot 2020-03-11 at 12 46 10 PM

**As seen in the new design**
Screenshot 2020-03-11 at 12 46 55 PM

hacksparrow:
Updated PR to remove summary boxes (only for LB4 docs).

_Justification_: Their presence is inconsistent (absent on the vast majority of pages), the content is redundant, and they are a distraction on the page.

**With summary box**
image

**Without summary box**
image

dhmlau:
@hacksparrow, I have a few more comments:
- the color in the code snippet is still light blue. could you please update it?
- the summary box is still here. Are you intended to remove it? 
- if we're going to keep the summary in LB3 docs, could you please add back the border? 

Thanks.

hacksparrow:
@dhmlau did you get the latest update?

Btw, I made the text color much lighter than what it was originally. Compare with the original https://github.com/strongloop/loopback.io/pull/937#pullrequestreview-371362096. It won't be 100% white, though. Too much contrast is no good.

Also, can you post screenshots along with your comments, so I get more context?

dhmlau:
_I think_ I got the latest changes, because when I pull, there's no more updates. 

Taking http://127.0.0.1:4001/doc/en/lb4/todo-tutorial-geocoding-service.html as an example, here is what i mean:
1. The text in the code snippet section is still showing light blue for me (See the 2nd code block).  I changed it on the browser to set `color` to white on the 1st code block.

Screen Shot 2020-03-12 at 12 33 58 PM

2. For summary box, do you mean the grey text in the screen shot below? I thought you meant to remove it for LB4 docs: 
Screen Shot 2020-03-12 at 12 35 13 PM

 
3. I'm good to remove the summary box for LB4 and keeping it for other versions of LB, or even removing it altogether. Personally, I'd prefer we do some formatting around that, instead of just removing the border.  For example like below (compared to the above screen shot):
Screen Shot 2020-03-12 at 12 39 30 PM
But we can work on it in a separate PR.

4. Could you please squash your commits before merging? Thanks.

hacksparrow:
Could be some sort of caching going on.

No more summary box here:

image

bajtos:
>  Let's not use shadows anywhere, I explained above why I made this change. Remove the italic styling based on your feedback.

Fair enough :+1:

> A slightly different green box:

Can we perhaps ask the designer who created the original LB4 design to help us come up with a good color for tips, one that will match the overall blue theme?

> Updated PR to remove summary boxes (only for LB4 docs).
> 
> Justification: Their presence is inconsistent (absent on the vast majority of pages), the content is redundant, and they are a distraction on the page.

Should we also remove the summary front-matter fields from the pages that do provide it? I mean with your changes is place, is the summary text used anywhere else? If it's not, then I think it's best to remove it altogether, to avoid possible confusion.

> https://github.com/strongloop/loopback.io/pull/937#issuecomment-596432879

Thank you @hacksparrow for the list of changes, it's a long one! I am afraid I don't have bandwidth to participate on further reviews & discussions, I'll leave it up to others to approve the changes.

I am looking forward to see the new design deployed live 🚀

dhmlau:
> Can we perhaps ask the designer who created the original LB4 design to help us come up with a good color for tips, one that will match the overall blue theme?

Good point. @hacksparrow, let me send you the designer's contact. Based on the blueprints his team provided in the past, could you please ask him what we should do with the tips, warning, note, etc.  https://github.com/strongloop/v4.loopback.io/tree/master/blueprint
Thanks!

css/customstyles.css
```diff
@@ -1284,16 +1255,18 @@ div.box.box1 {
...
-  width: 680px;
+  margin: 10px auto;
   padding: 10px;
 }
 
-pre.highlight {margin-top: 0;}
+pre.highlight {
+    margin-top: 0;
+    padding: 16px;
+}
```
dhmlau, line 1269: if i add `color: white;` to `pre.highlight`, the code snippet's text will become white again.

dhmlau:
@hacksparrow, I tried to copy over your changes to a brand new branch, and still didn't see the changes that you were referring to.  I'm thinking, since your changes in this PR look good in general, to move this PR forward, we can:
1. remove the `summary` in docs in loopback-next (@raymondfeng is planning to publish a release early next week, so we might want to take this window to have this PR landed first)
2. fix the code snippet text color to white
3. ask the designer on the background color for tips/note/warning. Ideally, we can wait for the designer if we just need the rgb color from them. But I'm fine with landing this PR and improve the color as the follow up PR. 

WDYT?

dhmlau (approved):

PR #954, opened 2020-03-25
feat: refactor user testimonial page

dhmlau:
There are increasing number of user testimonials, so I think it might be better to move to another page. 

In index.html, removing "what our users says" section but add a button: 
Screen Shot 2020-03-25 at 7 03 35 PM

"What our users say page:
Screen Shot 2020-03-25 at 7 06 59 PM

Since the length of the testimonials are not the same, so the "cards" are a bit uneven
Screen Shot 2020-03-25 at 7 07 34 PM

### DIFF
--- a/css/users.css
+++ b/css/users.css
@@ -1,0 +1,61 @@
+.r-masthead {
+    text-align: center;
+    color: var(--color-white);
+    background-image: linear-gradient(to bottom, #009cff, #3f5dff);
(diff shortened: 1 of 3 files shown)


raymondfeng (approved):
Nice 👍

agnes512 (approved):
Thanks for the change!

PR #959, opened 2020-04-03
fix: overflowing column in smaller window size

hacksparrow:
Fixed overflowing column in smaller window size.

image


dhmlau (approved):
LGTM. I've verified the fix worked for me. Thanks.

bajtos:
Thank you @hacksparrow for fixing the problem! ❤️

PR #966, opened 2020-04-15
reusable navbar template

dhmlau:
Before making changes to navbar, it's better to make it reusable so that we only need to change it in one file instead of multiple files.  This PR contains:
- adding navbar-template.html - navbar template that is extracted from other pages
- modifying the *.html to use the template
- updating the jquery library. The "slim" version won't work, so use the "min" version instead.

To test locally, you need to start the server (run `npm start`), instead of viewing it as a file because it has jQuery scripts.


navbar-template.html
```diff
@@ -0,0 +1,36 @@
...
+            <li class="nav-item">
+                <a class="nav-link" href="https://strongloop.com/strongblog/tag_LoopBack.html" target="_blank" rel="noopener noreferrer">Blog</a>
+            </li>
+            <li class="nav-item">
+                <a class="nav-link" href="https://ibm.biz/apitest" target="_blank" rel="noopener noreferrer">Test Your APIs</a>
+            </li>
+        </ul>
+
+    </div>
+</nav>
```
nabdelgadir: add new line here

nabdelgadir (commented):
Tested it out locally, and it seems like it's no longer changing the colour for the page you're on.

See `Getting Started` is darker in the current one:

Screen Shot 2020-04-16 at 1 48 46 PM

Screen Shot 2020-04-16 at 1 49 31 PM

nabdelgadir (approved):

PR #971, opened 2020-04-22
remote remote connector from LB4

dhmlau:
Since remote connector is not supported in LB4, we're removing it from the LB4 docs. 

Another PR will be submitted in loopback-next on the sidebar.


update-readmes.sh
```diff
@@ -28,7 +28,6 @@ strongloop loopback-connector-openapi master
 strongloop loopback-connector-oracle master
 strongloop loopback-oracle-installer master
 strongloop loopback-connector-postgresql master
-strongloop loopback-connector-remote master
```
agnes512: Will this affect LB3 docs https://loopback.io/doc/en/lb3/Remote-connector.html ?

dhmlau:
cc @bajtos

update-readmes.sh
dhmlau: Taking a closer look at the script, it actually will affect LB3 readme too. Let me remove it. Thanks.

agnes512 (approved):

bajtos (approved):
LGTM, thanks!

PR #989, opened 2020-05-23
Fixes #986

kaim01:
hope this will work


dhmlau (approved):
Thanks @kaim01. It seems to be working for me.

PR #1001, opened 2020-06-24
fix: fix update docs script

dhmlau:
Fix `update-lb4-docs` scripts after the restructuring of the sidebar.  There are two issues:
- The `Connector references` no longer at the top level, and it's now under `References guides`. So need to update how we're retrieving the children of the connector references.
- Not all the sidebar items have `url`, so need to add a check in there.


raymondfeng:
Close it to favor #1002

PR #1020, opened 2020-08-05
configure RenovateBot

dhmlau:
Using https://github.com/strongloop/loopback-next/blob/master/.renovaterc.json as the reference.

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -1,0 +1,39 @@
+{
+    "extends": [
+      ":gitSignOff",
+      "group:recommended",
+      "helpers:disableTypesNodeMajor"
+    ],
+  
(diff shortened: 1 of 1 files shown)


.renovaterc.json
```diff
@@ -0,0 +1,45 @@
...
+      ":gitSignOff",
+      "group:recommended",
+      "helpers:disableTypesNodeMajor"
+    ],
+  
+    "ignorePaths": [
+      "**/node_modules/**",
+      "**/__tests__/**",
+      "**/test/**"
+    ],
```
raymondfeng: I don't think we need to have `ignorePaths` section as this is not a monorepo.

raymondfeng (approved):

PR #1031, opened 2020-08-06
enable package-locks

dhmlau:
Enable package-locks

cc @raymondfeng

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,3 @@ node_modules
 .jekyll-metadata
 vendor
 .bundle
-package-lock.json
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,2 @@
-package-lock=false
+package-lock=true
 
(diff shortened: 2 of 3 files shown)


raymondfeng (approved):

raymondfeng:
DCO is missing. After this PR, we probably should also use `npm ci` instead of `npm install` in other scripts.

dhmlau:
> After this PR, we probably should also use npm ci instead of npm install in other scripts.

@raymondfeng, do you mean this https://github.com/strongloop/loopback.io#incorporating-external-readmes?

raymondfeng:
> @raymondfeng, do you mean this https://github.com/strongloop/loopback.io#incorporating-external-readmes?

Please note `npm ci` uses `package-lock.json` to ensure only versions in the lock file are used for repeatable behaviors. In addition to docs, we should check `*.js` files in this repo and `loopback-next/docs` repo to see if we use `npm i` for `loopback.io`.

dhmlau:
I checked the js files at the root level and in `js` folder, I don't see them using `npm install`.

PR #1079, opened 2020-10-07
feat: add expand by default

jannyHou:
Signed-off-by: jannyHou <juehou@ca.ibm.com>

connect to https://github.com/strongloop/loopback-next/issues/6477

Add a configure field called `expand` to control expand or hide a sidebar item by default. 
Add a new class `tree-parent-expand` in the css file to decorate the item set with `{expand: true}`.
The customjs file expands children for item decorated by `tree-parent-expand`.

Screen Shot 2020-10-07 at 3 08 31 PM

### DIFF
--- a/_includes/sidebar_menu_list_items.liquid
+++ b/_includes/sidebar_menu_list_items.liquid
@@ -1,8 +1,8 @@
 {% assign pageurl = page.url | split: "/" %}
 {% for item in include.menuitems %}
-  {% comment %} Is this supposed to be ouput on the `web`site?{% endcomment %}
+  {% comment %} Is this supposed to be output on the `web`site?{% endcomment %}
   {% if item.output contains "web" %}
-  <li class="{% if item.children.size > 0 %} tree-parent {%else%} leaf-node {% endif %} {% if pageurl.last == item.url %}active{% endif %}">
+  <li class="{% if item.children.size > 0 %} {% if item.expand == 'true' %} tree-parent-expand {%else%} tree-parent {% endif %} {%else%} leaf-node {% endif %} {% if pageurl.last == item.url %}active{% endif %}">
     {% if item.children.size > 0 %} <a class="show-hide" href="#"></span> {% endif %}
     {% if item.url %}
       <a class="page-link" href="{{item.url}}">{{ item.title }}</a>
--- a/css/customstyles.css
+++ b/css/customstyles.css
@@ -767,6 +767,10 @@ ul#mysidebar, #mysidebar ul{
   display:block;
 }
 
+.tree-parent-expand.expanded .nav-list{
+  display:block;
+}
+
 #mysidebar li.active a.show-hide,
 #mysidebar a.show-hide:hover,
 #mysidebar a.show-hide:active{
--- a/js/customscripts.js
+++ b/js/customscripts.js
@@ -56,9 +56,14 @@ $(function() {
 function setupSidebarTreeNav(){
 
     //hide all non-active nav-lists:
-    $('.nav-list').not('#mysidebar').each(function(i, list){
-      if(!$(list).parent().is('.active')){
-        $(list).hide();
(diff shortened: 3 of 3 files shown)


jannyHou:
I checked the sidebar display on phone, when expand, it should be like the following picture. Since "Building applications" and "Extending lb4" only have 2-3 children menus, I believe it won't affect the mobile UX. The whole menu can still display in one page without scrolling.

unnamed

dhmlau:
@jannyHou, how to test certain sidebar menus are expanded by default?

pages/en/lb4/sidebars/lb4_sidebar.yml
```diff
@@ -23,610 +23,683 @@ children:
   url: Inside-LoopBack-Application.html
   output: 'web, pdf'
 
-- title: 'Tutorials'
```
dhmlau: I believe this file is not meant to be pushed?

hacksparrow:
Can they be collapsed? If not, we should use the same `.` bullet like those of single pages instead of `-`.

jannyHou:
@hacksparrow 

> Can they be collapsed? If not, we should use the same . bullet like those of single pages instead of -

Yes they can be collapsed. The only difference is they are expanded by default. Actually only item with children menu will be decorated with `tree-parent-expand`.

pages/en/lb4/sidebars/lb4_sidebar.yml
jannyHou: @dhmlau Yes I change it here for preview. The PR to change sidebar titles should be in lb-next.

hacksparrow (approved):
👍

raymondfeng:
@jannyHou Does your PR expand all 2nd level items by default?

jannyHou:
@raymondfeng 
> Does your PR expand all 2nd level items by default?

No it shouldn't. Only the two 1st level items ("Extending LoopBack 4 " and "Building Applications") are expanded.

raymondfeng:
@jannyHou I feel it's better to expand more, such as:

- Tutorials
- Concepts
- Reference guides

jannyHou:
@raymondfeng Hmm, that would make the menu super long, longer than one page, e.g.:

Screen Shot 2020-10-09 at 4 37 08 PM
Screen Shot 2020-10-09 at 4 37 16 PM

Especially for mobile users.

raymondfeng:
@jannyHou I didn't mean to expand all levels. If we limit the expansion up to 2 levels, how will it look like?

jannyHou:
@raymondfeng The screenshot is expanding:

- Building Applications --> Tutorials
- Reference Guides

So they don't exceed 2 levels. And lots of 2nd level items haven not expanded yet.

bajtos (approved):
LGTM as far as I am concerned 👍🏻

raymondfeng:
@jannyHou Please rebase to fix the conflict of `pages/en/lb4/sidebars/lb4_sidebar.yml`

raymondfeng:
@jannyHou I checked out your branch and run the following commands:

```sh
rm -rf node_modules
npm i
npm run build
npm start
```
http://127.0.0.1:4001/doc/en/lb4/ does not show sidebar being expanded. Not sure why.

jannyHou:
@raymondfeng I reverted the change in `pages/lb4/sidebar/sidebar.yml` file cuz it should happen in https://github.com/strongloop/loopback-next/pull/6553, that's why you don't see the items expand.

I had a talk with @dhmlau and she says even with the sidebar change hers doesn't expand too. Investigating.

dhmlau:
Just want to clarify, @jannyHou, when there was a sidebar.yml in this PR, I did a local preview but didn't see any expanded items, but I didn't look whether `expand` boolean exists in that file either.

dhmlau (approved):
LGTM. I've verified locally.
Please remember to remove the *lb4_sidebar.yml files before merging. Thanks.

PR #1128, opened 2021-01-07
feat: use github actions to build and deploy

raymondfeng:
Signed-off-by: Raymond Feng <enjoyjava@gmail.com>


.github/workflows/site-build.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Install and Build 🔧
+        run: |
+          bundle install
+          npm ci
+          npm ls @loopback/docs 2>/dev/null || true
+          npm run build
+
+      - name: Deploy 🚀
+        uses: JamesIves/github-pages-deploy-action@3.6.2
```
bajtos, line 36: Do we need the deploy step? Isn't GitHub already deploying changes automatically for us?

bajtos: See https://github.com/strongloop/loopback.io/settings

Screen Shot 2021-01-08 at 10 42 07

bajtos: BTW I found it always annoying that this repo is using `gh-pages` as the main branch. It would be great to eventually change the configuration to use `main` (or `master`) instead. Such effort is out of scope of this pull request, I am just mentioning it here because it's related to the deployment process.

raymondfeng: IIRC, `gh-pages` is the default branch for github pages.

raymondfeng: > Do we need the deploy step? Isn't GitHub already deploying changes automatically for us?

I'm not sure. The `.travis.yaml` has the `deploy` task. See https://github.com/strongloop/loopback.io/pull/1128/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485L26

dhmlau (approved):

.github/workflows/site-build.yml
bajtos: > IIRC, `gh-pages` is the default branch for github pages.

Yes, it's the default, but the value can be changed. In the past, GitHub mandated `gh-pages` names for certain types of repositories, but that restriction was removed some time ago.

bajtos (approved):
I don't have any more meaningful feedback to offer. I guess it's time to land this PR as the first step towards GHA and then fix any issues we discover.

PR #1145, opened 2021-02-22
show pictures correctly

skypesky:


skypesky:
# What changes have I made?
Let the picture display normally, a very small change.

PR #1167, opened 2021-03-22
fix: fix docs deploy

dhmlau:
Signed-off-by: Diana Lau <dhmlau@ca.ibm.com>

### DIFF
--- a/.github/workflows/site-build.yml
+++ b/.github/workflows/site-build.yml
@@ -28,15 +28,13 @@ jobs:
       - name: Install and Build 🔧
         run: |
           bundle install
-          npm ci
+          npm install
           npm ls @loopback/docs 2>/dev/null || true
           npm run build
-
-#     - name: Deploy 🚀
-#       uses: JamesIves/github-pages-deploy-action@3.6.2
-#       with:
-#         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-#         BRANCH: gh-pages
-#         FOLDER: _site
-#         GIT_CONFIG_EMAIL: slnode@ca.ibm.com
-#         GIT_CONFIG_NAME: StrongLoop Bot
+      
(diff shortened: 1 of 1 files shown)


.github/workflows/site-build.yml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - gh-pages
+  pull_request:
```
dhmlau: this is for testing.

dhmlau:
@raymondfeng @achrinza, I tried it with my own fork that the only change that's needed to fix the LB4 docs issue is to use `npm install` instead of `npm ci`, so I'm removing the deploy portion of it.

.github/workflows/pr-build.yml
```diff
@@ -25,9 +25,9 @@ jobs:
...
-          bundle install
-          npm install
-          npm ls @loopback/docs 2>/dev/null || true
-          npm run build
+#      - name: Install and Build 🔧
+#        run: |
+#          bundle install
+#          npm install
+#          npm ls @loopback/docs 2>/dev/null || true
+#          npm run build
```
raymondfeng: I wonder why you commented the build step for the PR. The commands will NOT run with gh-pages branch and they should not mess up.

dhmlau: you're right. At first, i thought we may not need this but I guess it's needed to make sure the web site can build. Will add it later.
I tested it again with the deploy action. https://github.com/dhmlau/loopback.io/blob/gh-pages/.github/workflows/site-build.yml#L35-L40. Will do more testing again on my repo first and update this PR.

dhmlau:
Tested the changes with my fork: https://github.com/dhmlau/loopback.io/tree/gh-pages/.github/workflows.

raymondfeng (approved):

PR #1205, opened 2021-06-28
Propose of translated version to spanish

guidomaiola:


dhmlau:
@guidomaiola, thanks for the contribution. I'm wondering whether you meant to translate LoopBack 4 instead of LoopBack 2? Please keep in mind that LoopBack 2 has reached end-of-life.  Looking at your changes, it seems like you're referring to LoopBack 4. And the corresponding file is https://github.com/strongloop/loopback.io/blob/gh-pages/pages/es/lb4/index.md.

dhmlau:
@guidomaiola, since LB2 has reached end of life, we won't be accepting pull requests. So I'm closing this PR. Looking forward to your translation PR for LoopBack 4. Thanks.

PR #1333, opened 2022-03-31
chore: update changelog manually

dhmlau:
Signed-off-by: Diana Lau <dhmlau@ca.ibm.com>

For some reason, the latest `@loopback/docs` is not being picked up during build, so the changelogs aren't up-to-date. As an interim solution, i updated it manually by running the following command:
```
npm install @loopback/docs
npm run copydocs
```


achrinza (approved):
We may be able to leverage renovate and an on-push workflow to automate this.

PR #1382, opened 2022-09-10
docs(typeorm): add info for registering connection

bhavyasf:
I tried using `@loopback/typeorm` locally and suddenly getting an error because connection was not coming Then I got an idea to register the connection somewhere in `application.ts` which was also missing in the doc. Hence the PR.

Signed-off-by: bhavyasf <37434313+bhavyasf@users.noreply.github.com>

### DIFF
--- a/pages/en/lb4/readmes/loopback-next/extensions/typeorm/README.md
+++ b/pages/en/lb4/readmes/loopback-next/extensions/typeorm/README.md
@@ -58,12 +58,25 @@ export const SqliteConnection: ConnectionOptions = {
   synchronize: true,
 };
 ```
-
 Make sure to install the underlying database driver. For example, if you are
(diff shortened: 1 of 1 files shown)


achrinza:
Hi @bhavyasf, thanks for the PR!

Unfortunately LB4 docs are not edited here, but rather at https://github.com/loopbackio/loopback-next/tree/master/extensions/typeorm/README.md.

In addition, there is also a TypeORM Connection Booter which means this manual registration should not be necessary:

https://github.com/loopbackio/loopback-next/blob/master/extensions/typeorm/src/typeorm-connection.booter.ts#L69

dhmlau:
closing this PR per the [above comment](https://github.com/loopbackio/loopback.io/pull/1382#issuecomment-1242758332).

PR #1734, opened 2024-09-16
chore: update LoopBack 4 LTS EOL year

ivuorinen:
As per https://github.com/loopbackio/loopback-next/discussions/10403#discussioncomment-8702019 and https://github.com/loopbackio/loopback-next/blob/master/README.md#long-term-support

This commit also fixes 3 markdownlint problems.


achrinza (approved):

achrinza:
Thanks @ivuorinen! Appreciate the fixup.

PR #2049, opened 2026-07-08
ci: migrate to pure ghactions deploy; harden runner

### LINKED ISSUE #2048: Migrate away from `github-pages` gem
This gem has been [deprecated](https://github.com/github/pages-gem/issues/651#issuecomment-2712968982) for quite a while. This gem locks us into very specific gem versions, and will continue to slip further from upstream gem updates, and currently breaks with Ruby 4.x.

We're using a compatibility shim for the old GitHub Pages Legacy Worker workflow, which [also uses](https://github.com/actions/jekyll-build-pages/blob/c2d3f388e1c98cd181c6371f87c58e0a3313e693/Gemfile#L8) `github-pages` gem under the hood.

This shim ultimately uses the newer GitHub Actions deployment workflow under the hood, so we're not losing out on features if we migrate.

Hence, we should migrate away from the shim to keep the website maintainable in the long run.

References for migrating:
- https://www.josephguadagno.net/2024/04/03/migrate-to-github-actions-from-pages-legacy-worker
- https://github.com/actions/starter-workflows/blob/1035244887e26fbbd4f1017d919fb5995cc521c4/pages/jekyll.yml

achrinza:
## Summary

This moves away from the compatibility setup which simulates Pages Legacy Worker workflow.

This simplifies the setup, since it's no longer pegged to old dependencies via the `github-pages` gem.

Feature-wise is the same; The compat layer still ends up using to the same GitHub Actions-style deployment strategy anyways.

Furthermore, we can remove dependency on the external GitHub Action with long-lived SSH credentials.

This also hardens the runner.

## Testing

These changes have been tested on a fork: https://github.com/achrinzafork/loopback.io/commit/ce0098d15400fad36a82ea3b66a2bd52ccf1015b, with a slight deviation of adding `rubygems.org:443` into `allowed-endpoints` for Ruby cache misses.

## Before Merging

- Reconfigure GitHub Pages deployment method to from `Deploy from a branch` to `GitHub Actions`

    If we don't change this, GitHub will auto-deploy a workflow alongside this workflow, and the last one to complete will override the first one.

## After merging

- Remove `DEPLOY_KEY` [GitHub Actions secret](https://github.com/loopbackio/loopback.io/settings/secrets/actions)
- Revoke `DEPLOY_KEY` [from this repo](https://github.com/loopbackio/loopback.io/settings/keys)

closes: https://github.com/loopbackio/loopback.io/issues/2048
see: https://github.com/loopbackio/security/issues/26


achrinza:
- GitHub Pages deployment method has been changed to GitHub Actions
- `DEPLOY_KEY` has been removed and revoked.

